PR #6, opened 2018-07-10
XRDDEV-39

iluwatar:
- Refactor reading version information from xroad-proxy package


src/proxy/src/main/java/ee/ria/xroad/proxy/util/VersionUtil.java
```diff
@@ -0,0 +1,56 @@
...
+        try {
+            String cmd;
+            if (Files.exists(Paths.get("/etc/redhat-release"))) {
+                cmd = "rpm -q --queryformat '%{VERSION}-%{RELEASE}' xroad-proxy";
+            } else {
+                cmd = "dpkg-query -f '${Version}' -W xroad-proxy";
+            }
+            Process p = Runtime.getRuntime().exec(cmd);
+            p.waitFor();
+            version = IOUtils.toString(p.getInputStream(), Charset.defaultCharset());
```
jhyoty: Should check the exit status of the process. In case the query is unsuccessful (exit status != 0), version should be set to unknown.

src/proxy/src/main/java/ee/ria/xroad/proxy/util/VersionUtil.java
```diff
@@ -0,0 +1,56 @@
...
+        String version;
+        try {
+            String cmd;
+            if (Files.exists(Paths.get("/etc/redhat-release"))) {
+                cmd = "rpm -q --queryformat '%{VERSION}-%{RELEASE}' xroad-proxy";
+            } else {
+                cmd = "dpkg-query -f '${Version}' -W xroad-proxy";
+            }
+            Process p = Runtime.getRuntime().exec(cmd);
+            p.waitFor();
```
jhyoty: This can block forever (e.g. if the process produces more output than fits into the output buffer).

iluwatar:
Added the review fixes. The logic is now the same as in https://github.com/nordic-institute/X-Road/blob/develop/src/center-ui/app/controllers/about_controller.rb and https://github.com/nordic-institute/X-Road/blob/develop/src/proxy-ui/app/controllers/about_controller.rb.

src/common-util/src/main/java/ee/ria/xroad/common/util/HttpSender.java
```diff
@@ -111,6 +112,11 @@ private void doRequest(HttpRequestBase request) throws Exception {
 
         addAdditionalHeaders();
 
+        log.debug("Log request headers");
```
jhyoty: Maybe this should be trace-level logging? Unnecessary looping could be avoided by checking the enabled logging level (log.isTraceEnabled()).

PR #14, opened 2018-08-24
XRDDEV-85

jhyoty:
Fix bug in messagelog archiving


src/addons/messagelog/src/main/java/ee/ria/xroad/proxy/messagelog/LogArchiver.java
```diff
@@ -137,11 +140,15 @@ private boolean handleArchive(long maxTimestampId) throws Exception {
...
     }
 
-    private void archive(LogArchiveWriter archiveWriter,
+    private boolean archive(LogArchiveWriter archiveWriter,
             List<LogRecord> records) throws Exception {
+
+        boolean producedArchiveFile = false;
         for (LogRecord record : records) {
-            archiveWriter.write(record);
+            producedArchiveFile |= archiveWriter.write(record);
```
iluwatar: Why use bitwise OR here? Wouldn't simple assignment do the same or am I missing something.

PR #28, opened 2018-09-21
XRDDEV-94: Ubuntu 18.04 packaging

jhyoty:
* Add Ubuntu 18.04 packaging
* Packaging refactoring


ansible/hosts/lxd_hosts.txt
```diff
@@ -21,7 +21,7 @@ centos-ss
 
 #security servers (centos lxd containers, not supported in variant ee)
 [centos-ss]
-#xroad-lxd-rh-ss1 ansible_connection=lxd
+xroad-lxd-rh-ss1 ansible_connection=lxd
```
iluwatar: I'm not sure if this change was intended?

src/build_packages.sh
```diff
@@ -4,17 +4,18 @@ set -e
 
 XROAD=$(cd "$(dirname "$0")"; pwd)
 
-./compile_code.sh "$@"
+#./compile_code.sh "$@"
```
iluwatar: I don't think this should be commented out

PR #34, opened 2018-10-05
Estonian-6.19.0-1

kristoheero:
XRDEV-60 / XTE-425: Central Server: Added script for changing IP address of cluster nodes & updated document IG-CSHA respectively.


kristoheero:
Drop this pull request, see estonian-6.20.0-1.

PR #46, opened 2018-10-20
XRDDEV-145

petkivim:
Update time-stamping processing logic

- Add check that controls time-stamped message count at the end of the batch timestamping cycle
- If the time-stamped message count equals to `timestamp-records-limit`, launch time-stamping process again


src/addons/messagelog/src/main/java/ee/ria/xroad/proxy/messagelog/TaskQueue.java
```diff
@@ -86,6 +86,13 @@ protected void handleTimestampSucceeded(TimestampSucceeded message) {
         } finally {
             if (succeeded) {
                 indicateSuccess();
+                // If time-stamped records count equals to time-stamp records limit, there are probably
+                // still records to be time-stamped. Init another another time-stamping round to prevent
+                // messagelog records to begin to bloat.
+                if (message.getMessageRecords().length == MessageLogProperties.getTimestampRecordsLimit()) {
```
haapakanni:
```suggestion
                if (message.getMessageRecords().length >= MessageLogProperties.getTimestampRecordsLimit()) {
```

petkivim: The maximum number of records to be processed in one batch is set by `timestamp-records-limit` property. The property is used for setting max results count when the database query that's fetching the records to be time-stamped is created:

https://github.com/nordic-institute/X-Road/blob/develop/src/addons/messagelog/src/main/java/ee/ria/xroad/proxy/messagelog/TaskQueue.java#L189

```java
    private List<Task> getTimestampTasks(Session session) {
        return session.createQuery(getTaskQueueQuery()).setMaxResults(
                MessageLogProperties.getTimestampRecordsLimit()).list();
    }
```

Therefore, the number of records that are time-stamped in one batch cannot be greater than `timestamp-records-limit`.

haapakanni: Yes, that's correct, ">=" would just be defensive programming.

PR #50, opened 2018-10-26
XRDDEV-143

petkivim:
Make Signer module manager update interval configurable:

- Add new `signer.module-manager-update-interval` system property for configuring the update interval.
- Update Signer so that the new system property is used.
- Update System Properties document.
- Add a simple test for testing the new property.

### DIFF
--- a/src/signer/src/test/java/ee/ria/xroad/signer/SignerTest.java
+++ b/src/signer/src/test/java/ee/ria/xroad/signer/SignerTest.java
@@ -1,0 +1,51 @@
+/**
+ * The MIT License
+ * Copyright (c) 2018 Estonian Information System Authority (RIA),
+ * Nordic Institute for Interoperability Solutions (NIIS), Population Register Centre (VRK)
+ * Copyright (c) 2015-2017 Estonian Information System Authority (RIA), Population Register Centre (VRK)
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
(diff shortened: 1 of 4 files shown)


src/signer/src/test/java/ee/ria/xroad/signer/SignerTest.java
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Class for testing {@link Signer}
+ */
+public class SignerTest {
+
+    @Test
+    public void testModuleManagerUpdateIntervalProperty() {
+        // Test for default value
+        assertEquals(SystemProperties.SIGNER_MODULE_MANAGER_UPDATE_INTERVAL + " should be 60", 60,
```
jhyoty: If DEFAULT_SIGNER_MODULE_MANAGER_UPDATE_INTERVAL is changed, the test will fail. Would be better to use the defaut constant (could be public) unless the default must be 60.

src/signer/src/test/java/ee/ria/xroad/signer/SignerTest.java
```diff
@@ -0,0 +1,51 @@
...
+                SystemProperties.getModuleManagerUpdateInterval());
+
+        // Set value to 300
+        System.setProperty(SystemProperties.SIGNER_MODULE_MANAGER_UPDATE_INTERVAL, "300");
```
jhyoty: If the suggested change is implemented and the default is later updated to 300, this test will no longer verify that the value can be changed. It would be better to use e.g. DEFAULT+1

src/signer/src/test/java/ee/ria/xroad/signer/SignerTest.java
petkivim: Changed the default constant public and updated the test to use the constant instead of hard-coded values.

src/signer/src/test/java/ee/ria/xroad/signer/SignerTest.java
petkivim: Changed the default constant public and updated the test to use the constant instead of hard-coded values.

PR #51, opened 2018-10-28
XRDDEV-54

petkivim:
Update error message when logging in to uninitialized Central Server or Security Server with wrong user account.

- Add authorization check in the beginning of the init process.
- Change error message to `not authorized`.
- Remove unused translation strings.


petkivim:
Feature already implemented.

PR #54, opened 2018-11-02
XRDDEV-191 

iluwatar:
- Added environmental monitoring daemon and its interfaces to the X-Road architecture document


iluwatar:
There is a small watermark in the image since it was created by the Visual Paradigm CE.

arc-g_logical_structure_of_x_road

PR #65, opened 2018-11-22
Xrddev 177

carohauta:


src/common-util/src/main/java/ee/ria/xroad/common/Version.java
```diff
@@ -23,13 +23,41 @@
...
+        }
+
+        if (commitHash != null) {
+            if (commitDate == null) {
+                sb.append("-");
+            }
+            sb.append(commitHash);
+        }
+
+        return sb.toString();
```
jhyoty: Maybe building the full version could be in a static initializer block. Information that this is a "release" version is missing.

src/common-util/src/main/java/ee/ria/xroad/common/Version.java
```diff
@@ -3,17 +3,17 @@
  * Copyright (c) 2018 Estonian Information System Authority (RIA),
  * Nordic Institute for Interoperability Solutions (NIIS), Population Register Centre (VRK)
  * Copyright (c) 2015-2017 Estonian Information System Authority (RIA), Population Register Centre (VRK)
- *
+ * <p>
```
jhyoty: Please configure IDE auto formatter so that it does not convert empty lines to paragraph tags (creates unnecessary changes)

src/gradle.properties
```diff
@@ -2,3 +2,5 @@
 // We need to enable access, so that we can refer to external schemas when
 // generating classes with JAXB.
 systemProp.javax.xml.accessExternalSchema=all
+xroadVersion=6.20.0
+xroadBuildType=0
```
jhyoty: Maybe "snapshot" and "release" would be clearer. Or the version classifier could actually be an arbitrary string ("beta", "rc1",...).

PR #75, opened 2018-12-21
XRDDEV-168 SonarQube Ruby issue

carohauta:
* remove unnecessary wrapper method when rendering page heading


carohauta:
default_content_for -method is not used anywhere else so getting rid of the whole method should fix the SonarQube issue

PR #81, opened 2019-01-03
XRDDEV-256

carohauta:
* check AsicVerifier version with the version arg
* version string removed from AsicVerifier.jar

### DIFF
--- a/src/asicverifier/src/main/java/ee/ria/xroad/asicverifier/AsicVerifierMain.java
+++ b/src/asicverifier/src/main/java/ee/ria/xroad/asicverifier/AsicVerifierMain.java
@@ -26,6 +26,7 @@ package ee.ria.xroad.asicverifier;
 
 import ee.ria.xroad.common.CodedException;
 import ee.ria.xroad.common.SystemProperties;
+import ee.ria.xroad.common.Version;
 import ee.ria.xroad.common.asic.AsicContainer;
 import ee.ria.xroad.common.asic.AsicContainerEntries;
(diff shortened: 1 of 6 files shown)


src/asicverifier/src/main/java/ee/ria/xroad/asicverifier/AsicVerifierMain.java
```diff
@@ -52,7 +53,9 @@ private AsicVerifierMain() {
      * @throws Exception in case of errors
      */
     public static void main(String[] args) throws Exception {
-        if (args.length < 2) {
+        if (args.length == 1 && args[0].equals("version")) {
```
jhyoty: A common convention is to use `--version` option for displaying the version of the utility.

src/asicverifier/src/main/java/ee/ria/xroad/asicverifier/AsicVerifierMain.java
```diff
@@ -139,6 +142,11 @@ private static void writeToFile(String fileName, String contents) throws IOExcep
 
     private static void showUsage() {
         System.out.println("Usage: AsicVerifier "
-                + "<configuration path> <asic container>");
+                + "<configuration path> <asic container>"
+                + "\nor: java -jar asicverifier.jar version");
+    }
```
jhyoty: maybe simply: `Usage: java -jar asicverifier.jar [--version] <configuration path> <asic container>`
Usage could also reproduce the version information.

PR #83, opened 2019-01-07
XRDDRV-149 Version update tool

carohauta:


carohauta:
Pull request will be updated: still need to validate the version number format

carohauta:
Added version format validator

PR #97, opened 2019-02-13
hsm recovery

tomthoros:
Current state is that in case of a module failure due to a PKCS11 exception, the module is restarted. If module (re)initialization fails, the actor is stopped and the only way to get hsm back up is to manually restart signer.

The purpose of this PR is to make hsm recovery happen with a configurable exponential back-off strategy that optionally, in case of no success for some time makes signer exit with error code 1 causing it to be restarted.

Some points to emphasize:

    * Module and token full actor paths changed due to Akka's supervisor strategy being implemented by a wrapper actor. See ServiceLocator.java.

    * a PKCS11 failure in token worker is escalated to module worker. So a single failing token will cause the entire module to be restarted.

    * Changed the way module workers are periodically updated. So that in case module is restarted, it is also updated right away.

    * I actually didn't want to include the BackoffSupervisor for softtokens, but since adding one for the hardware modules causes the actor ref path to change, i thought it'd be nice for the paths to be the same for both software and hardware modules.

Additionally I removed the `Await.ready(actorSystem.terminate(), Duration.Inf()); ` from Signer shutdown method because it only reaches there after `Await.result(actorSystem.whenTerminated(), Duration.Inf());`. It seemed to deadlock there for some reason otherwise.

### DIFF
--- a/src/addons/hwtoken/src/main/java/ee/ria/xroad/signer/tokenmanager/module/HardwareModuleManagerImpl.java
+++ b/src/addons/hwtoken/src/main/java/ee/ria/xroad/signer/tokenmanager/module/HardwareModuleManagerImpl.java
@@ -24,7 +24,10 @@
  */
 package ee.ria.xroad.signer.tokenmanager.module;
 
+import ee.ria.xroad.signer.SignerMain;
+
 import akka.actor.Props;
+import akka.actor.Terminated;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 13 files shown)


raits:
Hello,

Unfortunately, we have to reject this PR as we do not have the associated signed CLA for it.

Please submit a signed CLA and open a new PR in case you wish the contribution to be reviewed.

PR #105, opened 2019-02-28
XRDDEV-358 OpMon REST support

carohauta:
Related to #104


jhyoty:
There are some conflicts and this PR seems to re-introduce the changes from #104. Please fix the PR so that it contains only the relevant changes.

PR #117, opened 2019-03-26
XRDDEV-419

TJaakkola:
Add x-request-id (unique id for each rest request) to x-road message protocol


jhyoty:
The client security server should verify that the request id in response matches the request. See ee.ria.xroad.proxy.clientproxy.ClientRestMessageProcessor#checkConsistency

PR #121, opened 2019-03-27
XRDDEV-432: Prevent adding rest services with duplicate service codes

TJaakkola:


jhyoty:
It would be better to enforce the uniqueness of the serviceId on the server-side (like adding the wsdl description does).

PR #128, opened 2019-04-04
Support for adding autologin pin via environment varible

samueljon:
As a common way of injecting data to containers I have added environment variable that is used at startup if it is defined. This variable is then written to /etc/xroad/autologin

* Added support for injecting autologin pin at runtime which allows usage of secrets f.example in kubernetes or openshift
* Fixed linting errors in readme
* Added example docker-compose.yml in README.md

### DIFF
--- a/Docker/securityserver/README.md
+++ b/Docker/securityserver/README.md
@@ -12,45 +12,88 @@ The installed security server is in uninitialized state.
 Admin UI credentials: xrd/secret
 
 ## Running
-```
-# Simple example 
+
+```shell
+# Simple example
 docker run --name my-ss niis/xroad-security-server
+
+# Simple example passing in the XROAD_AUTOLOGIN_PIN for autologin
+docker run --name my-ss niis/xroad-security-server -e XROAD_AUTOLOGIN_PIN="1234"
 
 # Running exact version instead of the default latest version
 docker run --name my-ss niis/xroad-security-server:bionic-6.20.0
+
+# Running exact version instead of the default latest version passing in the XROAD_AUTOLOGIN_PIN for autologin
+docker run --name my-ss niis/xroad-security-server:bionic-6.20.0 -e XROAD_AUTOLOGIN_PIN="1234"
 ```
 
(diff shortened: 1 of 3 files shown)


Docker/securityserver/README.md
```diff
@@ -12,45 +12,88 @@ The installed security server is in uninitialized state.
...
 ## Running
-```
-# Simple example 
+
+```shell
+# Simple example
 docker run --name my-ss niis/xroad-security-server
 
+# Simple example passing in the XROAD_AUTOLOGIN_PIN for autologin
+docker run --name my-ss niis/xroad-security-server -e XROAD_AUTOLOGIN_PIN="1234"
```
jhyoty: This should probably be:
```suggestion
docker run --name my-ss -e XROAD_AUTOLOGIN_PIN="1234" niis/xroad-security-server 
```
There is a similar problem in the other examples as well.

Docker/securityserver/files/ss-entrypoint.sh
```diff
@@ -26,4 +26,13 @@ else
     echo "WARN: Installed version ($INSTALLED_VERSION) does not match packaged version ($PACKAGED_VERSION)" >&2
 fi
 
+if [  -z "${XROAD_AUTOLOGIN_PIN}" ]
+then
+    echo "WARN: XROAD_AUTOLOGIN_PIN variable is not set"
```
jhyoty: The warning seems a bit noisy since this is an optional method for setting the pin code.

samueljon: Should we change it to INFO: or would you like to remove-it all-together @jhyoty ?

jhyoty: I was thinking about removing it altogether, but maybe it could also check if /etc/xroad/autologin is already present and omit the warning/info in that case.

jhyoty:
This PR is outdated; new PR https://github.com/nordic-institute/X-Road/pull/153

PR #141, opened 2019-05-01
Added TLSv1.3 support to nginx config files

KarenAvetyan:
Nginx supports TLS 1.3 since version 1.13.0 (released in April 2017) when built against OpenSSL 1.1.1.
If Nginx is built against older OpenSSL version TLS 1.3 is not active and TLS 1.2 is used.


petkivim:
Thank you for submitting a pull request.

We're currently redesigning and updating the X-Road architecture, and Nginx won't be used in the X-Road technology stack in the future releases anymore. It will be removed from the Security Server in version 6.24.0 and from the Central Server in some later version. Therefore, we do not add support for TLS 1.3 to the Nginx configuration at this point, but rather add it to the new tech stack later.

For the reasons explained above, this pull request is not approved.

PR #143, opened 2019-05-02
XRDDEV-501 Fix Test TSA on bionic

iluwatar:


iluwatar:
The check fails because https://github.com/nordic-institute/X-Road/pull/142 has not been merged yet

ansible/roles/xroad-ca/tasks/main.yml
```diff
@@ -1,18 +1,25 @@
...
   package:
-    name: acl
-    state: present
+      name: acl
+      state: present
 
 - name: nginx-light
   apt: name=nginx-light state=present
 
+- name: instal incron
```
jhyoty: installing incron is unnecessary (since the experimental ca web interface is not present)

PR #150, opened 2019-05-21
WIP XRDDEV-508

raulmartinez-leadin:


src/packages/src/xroad/ubuntu/generic/control
```diff
@@ -150,7 +150,7 @@ Description: X-Road security server with Estonian settings
 Package: xroad-opmonitor
 Conflicts: xroad-center
 Architecture: all
-Depends: ${misc:Depends}, xroad-base (=${binary:Version}), xroad-confclient (=${binary:Version})
+Depends: ${misc:Depends}, net-tools, iproute2, xroad-base (=${binary:Version}), xroad-confclient (=${binary:Version})
```
jhyoty: net-tools, iproute2 are used also by other installation scripts, so putting them in xroad-base -package would probably be better. Instead of netstat (from net-tools), one could use pg_isready.

doc/Manuals/ig-ss_x-road_v6_security_server_installation_guide.md
```diff
@@ -180,6 +180,10 @@ Requirements to software and settings:
 
         sudo locale-gen en_US.UTF-8
 
+-   Ensure that the packages `locales` and `software-properties-common` are present
+
+        sudo apt-get install locales software-properties-common
+
```
jhyoty: This bullet point could be before the previous one, since the command `locale-gen` will fail if `locales` is not installed.
Also, version history and the document version should be updated.

PR #163, opened 2019-07-05
Remove RANDFILE var from CA.cnf

yamakataoka:
This fixes TEST CA's openssl internal error, when we sign CSR file.

Error massage
```
Can't load ./private/.rnd into RNG
140277469962688:error:2406F079:random number generator:RAND_load_file:Cannot open file:../crypto/rand/randfile.c:88:Filename=./private/.rnd
```

system
OS: Ubuntu 18.04
Openssl: 1.1.1

### DIFF
--- a/ansible/roles/xroad-ca/common-files/home/ca/CA/CA.cnf
+++ b/ansible/roles/xroad-ca/common-files/home/ca/CA/CA.cnf
@@ -2,7 +2,6 @@
 HOME		= .
 INTERMEDIATE    = intermediate
 INTERMEDIATE    = ${ENV::INTERMEDIATE}
-RANDFILE	= $HOME/private/.rnd
 
 [ ca ]
 # `man ca`


jhyoty:
This problem can be caused by wrong file permissions (please check that `/home/ca/CA/private/.rnd` is owned by the `ca` user) or not using the `ca` user for signing the CSRs. If the `/home/ca/CA/private/.rnd` file is missing for some reason, you can create an empty one (must be owned by the `ca` user) and it will be updated when signing the next cert.

jhyoty:
https://github.com/nordic-institute/X-Road/commit/98b16b00 should ensure that the .rnd file exists in new test CA installations.

yamakataoka:
@jhyoty 
There is no `/home/ca/CA/private/.rnd` file. Yes, making empty file solved the problem.
Okey, could you tell me how to do new test CA installation on existing dev environment?

PR #165, opened 2019-07-05
XRDDEV-560

petkivim:
Implement Security Server `ownerChange` management service:

- Add `ownerChange` management service.
- Update CS UI so that requests can be approved / declined.
  - UI does not provide features for creating `ownerChange` requests manually - a complementary request is created automatically from a request from Security Server is received.
- Implement automatic approval of `ownerChange` management requests.
- Update documentation.
- Update `managementservices.wsdl ` and  `request.xsd`

### DIFF
--- a/src/center-common/app/models/owner_change_request.rb
+++ b/src/center-common/app/models/owner_change_request.rb
@@ -117,6 +117,14 @@ class OwnerChangeRequest < RequestWithProcessing
             :user => sec_serv_user))
     end
 
+    # New owner must be registered as a client on the security server
+    if !client.security_servers.include?(server)
+      raise InvalidOwnerChangeRequestException.new(
+          I18n.t("requests.not_registered_client",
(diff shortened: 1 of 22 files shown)


src/center-common/app/models/owner_change_request.rb
```diff
@@ -0,0 +1,145 @@
...
+      raise InvalidOwnerChangeRequestException.new(
+          I18n.t("requests.duplicate_owner_change_requests",
+            :user => sec_serv_user,
+            :security_server => security_server,
+            :received => previous_request.created_at,
+            :id => previous_request.id))
+    end
+  end
+
+  def validate_request
```
jhyoty: It probably should be verified that the new owner is a registered client of the security server.

petkivim: Fixed in the commit 86f9e83.

PR #177, opened 2019-08-14
XRDDEV-136

iluwatar:
- Add automatic backups for Central and Security Server
- By default backs up daily and retention policy removes backups older than 30 days
- Add `-S` switch to command line backup that allows skipping database dump. Restore can handle backups with and without included database dumps.
- Update Central and Security Server user manuals


src/packages/src/xroad/common/proxy/usr/share/xroad/scripts/get_security_server_id.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+PW=$(crudini --get /etc/xroad/db.properties '' serverconf.hibernate.connection.password)
+USER=$(crudini --get /etc/xroad/db.properties '' serverconf.hibernate.connection.username)
+PORT=$(crudini --get /etc/xroad/db.properties '' serverconf.hibernate.connection.url | cut -d '/' -f 3 | cut -d ':' -f2)
+
```
jhyoty: should also parse host so that the script works if the DB is not local.

src/packages/src/xroad/common/proxy/usr/share/xroad/scripts/autobackup_xroad_proxy_configuration.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+ID=$(source /usr/share/xroad/scripts/get_security_server_id.sh)
+if [[ -n "${ID}" ]] ; then
+  SCRIPT="/usr/share/xroad/scripts/backup_xroad_proxy_configuration.sh"
+  FILENAME="/var/lib/xroad/backup/ss-automatic-backup-$(date +\%Y_\%m_\%d_\%H\%M\%S).tar"
```
jhyoty: escaping % in unnecessary

iluwatar:
Pushed the requested changes

PR #182, opened 2019-08-22
XRDDEV-571

jhyoty:
Do not use PostgreSQL 'lo' extension. Fixes xroad-addon-messagelog does not work in AWS RDS.


jhyoty:
retest this please

PR #185, opened 2019-08-28
XRDDEV-540

jhyoty:
Check if the X-Road admin user already belongs to a group.
Do not fail installation if modifying X-Road admin user groups fails.


jhyoty:
retest this please

PR #208, opened 2019-10-02
XRDDEV-666

iluwatar:
- Update Ansible scripts to support installing Security Server with remote database
- Update documention

### DIFF
--- a/ansible/roles/xroad-ss/tasks/rhel.yml
+++ b/ansible/roles/xroad-ss/tasks/rhel.yml
@@ -4,6 +4,38 @@
   args:
     warn: no
 
+- name: Add the xroad user
+  user:
+    name: xroad_ui_user
+  when: database_admin_password != ""
+
+- name: Create xroad directory
+  file:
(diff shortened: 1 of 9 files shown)


ansible/roles/xroad-ss/tasks/rhel.yml
```diff
@@ -21,6 +53,14 @@
 - name: add xroad admin user
   command: "xroad-add-admin-user {{ xroad_ui_user }}"
 
+- name: disable service postgresql and ensure it is masked in case we are using remote database server
+  systemd:
+    name: postgresql
+    state: stopped
+    enabled: no
+    masked: yes
+  when: database_admin_password != ""
```
jhyoty: * could be lifted to the main.yml to avoid duplication (same for RHEL and Ubuntu)
* stopping the db can break things:
  * the configured "remote database" used could actually be the local host db
  * there can be other uses for the local db --> stopping it should at least be configurable

iluwatar:
Pushed the requested changes

PR #262, opened 2019-11-20
XRDDEV-577: Add warning to openapi3 documentation about possible unde…

TJaakkola:
…sired feature of addClient endpoint


src/proxy-ui-api/src/main/resources/openapi-definition.yaml
```diff
@@ -505,7 +505,7 @@ paths:
         - security server
       summary: add new client for the security server
       operationId: addClient
-      description: Adds new client to the system
+      description: Adds new client to the system. Note that with this endpoint it is possible to add a client with an unregistered member which may be undesired.
```
jansu76: I think this explanation does not quite explain how ignore_warnings works (see https://jira.niis.org/browse/XRDDEV-577?focusedCommentId=11390&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-11390)

Both old and new proxy-ui have a mechanism for returning "warnings" related to some operation and letting user decide whether to ignore those warnings. More details:
- https://confluence.niis.org/pages/viewpage.action?pageId=8520435#APIGuidelines&Examples-Warnings

I think openapi-definition should document that:
1. addClient checks if the client is already registered
2. if it is not registered, and ClientAdd.ignore_warnings = false, addClient fails (error object shows the warning)
3. if it is not registered, and ClientAdd.ignore_warnings = true, addClient succeeds
4. (maybe?) explain this point from guidelines: "Technically it is possible to execute REST request with ignoreWarnings = true already in the first request, but this is not recommended and not the way the mechanism is intended to be used"

So IMO the point of this ticket is to explain to the end user "how and why should I use ClientAdd.ignore_warnings?"

jansu76: Did some editorial changes since I could not help myself. Also merged with develop to fix the conflict.

PR #263, opened 2019-11-20
XRDDEV-684 Generate key

carohauta:
Generate a new key for token.


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/TokensApiController.java
```diff
@@ -126,13 +139,26 @@ private Token getTokenFromService(String id) {
...
-        return new ResponseEntity<>(token, HttpStatus.OK);
+    }
+
+    @PreAuthorize("hasAuthority('GENERATE_KEY')")
+    @Override
+    public ResponseEntity<Key> addKey(String tokenId, KeyLabel keyLabel) {
+        try {
+            KeyInfo keyInfo = keyService.addKey(tokenId, keyLabel.getLabel());
+            Key key = keyConverter.convert(keyInfo);
+            return new ResponseEntity<>(key, HttpStatus.OK);
```
jansu76: Instead of HTTP 200, should return 201 with some special headers, using ApiUtil.createCreatedResponse

carohauta: Oh yeah I had already forgot that! WIP

carohauta: Alrighty then: https://github.com/nordic-institute/X-Road/pull/263/files#diff-3cc2815ca4e4e2fe42be1e7394e4975fR157

PR #278, opened 2019-12-04
XRDDEV-349 CI pipeline

olliru:
https://jira.niis.org/browse/XRDDEV-349

Adds gradle task e2eTest for executing Nightwatch tests on CI

### DIFF
--- a/src/proxy-ui-api/build.gradle
+++ b/src/proxy-ui-api/build.gradle
@@ -110,10 +110,23 @@ task checkFrontAudit(type: NpmTask, dependsOn: 'npmInstall') {
 }
 
 // run e2e tests vs remote url
+if (!project.hasProperty("testTargetUrl")) {
+    ext.testTargetUrl = ""
+}
+
 task e2eTest(type: NpmTask, dependsOn: 'npmInstall') {
+
+    doFirst {
+        if (testTargetUrl.isEmpty()) {
+            throw new GradleException('testTargetUrl property was empty, it is mandatory for e2eTests')
+        }
+    }
+    
     inputs.dir("frontend")
     outputs.dir("frontend")
+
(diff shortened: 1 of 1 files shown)


src/proxy-ui-api/build.gradle
```diff
@@ -109,6 +109,16 @@ task checkFrontAudit(type: NpmTask, dependsOn: 'npmInstall') {
     }
 }
 
+// run e2e tests vs remote url
+task e2eTest(type: NpmTask, dependsOn: 'npmInstall') {
+    inputs.dir("frontend")
+    outputs.dir("frontend")
+    args = ['run', 'test:e2e', '--', '--url=' + testTargetUrl]
```
jansu76: Now if you try to do just a regular build without running end2end tests, it fails if you don't provide a dummy testTargetUrl property: 

```
../gradlew build

> Configure project :proxy-ui-api

FAILURE: Build failed with an exception.

* Where:
Build file '/home/janne/projects/niis/xroad-ui/git/X-Road-REST-UI/src/proxy-ui-api/build.gradle' line: 116

* What went wrong:
A problem occurred evaluating project ':proxy-ui-api'.
> Could not get unknown property 'testTargetUrl' for task ':proxy-ui-api:e2eTest' 
```
Maybe some conditional logic a bit like https://stackoverflow.com/a/20694968/1469083 ? Ideally it would

1. fail the build if running e2eTest and testTargetUrl is missing (I don't think we can hardcode a sensible default without polluting the publicly shared build.gradle with some server url)
2. not fail if not running e2eTest and testTargetUrl is missing

jansu76:
Successful build in https://jenkins.niis.org/view/api-based-ui/job/rest-ui-build-pull-request/156/

PR #282, opened 2019-12-04
XRDDEV-764: Add activate certificate button and apicall

TJaakkola:


TJaakkola:
These changes will merged as a part of another task

PR #289, opened 2019-12-11
XRDDEV-777 delete csr

jansu76:
https://jira.niis.org/browse/XRDDEV-777

- implements PUT /keys/{id}/delete-csr
- refactor 3 other endpoints: https://niis.slack.com/archives/GGBHVCYSY/p1575887732003200
- fix deleteClientTlsCertificate wrong http status (200 instead of 204)


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/KeyService.java
```diff
@@ -137,5 +155,52 @@ static boolean isCausedByKeyNotFound(CodedException e) {
...
+    private static String signerFaultCode(String detail) {
+        return SIGNER_X + "." + detail;
+    }
+
+    static final String KEY_NOT_FOUND_FAULT_CODE = signerFaultCode(X_KEY_NOT_FOUND);
+    static final String CSR_NOT_FOUND_FAULT_CODE = signerFaultCode(X_CSR_NOT_FOUND);
+
+    public void deleteCsr(String keyId, String csrId) throws KeyNotFoundException, CsrNotFoundException {
+        KeyInfo keyInfo = getKey(keyId);
+        CertRequestInfo csrInfo = getCsr(keyInfo, csrId);
```
TJaakkola: Is this line necessary? crsInfo is not used at all.

TJaakkola: To me it seems it is necessary for throwing exception if csr is not found. I guess declaring a variable could be omitted.

jansu76: Yes, that was the reason. Removed variable and added a comment.

PR #290, opened 2019-12-11
XRDDEV-768 XRDDEV-769 XRDDEV-764 XRDDEV-765

TJaakkola:
This shouldn't be merged to develop before XRDDEV-707. 

Backend & frontend implementation for activating and disabling a certificate.

This PR contains the following tasks:

XRDDEV-768
XRDDEV-769
XRDDEV-764
XRDDEV-765


src/proxy-ui-api/frontend/src/locales/en.json
```diff
@@ -172,6 +174,8 @@
     "removeAll": "Remove All",
     "save": "Save",
     "continue": "Continue",
+    "activate": "Activate",
+    "deactivate": "Disable",
```
jansu76: Why is this key named `deactivate` (not `disable`), when the related success message is `disableSuccess`? 

Are localization keys named after UI texts or API terms? Picking one consistent naming convention would improve readability.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/TokenCertificatesApiController.java
```diff
@@ -63,6 +63,32 @@ public TokenCertificatesApiController(TokenCertificateService tokenCertificateSe
         this.tokenCertificateConverter = tokenCertificateConverter;
     }
 
+    @Override
+    @PreAuthorize("hasAuthority('ACTIVATE_TOKEN')")
```
jansu76: We do not have `ACTIVATE_TOKEN` permission in permissions.yml (anymore?).

Instead there is `activate_disable_auth_cert` and `activate_disable_sign_cert`.

Controller should still check for "activate_disable_auth_cert or activate_disable_sign_cert", but the more strict, cert type specific check needs to be done in addition to that. 

Permission check needs to know which type is in question. Probably this must to be done in service, see https://github.com/nordic-institute/X-Road/blob/1b453f07225d5b279c4a9a60a82cdf4354fd6102/src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java#L301-L303 (maybe refactor so that common parts are shared?)

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/service/TokenCertificateServiceTest.java
```diff
@@ -52,9 +53,12 @@
...
 
+import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.fail;
 import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.doAnswer;
 import static org.mockito.Mockito.when;
+import static org.niis.xroad.restapi.service.TokenCertificateService.CERT_NOT_FOUND_FAULT_CODE;
 
 /**
  * Test TokenCertificateService
```
jansu76: This test does not mock all `TokenCertificateService` dependencies, so it is a bit integration test-y. I guess it should either a) be named `TokenCertificateServiceIntegrationTest` or b) add mocks for rest of the dependencies, such as GlobalConfFacade.

The problem with mocking only some of the dependencies is that it is easy to unknowingly use real dependencies and for example have GlobalConfFacade read a local /etc/xroad/globalconf in local dev env, and then have the tests fail on jenkins.

But this is a pre-existing problem and not related to this specific task.

src/proxy-ui-api/frontend/src/views/CertificateDetails/CertificateDetails.vue
```diff
@@ -1,17 +1,23 @@
 <template>
   <div class="wrapper xrd-view-common">
-    <div class="new-content">
+        <div class="new-content">
       <subViewTitle :title="$t('cert.certificate')" @close="close" />
-      <template v-if="certificate">
-        <div class="cert-hash-wrapper">
-          <certificateHash :hash="certificate.hash" />
+      <div class="details-view-tools" v-if="certificate">
+        <large-button v-if="!certificate.active" outlined @click="activateCertificate(certificate.certificate_details.hash)">{{$t('action.activate')}}</large-button>
```
jansu76: See comment on `TokenCertificateService`. It is not enough to check if !certificate.active, you need to also check if certificate is saved to configuration.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
```diff
@@ -331,6 +332,46 @@ private CertificateInfo importCertificate(byte[] certificateBytes, boolean isFro
...
+
+    /**
+     * Deactivates certificate by given certificateId
+     *
+     * @param certificateId
+     * @throws CertificateNotFoundException
+     */
+    public void deactivateCertificate(String certificateId) throws CertificateNotFoundException {
+        try {
+            signerProxyFacade.deactivateCert(certificateId);
```
jansu76: Should also check for business logic rules for "is deactivate enabled".

This is defined in the old UI in https://github.com/nordic-institute/X-Road/blob/1b453f07225d5b279c4a9a60a82cdf4354fd6102/src/proxy-ui/public/javascripts/keys.js#L47-L55 and is not validated in old controllers at all.

From that, the rule for "deactivate possible" seems to be
1. currently activated
2. cert is _saved to configuration_

It probably is best to implement this in https://jira.niis.org/browse/XRDDEV-806. Service implementation will add something like this https://github.com/nordic-institute/X-Road/blob/8f3e4eadd92849b795779d17393ef66975bce53b/src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java#L617-L620

Frontend rendering also needs to take into account `possible_actions` for this certificate, to show or not show the button.

See https://confluence.niis.org/display/XRDDEV/Checking+allowed+operations for more

src/proxy-ui-api/frontend/src/locales/en.json
TJaakkola: The term used in openapi-definition is deactivate while disable is used in the task description. This is the only place 'disable' is used. Simultaneously another task uses the term 'activate', not 'enable'. 

Not sure whether all deactive -words shoud be changed to disable or the other way around

TJaakkola: Also if we use 'disable' then in addition to refactoring deactivate -> disable in api, should we refactor activate -> enable?

TJaakkola: Using activate/deactive is one word change and using disable/enable needs change in multiple files (not a complicated change though)

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/TokenCertificatesApiController.java
TJaakkola: Yes, ACTIVATE_TOKEN is there https://github.com/nordic-institute/X-Road/blob/develop/src/proxy-ui-api/src/main/resources/permissions.yml#L54. Should it be removed?

I'll change the controller checks and use that verify.

jansu76: Sorry, yeah, I don't know why I said it was not there! 

It is not the correct permission anyhow, since this API is (de)activates a cert, not a token. So, activate_disable_auth_cert and activate_disable_sign_cert are the correct ones.

I will look into the activate_token permission removal myself: https://niis.slack.com/archives/GGBHVCYSY/p1577696790001400

src/proxy-ui-api/frontend/src/locales/en.json
jansu76: I would not change the API easily as it can be quite a lot of work if you need to make the change consistently across controller + services + helpers etc. The terms in API usually reflect the core classes and e.g. SignerProxy. If some term is found to be unquestionably bad, ok then, but not just because of this localization key issue. 

I don't think the terms in UI and API need to be necessarily the same. It would be obviously nicer if they were, but sometimes the terms in core classes are a bit too technical and UI prefers different terms. And having the whole backend (api + proxy-ui-api + core) speak same language is better than having (ui + api) speak one language and (core) speak a different one, IMO.

If we look at this single `en.json` file then in any case I guess it should be internally consistent? So either

- `deactivate` and `deactivateSuccess` OR
- `disable` and `disableSuccess`

?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
```diff
@@ -351,6 +394,36 @@ public CertificateInfo importCertificate(byte[] certificateBytes) throws Invalid
...
     }
 
+    /**
+     * Check user authority to the given certificate
+     *
+     * @param certificateBytes
+     * @throws InvalidCertificateException
+     * @throws AccessDeniedException
+     */
+    public void checkCertificateAuthority(byte[] certificateBytes) throws InvalidCertificateException,
```
jansu76: Some problems: wrong permissions and a bit misleading name

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
```diff
@@ -401,7 +420,7 @@ public CertificateInfo importCertificate(byte[] certificateBytes) throws Invalid
      * @throws InvalidCertificateException
      * @throws AccessDeniedException
      */
-    public void checkCertificateAuthority(byte[] certificateBytes) throws InvalidCertificateException,
+    public void verifyAuthorityToCertificate(byte[] certificateBytes) throws InvalidCertificateException,
```
jansu76: Maybe name could still be changed a bit. `verifyActivateDisableAuthority(certificateBytes)`?
There is already another type of authority to certificate -verification happening in this service, IMPORT_AUTH_CERT and IMPORT_SIGN_CERT authority on lines 306 and 312. It would be good if the method name would make it clear that this method does not handle that type of "authority to certificate".

src/proxy-ui-api/frontend/src/views/CertificateDetails/CertificateDetails.vue
jansu76: will be handled by https://jira.niis.org/browse/XRDDEV-815

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/TokenCertificatesApiController.java
jansu76: is fixed now

src/proxy-ui-api/frontend/src/locales/en.json
jansu76: we have agreed to use API terms in the keys

PR #295, opened 2019-12-20
XRDDEV-770 and 771 Register / unregister auth cert

carohauta:
Feature for registering or unregistering an auth cert.
JIRA: https://jira.niis.org/browse/XRDDEV-770 and https://jira.niis.org/browse/XRDDEV-771


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
```diff
@@ -372,6 +382,83 @@ private ClientId getClientIdForSigningCert(String instanceIdentifier, X509Certif
...
     }
 
+    /**
+     * Send the authentication certificate registration request to central server
+     * @param hash certificate hash
+     * @param securityServerAddress IP address or DNS name of the security server
+     * @throws CertificateNotFoundException
+     * @throws GlobalConfService.GlobalConfOutdatedException
+     */
+    public void registerAuthCert(String hash, String securityServerAddress) throws CertificateNotFoundException,
```
jansu76: This should in principle check business logic rules and process registration requests only if cert is in a state that allows it. Parts like this in old implementation:

```
        :register_enabled => key.usage == KeyUsageInfo::AUTHENTICATION &&
          [CertificateInfo::STATUS_SAVED].include?(cert.status),
```

But this logic will be implemented by XRDDEV-778 and 806. If one or both of those issues merges before this, it would be good to apply their validation mechanism to this. 806 offers a bit simpler way to do it. Would be something like this

```
possibleActionsRuleEngine.requirePossibleCertificateAction(PossibleActionEnum.REGISTER,
            tokenInfo, keyInfo, certificateInfo);
```

and would require lookup for tokenInfo, keyInfo (there's a new single operation getter for both based on cert hash).

If 778 and/or 806 are not merged before this PR, this is just FYI

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
```diff
@@ -372,6 +382,83 @@ private ClientId getClientIdForSigningCert(String instanceIdentifier, X509Certif
...
+        try {
+            managementRequestService.sendAuthCertRegisterRequest(securityServerId, securityServerAddress,
+                    certificateInfo.getCertificateBytes());
+            signerProxyFacade.setCertStatus(certificateInfo.getId(), CertificateInfo.STATUS_REGINPROG);
+        } catch (GlobalConfService.GlobalConfOutdatedException e) {
+            throw e;
+        } catch (Exception e) {
+            if (e instanceof CodedException) {
+                CodedException ce = (CodedException) e;
+                if (isCausedByCertNotFound(ce)) {
```
jansu76: Since getCertificateInfo already throws CertificateNotFoundException, there's a possibility to simplify exception handling by skipping isCausedByCertNotFound here.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
```diff
@@ -372,6 +382,83 @@ private ClientId getClientIdForSigningCert(String instanceIdentifier, X509Certif
...
+        }
+    }
+
+    /**
+     * Send the authentication certificate deletion request to central server
+     * @param hash certificate hash
+     * @throws CertificateNotFoundException
+     * @throws GlobalConfService.GlobalConfOutdatedException
+     */
+    public void unregisterAuthCert(String hash) throws CertificateNotFoundException,
```
jansu76: unregisterAuthCert and registerAuthCert are mostly identical, so it _might_ make sense to share code instead of duplicating the logic? I am not 100% sure though.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ManagementRequestService.java
```diff
@@ -0,0 +1,110 @@
...
+            throw new RuntimeException(e);
+        }
+    }
+
+    private ManagementRequestSender createManagementRequestSender()
+            throws GlobalConfService.GlobalConfOutdatedException {
+        globalConfService.verifyGlobalConfValidity();
+        ServerConfType serverConf = serverConfService.getServerConf();
+        ClientId sender = serverConf.getOwner().getIdentifier();
+        ClientId receiver = globalConfFacade.getManagementRequestService();
```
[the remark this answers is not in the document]
carohauta: I can see the confusion. I will refactor the name

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
carohauta: True. And furthermore if `getCertificateInfo` finds the cert and the latter method logic doesn't -> it is more likely a runtime error at that point anyway.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
carohauta: I looked into this approach but the outcome was IMO a little too complex compared to the small amount of duplicate code in it now.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
carohauta: Maybe this task should be merged only after 778 and 806 since this is a smaller one

carohauta:
[SonarQube analysis](https://sonarqube.niis.org/dashboard?id=xroad&pullRequest=295) for this PR has been rerun and it passes. For some reason the status doesn't update in GitHub

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
carohauta: Will be adding this check next

carohauta: Added and tested

PR #302, opened 2019-12-30
XRDDEV-591

iluwatar:
- Security Server respects Accept-header from client when providing error responses
- Application/json, application/xml, text/xml are supported, where application/json is the default
- Added automatic tests for the feature

### DIFF
--- a/src/proxy/src/main/java/ee/ria/xroad/proxy/clientproxy/ClientRestMessageHandler.java
+++ b/src/proxy/src/main/java/ee/ria/xroad/proxy/clientproxy/ClientRestMessageHandler.java
@@ -31,10 +31,12 @@ import ee.ria.xroad.common.conf.globalconf.GlobalConf;
 import ee.ria.xroad.common.message.RestMessage;
 import ee.ria.xroad.common.opmonitoring.OpMonitoringData;
 import ee.ria.xroad.common.util.MimeUtils;
+import ee.ria.xroad.common.util.XmlUtils;
 import ee.ria.xroad.proxy.conf.KeyConf;
 import ee.ria.xroad.proxy.util.MessageProcessorBase;
 
 import com.google.gson.stream.JsonWriter;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.http.client.HttpClient;
 import org.eclipse.jetty.http.HttpStatus;
 import org.w3c.dom.Document;
@@ -44,17 +46,9 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
-import javax.xml.parsers.ParserConfigurationException;
-import javax.xml.transform.Transformer;
-import javax.xml.transform.TransformerConfigurationException;
-import javax.xml.transform.TransformerException;
-import javax.xml.transform.TransformerFactory;
-import javax.xml.transform.dom.DOMSource;
-import javax.xml.transform.stream.StreamResult;
 
 import java.io.IOException;
 import java.io.PrintWriter;
-import java.io.StringWriter;
 
 import static ee.ria.xroad.common.ErrorCodes.X_SSL_AUTH_FAILED;
 
@@ -64,6 +58,7 @@ import static ee.ria.xroad.common.ErrorCodes.X_SSL_AUTH_FAILED;
  * the next handler (i.e. throws exception instead), if it cannot process
  * the request itself.
  */
+@Slf4j
 class ClientRestMessageHandler extends AbstractClientProxyHandler {
 
     private String requestAcceptType;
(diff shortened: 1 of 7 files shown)


src/proxy/src/main/java/ee/ria/xroad/proxy/clientproxy/ClientRestMessageHandler.java
```diff
@@ -92,17 +101,39 @@ public void sendErrorResponse(HttpServletResponse response, CodedException ex) t
...
         response.setHeader("X-Road-Error", ex.getFaultCode());
-
-        final JsonWriter writer = new JsonWriter(new PrintWriter(response.getOutputStream()));
-        writer.beginObject()
-                .name("type").value(ex.getFaultCode())
-                .name("message").value(ex.getFaultString())
-                .name("detail").value(ex.getFaultDetail())
-                .endObject()
-                .close();
+        if (requestAcceptType.equalsIgnoreCase("text/xml")
```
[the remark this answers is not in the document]
iluwatar: Pushed an update to handle multiple types in Accept-header

src/proxy/src/main/java/ee/ria/xroad/proxy/clientproxy/ClientRestMessageHandler.java
```diff
@@ -103,9 +104,9 @@ public void sendErrorResponse(HttpServletResponse response, CodedException ex) t
         }
         response.setCharacterEncoding(MimeUtils.UTF8);
         response.setHeader("X-Road-Error", ex.getFaultCode());
-        if (requestAcceptType.equalsIgnoreCase("text/xml")
-                || requestAcceptType.equalsIgnoreCase("application/xml")) {
-            response.setContentType(requestAcceptType);
+        if (StringUtils.containsIgnoreCase(requestAcceptType, "text/xml")
+                || StringUtils.containsIgnoreCase(requestAcceptType, "application/xml")) {
```
jhyoty: This partially works, but in addition to (application|text)/xml it accepts any content type containing that string, e.g. `text/xml-patch+xml` or `application/xml-patch+xml` (https://www.iana.org/assignments/media-types/media-types.xhtml).

In addition, one could expect that `Accept: text/xml` would return the XML representation using the same content type, and `Accept: text/*` would also return the XML representation (`text/xml`) and not `application/json`.

src/proxy/src/main/java/ee/ria/xroad/proxy/clientproxy/ClientRestMessageHandler.java
```diff
@@ -62,7 +72,7 @@
     MessageProcessorBase createRequestProcessor(String target,
             HttpServletRequest request, HttpServletResponse response,
             OpMonitoringData opMonitoringData) throws Exception {
-
+        requestAcceptType = request.getHeader("Accept");
```
jhyoty: getHeader returns only the first header by that name. In case there are multiple Accept headers (possible), the other values are not considered.

iluwatar:
Pushed requested changes

iluwatar:
Pushed additional changes and test for Accept-header like `application/xml; q=0.2`

src/proxy/src/main/java/ee/ria/xroad/proxy/clientproxy/ClientRestMessageHandler.java
```diff
@@ -92,17 +119,59 @@ public void sendErrorResponse(HttpServletResponse response, CodedException ex) t
...
+                    .close();
+        }
     }
 
+    private String decideErrorResponseContentType(String acceptHeaderValue) {
+        List<String> allPieces = new ArrayList<>();
+        String[] commaSplit = acceptHeaderValue.trim().split("\\s*,\\s*");
+        for (String s1: commaSplit) {
+            String[] colonSplit = s1.trim().split("\\s*;\\s*");
+            for (String s2: colonSplit) {
```
jhyoty: The first part is the media type, so the rest can be ignored.

src/proxy/src/main/java/ee/ria/xroad/proxy/clientproxy/ClientRestMessageHandler.java
```diff
@@ -92,17 +119,59 @@ public void sendErrorResponse(HttpServletResponse response, CodedException ex) t
...
+                allPieces.add(s2);
+            }
+        }
+        if (allPieces.stream().anyMatch(TEXT_XML::equalsIgnoreCase)) {
```
jhyoty: if..else..else.. could be shortened to (assuming media type is normalized to lower-case previously)
`return allPieces.stream().filter(XML_TYPES::contains).findAny().orElse(APPLICATION_JSON)`

iluwatar: It doesn't handle `text/*` correctly (should return `text/xml` content type)

jhyoty: Add a tranformation for that:
allPieces.stream().filter(XML_TYPES::contains)
.findAny()
.map("text/* to text/xml, else original")
.orElse(APPLICATION_JSON)

iluwatar: Ok, updated

src/proxy/src/main/java/ee/ria/xroad/proxy/clientproxy/ClientRestMessageHandler.java
```diff
@@ -52,8 +62,16 @@
...
  * the request itself.
  */
+@Slf4j
 class ClientRestMessageHandler extends AbstractClientProxyHandler {
 
+    private static final String TEXT_XML = "text/xml";
+    private static final String APPLICATION_XML = "application/xml";
+    private static final String TEXT_ANY = "text/*";
+    private static final String APPLICATION_JSON = "application/json";
+    private String requestAcceptType;
```
jhyoty: requestAcceptType can be overwritten by a simultaneous request. The ClientRestMessageHandler instance is shared by all requests; it can not have per-request state.

src/proxy/src/main/java/ee/ria/xroad/proxy/clientproxy/ClientRestMessageHandler.java
iluwatar: Updated

src/proxy/src/main/java/ee/ria/xroad/proxy/clientproxy/ClientRestMessageHandler.java
iluwatar: Fixed

PR #306, opened 2020-01-02
XRDDEV-806 allowed operations for key, token

jansu76:
Do not merge before https://github.com/nordic-institute/X-Road/pull/294

https://jira.niis.org/browse/XRDDEV-806

- allowed_operations for key, token
- some fixes for cert, csr allowed_operations
- adds allowed_operation checks for existing service methods:
 - token.add_key
 - token.activate
 - token.deactivate
 - token.update_friendly_name
 - key.update_friendly_name
 - key.generate_auth_csr and key.generate_sign_csr

jenkins builds are currently broken due to dependencyCheck plugin problem, successful build with dependencyCheck disabled: https://jenkins.niis.org/job/pr-build-pipeline/411/


src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/service/TokenCertificateServiceTest.java
```diff
@@ -79,58 +119,359 @@
...
     private KeyService keyService;
 
+    @MockBean
+    private GlobalConfService globalConfService;
+
+    @MockBean
+    private GlobalConfFacade globalConfFacade;
+
+    @MockBean
+    private ClientRepository clientRepository;
```
carohauta: These are probably redundant here

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/PossibleActionsRuleEngine.java
```diff
@@ -0,0 +1,279 @@
...
+     */
+    public EnumSet<PossibleActionEnum> getPossibleTokenActions(TokenInfo tokenInfo) {
+        EnumSet<PossibleActionEnum> actions = EnumSet.noneOf(PossibleActionEnum.class);
+
+        if (tokenInfo.isActive()) {
+            actions.add(PossibleActionEnum.GENERATE_KEY);
+        }
+
+        if (tokenInfo.isActive()) {
+            actions.add(PossibleActionEnum.TOKEN_DEACTIVATE);
```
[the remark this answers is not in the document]
jansu76: It could, but I think I would prefer to keep it separate. I think it helps to be able to look at each PossibleActionEnum value individually and see clearly all the conditions that cause one individual action to be enabled. And that maybe makes it a bit easier to take one action and compare it's activation to the old codebase.

This sounds a bit silly since those two tokenInfo.isActive's are so simple and close to each other but that was kind of the thought behind it, it was not completely coincidental. The principle is maybe more clear in `getPossibleCertificateActions` and `getPossibleKeyActions`, where I am sure the if-blocks could be optimized, combined and simplified somewhat by combining actions, but keeping them separated is better. And using the same approach for `getPossibleTokenActions` is nice, IMO.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/PossibleActionsRuleEngine.java
```diff
@@ -0,0 +1,279 @@
...
+
+        return actions;
+    }
+
+    /**
+     * get possible actions for a csr
+     */
+    public EnumSet<PossibleActionEnum> getPossibleCsrActions(TokenInfo tokenInfo,
+            KeyInfo keyInfo,
+            CertRequestInfo certRequestInfo) {
```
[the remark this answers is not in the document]
jansu76: Fixing that.

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/service/TokenCertificateServiceTest.java
jansu76: I guess they are not? I mean if I don't mock those, TokenCertificateService will e.g. use real GlobalConfService, which we do not want here since it's not meant to be an integration test.

This is pretty clumsy way to do it though, I agree, and usually you forget to add some dependencies as mocks, and can easily call the real implementation by mistake. I wonder if there's a better way to do this, or if it makes sense to use this approach at all.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/PossibleActionsRuleEngine.java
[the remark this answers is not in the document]
jansu76: fixed

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/service/TokenCertificateServiceTest.java
jansu76: Thought a bit more about and realized that if we mock everything, TokenCertificateServiceTest is then a pure unit test, so what sense does it make to implement it as a (slow) SpringBootTest? Would it not be better to just have it as a regular junit test which is more lightweight and guarantees we don't accidentally use some injected real dependencies?

And then a counterpoint, spring test framework makes constructing, injecting and mocking the dependencies easier (probably, not sure how it would be done in pure mockito) and some of the tests use spring features such as mock users.

I have a feeling this is related to the problem with slow tests, I need to ask for help somewhere. I don't think it should stop this PR though.

carohauta: In this case I couldn't find any usage (direct or indirect) for the fields in question. But it might be good to mock them anyway to ensure nothing calls the real services.

It would be very nice to get the tests to execute faster but yeah that is for another task

PR #309, opened 2020-01-08
XRDDEV-794

TJaakkola:
Implements apis for adding REST services with:
- REST API base path
- OpenAPI 3 description URL

### DIFF
--- a/src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
+++ b/src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
@@ -66,6 +66,7 @@ import org.niis.xroad.restapi.service.TokenService;
 import org.niis.xroad.restapi.service.UnhandledWarningsException;
 import org.niis.xroad.restapi.util.ResourceUtils;
 import org.niis.xroad.restapi.wsdl.InvalidWsdlException;
+import org.niis.xroad.restapi.wsdl.OpenApiParser;
 import org.niis.xroad.restapi.wsdl.WsdlParser;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.core.io.Resource;
@@ -321,6 +322,7 @@ public class ClientsApiController implements ClientsApi {
         ClientId clientId = clientConverter.convertId(id);
         String url = serviceDescription.getUrl();
         boolean ignoreWarnings = serviceDescription.getIgnoreWarnings();
+        String restServiceCode = serviceDescription.getRestServiceCode();
 
         ServiceDescriptionType addedServiceDescriptionType = null;
         if (serviceDescription.getType() == ServiceType.WSDL) {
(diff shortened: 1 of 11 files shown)


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
```diff
@@ -336,9 +341,25 @@ private ClientType getClientType(String encodedId) {
                 // deviation data (errorcode + warnings) copied
                 throw new ConflictException(e);
             }
-
         } else if (serviceDescription.getType() == ServiceType.OPENAPI3) {
-            return new ResponseEntity<>(HttpStatus.NOT_IMPLEMENTED);
+            try {
+                addedServiceDescriptionType = serviceDescriptionService.addOpenapi3ServiceDescription(clientId, url,
+                        restServiceCode, ignoreWarnings);
```
carohauta: `restServiceCode` might be null here which causes problems later. Needs a null check and proper exception (BadRequestException)

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
```diff
@@ -336,9 +341,25 @@ private ClientType getClientType(String encodedId) {
...
+            } catch (ClientNotFoundException e) {
+                throw new ResourceNotFoundException(e);
+            } catch (ServiceDescriptionService.UrlAlreadyExistsException
+                    | ServiceDescriptionService.ServiceCodeAlreadyExistsException e) {
+                throw new ConflictException(e);
+            }
+        } else if (serviceDescription.getType() == ServiceType.REST) {
+            try {
+                addedServiceDescriptionType = serviceDescriptionService.addRestEndpointServiceDescription(clientId,
+                        url, serviceDescription.getRestServiceCode());
```
carohauta: `serviceDescription.getRestServiceCode()` might return null here which causes problems later. Needs a null check and proper exception (BadRequestException). Could also use the above introduced `restServiceCode` variable.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
```diff
@@ -235,55 +251,210 @@ public ServiceDescriptionType addWsdlServiceDescription(ClientId clientId, Strin
...
+    /**
+     * Check whether the ServiceDescriptions url already exists in the linked Client
+     *
+     * @param serviceDescription
+     * @throws UrlAlreadyExistsException
+     */
+    private void checkDuplicateUrl(ServiceDescriptionType serviceDescription) throws UrlAlreadyExistsException {
+        Boolean hasDuplicates = serviceDescription.getClient().getServiceDescription().stream()
+                .map(other -> !serviceDescription.equals(other) && serviceDescription.getUrl().equals(other.getUrl()))
+                .reduce(false, (prev, curr) -> prev || curr);
```
carohauta: Could also be simplified with `Stream#anyMatch`
```java
Boolean hasDuplicates = serviceDescription.getClient().getServiceDescription().stream()
                .anyMatch(other -> !serviceDescription.equals(other) && serviceDescription.getUrl().equals(other.getUrl()));
```

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
```diff
@@ -290,14 +291,13 @@ public ServiceDescriptionType addOpenapi3ServiceDescription(ClientId clientId, S
...
         verifyAuthority("ADD_OPENAPI3");
 
-        // Parse openapi definition
-        OpenApiParser.Result result = null;
-        try {
-            result = openApiParser.parse(url);
-        } catch (OpenApiParser.ParsingException e) {
-            throw e;
+        if (serviceCode == null) {
+            throw new BadRequestException("Missing ServiceCode");
```
carohauta: The `BadRequestException` should be moved into the controller. 

This particular case is somewhat unusual because the serviceCode param should actually be validated already in the API layer but we can't do that because we have a shared endpoint for adding WSDL's and OPENAPI/REST types.

carohauta: There is also another one [here](https://github.com/nordic-institute/X-Road/pull/309/commits/6ac518dbfb9a83dd12384cc9b3192012a662ae7b#diff-d5a64e5ee379cb7ea245a3fc7a9247e0R406-R408)

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
```diff
@@ -336,9 +342,27 @@ private ClientType getClientType(String encodedId) {
...
+            }
+        } else if (serviceDescription.getType() == ServiceType.REST) {
+            try {
+                addedServiceDescriptionType = serviceDescriptionService.addRestEndpointServiceDescription(clientId,
+                        url, restServiceCode);
+            } catch (ClientNotFoundException e) {
+                throw new ResourceNotFoundException(e);
+            } catch (MissingParameterException e) {
+                throw new BadRequestException(e);
+            }
```
carohauta: Redundant catch

carohauta: Ignore this. This is fine

PR #312, opened 2020-01-09
XRDDEV-792 delete CSR

Riippi:
https://jira.niis.org/browse/XRDDEV-792

- delete CSR
- added new css icons in keys list
- removed one Vuetify style override because the same can be achieved now with an "official" parameter


src/proxy-ui-api/frontend/src/views/KeysAndCertificates/SignAndAuthKeys/KeysTable.vue
```diff
@@ -199,6 +220,28 @@ export default Vue.extend({
     importCert(hash: string): void {
       this.$emit('importCertByHash', hash);
     },
+    deleteCsr(req: TokenCertificateSigningRequest, key: Key): void {
+      this.confirmDeleteCsr = true;
+      this.selectedCsr = req;
+      this.selectedKey = key;
+    },
+    doDeleteCsr(): void {
```
TJaakkola: I'd try to come up with better method names as deleteCsr & doDeleteCsr sound basicly the same thing, but still the logic inside is quite different.

PR #319, opened 2020-01-15
XRDDEV-843 unregister auth cert bug

carohauta:
Fixes a bug where unregistering an auth cert failed and it couldn't be marked for deletion.

Adds a new REST API endpoint: `/token-certificates/{hash}/mark-for-deletion` that can be called to explicitly mark a cert as `DELETION_IN_PROGRESS`


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
```diff
@@ -556,29 +561,82 @@ public void registerAuthCert(String hash, String securityServerAddress) throws C
...
+                managementRequestSenderService.sendAuthCertDeletionRequest(securityServerId,
+                        certificateInfo.getCertificateBytes());
+            }
             signerProxyFacade.setCertStatus(certificateInfo.getId(), CertificateInfo.STATUS_DELINPROG);
-        } catch (GlobalConfService.GlobalConfOutdatedException | CodedException e) {
+        } catch (CodedException ce) {
+            if (isCausedBySslAuthFailure(ce)) {
+                throw new NoValidAuthCertificateException(ce);
+            }
+            throw ce;
```
carohauta: Frontend could handle this situation so that when `unregisterAuthCertificate` endpoint returns anything else than `http 204` -> a popup will be offered to mark the cert for deletion.

Now I would suggest getting rid of the `CodedException` error handling here (and let our `ExceptionTranslator` handle it as a 500) because we are sending an actual X-Road message and there are a lot of different errors that might occur.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/TokenCertificatesApiController.java
```diff
@@ -203,14 +203,31 @@ public TokenCertificatesApiController(TokenCertificateService tokenCertificateSe
...
+    }
+
+    @Override
+    @PreAuthorize("hasAuthority('SEND_AUTH_CERT_DEL_REQ')")
+    public ResponseEntity<Void> markAuthCertForDeletion(String hash) {
+        try {
+            tokenCertificateService.markAuthCertForDeletion(hash);
+        } catch (CertificateNotFoundException e) {
+            throw new ResourceNotFoundException(e);
+        } catch (GlobalConfService.GlobalConfOutdatedException | TokenCertificateService.InvalidCertificateException
```
jansu76: Not a change request about this PR, but a thought. Are we misusing http 400 bad request? Like "global conf outdated" is not a client error, it is not result of a bad API call? Perhaps some other catches here as well.

But it is a widely used pattern in our implementation, so not directly related to this PR.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
jansu76: Yeah. Maybe the last suggestion, about XroadMessageException, would be good to make our implementation work the same way as old implementation. More about this in slack discussion.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
```diff
@@ -642,11 +700,16 @@ static boolean isCausedByCertNotFound(CodedException e) {
...
+    static boolean isCausedBySslAuthFailure(CodedException e) {
+        return SSL_AUTH_FAULT_CODE.equals(e.getFaultCode());
+    }
+
     static final String DUPLICATE_CERT_FAULT_CODE = signerFaultCode(X_CERT_EXISTS);
     static final String INCORRECT_CERT_FAULT_CODE = signerFaultCode(X_INCORRECT_CERTIFICATE);
     static final String CERT_WRONG_USAGE_FAULT_CODE = signerFaultCode(X_WRONG_CERT_USAGE);
     static final String CSR_NOT_FOUND_FAULT_CODE = signerFaultCode(X_CSR_NOT_FOUND);
     static final String CERT_NOT_FOUND_FAULT_CODE = signerFaultCode(X_CERT_NOT_FOUND);
+    static final String SSL_AUTH_FAULT_CODE = clientProxyFaultCode(X_CERT_NOT_FOUND);
```
jansu76: This is a duplicate of CERT_NOT_FOUND_FAULT_CODE, on purpose or not? Anyways probably deprecated if unregisterAuthCertAndMarkForDeletion exception handling is refactored.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
```diff
@@ -556,29 +561,82 @@ public void registerAuthCert(String hash, String securityServerAddress) throws C
...
+     * @throws KeyNotFoundException
+     * @throws CertificateNotFoundException
+     */
+    public void markAuthCertForDeletion(String hash) throws SignCertificateNotSupportedException,
+            ActionNotPossibleException, GlobalConfService.GlobalConfOutdatedException, InvalidCertificateException,
+            KeyNotFoundException, CertificateNotFoundException {
+        try {
+            unregisterAuthCertAndMarkForDeletion(hash, true);
+        } catch (NoValidAuthCertificateException e) {
+            // Not a possible exception because the unregister request was never sent
```
jansu76: I think silent catches should still be avoided always, even if we know that exception is impossible currently. Throwing runtime exception is better since at least it protects against mistakes done in future, when e.g. someone modifies unregisterAuthCertAndMarkForDeletion and impossible becomes possible. Also `throw new RuntimeException(e);` is easy to add.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
jansu76: Sorry, no. Did not see the clientProxy vs signer part initially.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
carohauta: I agree that normally they should be avoided but this is somewhat the exception that proves the rule (pardon the pun); I was thinking that skipping the `NoValidAuthCertificateException` is exactly what we want to do here (since this is the `skipUnregister` method). This also hides a redundant exception from the controller layer. Anyway this will probably change with the refactoring that I'm doing to this

jansu76: I am not sure I understand. `catch (NoValidAuthCertificateException e)` should never happen currently, right? We know that by reasoning, but the poor compiler does not know it.

Since it never happens, adding a `throw new RuntimeException(e);` after the comment (I think the comment should stay, it is good to have) does not do anything. 

If instead our reasoning was wrong, or someone later modifies unregisterAuthCertAndMarkForDeletion so that it _does_ throw NoValidAuthCertificateException for some reason (stranger things have happened), we get a RuntimeException instead of http 200 OK. 

NoValidAuthCertificateException is still hidden from the controller layer.

I am guessing Sonar would also complain about current implementation.

So I am suggesting this:

```
        try {
            unregisterAuthCertAndMarkForDeletion(hash, true);
        } catch (NoValidAuthCertificateException e) {
            // Not a possible exception because the unregister request was never sent
            throw new RuntimeException(e);
        }
```

carohauta: It is now `ManagementRequestSendingFailedException` and it is also thrown to the controller layer, where it is wrapped inside `InternalServerErrorException`, because in case of e.g. `IOException` we still want to see the error message in frontend.

But to elaborate more on the previous subject: I was thinking something along the lines that the user is deliberately using an API and a method that is supposed to only mark the cert as `DELETE_IN_PROGRESS` and not send any requests at all -> there is no reason to stop code execution in the case of `NoValidAuthCertificateException` because the cert has been, in fact, successfully marked as `DELETE_IN_PROGRESS`. Of course this probably indicates a flaw in design over anything else and like you said it is probably not ok for SonarQube. 

I'll finish the rest of the task and ask for more comments

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
carohauta: There are now separate try-catch-blocks for both `sendAuthCertDeletionRequest` and `setCertStatus` and they provide different exceptions:

`sendAuthCertDeletionRequest` -> `ManagementRequestSendingFailedException` -> `InternalServerErrorException`

`setCertStatus` -> `RuntimeException`

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/TokenCertificatesApiController.java
carohauta: That is a good point. I think that 400 Bad Request is being misused pretty widely in the Internet and it is sometimes considered a general exception even though it should be on the client.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
jansu76: TokenCertificateService.markAuthCertForDeletion could still remove one exception type which should never happen (but should still result in RuntimeException, for safety):

```
    public void markAuthCertForDeletion(String hash) throws SignCertificateNotSupportedException,
            ActionNotPossibleException, GlobalConfService.GlobalConfOutdatedException, InvalidCertificateException,
            KeyNotFoundException, CertificateNotFoundException {
        try {
            unregisterAuthCertAndMarkForDeletion(hash, true);
        } catch (ManagementRequestSenderService.ManagementRequestSendingFailedException e) {
            // should never happen since we're not sending management requests
            throw new RuntimeException(e);
        }
    }

```

PR #326, opened 2020-01-16
XRDDEV-793 & XRDDEV-795

TJaakkola:
UIs for adding and editing REST/OPENAPI3 service


src/proxy-ui-api/frontend/src/views/Clients/Services/AddRestDialog.vue
```diff
@@ -1,13 +1,32 @@
...
+        <div class="dlg-edit-row">
+          <div class="dlg-row-title">{{$t('services.serviceType')}}</div>
+
+          <ValidationProvider
+            rules="required"
+            name="serviceType"
+            v-slot="{ errors }"
+            class="validation-provider dlg-row-input">
+
+            <v-radio-group v-model="serviceType" name="serviceType" :error-messages="errors" row>
```
Riippi: This causes [vue-i18n] Value of key 'fields.serviceType' is not a string!  warning in browser console. 
In en.json file there is "fields"  group that should have text for "serviceType".  That is where i18n gets the localisations for the field names.
I don't know in what situation the text would actually be visible in the ui, but that would fix the warning.

src/proxy-ui-api/frontend/src/views/ServiceDescriptionDetails/ServiceDescriptionDetails.vue
```diff
@@ -131,11 +148,13 @@ export default Vue.extend({
...
   data() {
     return {
-      confirmDelete: false,
+      confirmWSDLDelete: false,
+      confirmRESTDelete: false,
       confirmEditWarning: false,
       warningInfo: [],
       touched: false,
-      serviceDesc: undefined,
+      serviceDesc: {} as any,
```
Riippi: Maybe import ServiceDescription from types.ts and use that instead of any?
serviceDesc: undefined as undefined | ServiceDescription,

PR #328, opened 2020-01-20
XRDDEV-818 Get current servercode

carohauta:
API endpoint `/security-servers?current_server=true` to get the current SecurityServer. Also adds `server_address` to `SecurityServer` OpenApi object.

JIRA: https://jira.niis.org/browse/XRDDEV-818


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/SecurityServersApiController.java
```diff
@@ -71,4 +82,20 @@ public SecurityServersApiController(GlobalConfService globalConfService,
         SecurityServer securityServer = securityServerConverter.convert(securityServerId);
         return new ResponseEntity<>(securityServer, HttpStatus.OK);
     }
+
+    @Override
+    @PreAuthorize("hasAuthority('VIEW_SECURITY_SERVERS')")
```
jansu76: Pending question about permissions needs clarification https://niis.slack.com/archives/GGBHVCYSY/p1579849271003200

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
```diff
@@ -4286,6 +4324,12 @@ components:
           example: server123
           minLength: 1
           maxLength: 255
+        server_address:
+          type: string
+          format: text
+          description: security server ip address
```
jansu76: maybe description should say just "address" or something, since it is not guaranteed to be an _ip address_, can be a name also

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/SecurityServersApiController.java
carohauta: This has been clarified to be fine

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
carohauta: Improved the description

PR #344, opened 2020-01-29
XRDDEV-837 Generate internal TLS key cert

carohauta:
Generate internal TLS key and cert
https://jira.niis.org/browse/XRDDEV-837


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceException.java
```diff
@@ -65,6 +68,28 @@ public ServiceException(String msg, ErrorDeviation errorDeviation) {
...
-
+    /**
+     * Throws the caught ServiceException as a {@link DeviationAwareRuntimeException}. The cause and
+     * {@link ErrorDeviation#metadata ErrorDeviation metadata} will be transferred from the original exception
+     * but a new error code must be provided. If the underlying exception does not have error metadata then the
+     * localized message of the underlying exception is used as metadata instead. If there is no underlying exception
+     * at all, then only the error code will be returned within the ErrorDeviation object.
+     * @param newErrorCode the new error code of the exception
+     */
+    public void throwAsDeviationAwareRuntimeException(String newErrorCode) {
```
jansu76: Not doing a full review now, but just noticed this and wanted to add my thoughts.

This raises a few questions for me:
- many DeviationAwareExceptions offer constructors that make it possible to copy deviations from another exception (https://github.com/nordic-institute/X-Road/blob/8b7f2a6f789997c822dccd6120bb1ce4f3777f84/src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/BadRequestException.java#L42-L44). I thought it would maybe better to consider putting this kind of functionality to a DeviationAwareRuntimeException constructor instead. That way "deviation copying" theme would always be handled using the same style (target exception / it's constructors) and not sometimes by target exception, sometimes by source exception. Granted there are some problems with that option (lots of constructors)
- storing localizedMessage to metadata is a bit questionable. Metadata's reason of existence was originally to transfer some structured metadata to frontend: https://confluence.niis.org/pages/viewpage.action?pageId=8520435#APIGuidelines&Examples-Errorhandling. As in "duplicateService.v1" in the example. I know we are already breaking this (for example CodedException handling in exceptiontranslator). I fear adding a generic ServiceException.throwAsDeviationAwareRuntimeException will lead into wide use to easily and quickly translate ServiceExceptions into unchecked ones, and this creates an end result where metadata loses it's original purpose, you don't really know what kind of data it contains, and it often leaks exception messages to frontend even if that was not the intention.

I think the API and error object design started from the idea that we want to hide things like the stacktrace, and original unprocessed error message, from the client. And sometimes add something extra, the error metadata, which should be limited and well known and documented in the API - when calling this endpoint you may receive this metadata.

Then some contradictory requirements started piling up - we want to display all the error detail the old system does, and we don't want to be killed by the complex error handling, and this has caused the metadata idea to suffer and rot a little bit and become fuzzy.

There's always a bit of contradiction here - on the other hand we want to "show same level of detail in error messages as the old UI" - and on the other hand we want to follow https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/REST_Security_Cheat_Sheet.md#error-handling. I don't think we can create a universal solution for this, we need to decide on it case by case (and consult @petkivim and @raits maybe more often on these), and usually one of the requirements has to compromise (either UI displays less error detail than the old, or we break OWASP recommendations).

I think I would e.g. feel better if

```
    public void generateInternalTlsKeyAndCertificate() {
        try {
            externalProcessRunner.execute(generateCertScriptPath, generateCertScriptArgs.split("\\s+"));
            restartXroadProxy();
        } catch (ProcessNotExecutableException | ProcessFailedException e) {
            e.throwAsDeviationAwareRuntimeException(KEY_CERT_GENERATION_FAILED);
        }
    }

```
would make it more explicit on how it transforms e.localizedMessage -> e.error.metadata. When thinking about just this code, the difference is small, but I am more worried about how other code will (mis)use ServiceException.throwAsDeviationAwareRuntimeException

carohauta: Discussion in Slack: https://niis.slack.com/archives/GGBHVCYSY/p1580295667011100

carohauta: The process output is no longer returned to the API caller. Instead it is logged so that one can easily find it from the logs with the `correlation-id` that is provided with the response headers. The response body in error cases now only include an error code and status.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ExternalProcessRunner.java
```diff
@@ -0,0 +1,103 @@
...
+            process.destroy();
+            throw new ProcessNotExecutableException(e);
+        }
+
+        int exitCode;
+
+        try {
+            exitCode = process.waitFor();
+        } catch (InterruptedException e) {
+            // we don't want to throw the InterruptedException from here but we want to retain the interrupted status
```
jansu76: Good that there's the `Thread.currentThread().interrupt();` which is important. 

I am a bit unsure about wrapping it in ProcessNotExecutableException though. Would just throwing InterruptedException from the methods be better? InterruptedException has a special meaning and wrapping it in ProcessNotExecutableException kind of hides this info from the callers. Declaring InterruptedException would probably be more "standard" / idiomatic way of handling this.  

For example https://stackoverflow.com/questions/3976344/handling-interruptedexception-in-java has lots of discussion which I kind of interpret advocating throwing InterruptedException unless there's special reason not to.

We would have to map InterruptedException -> InternalServerErrorException in the controller anyways, so the change would mostly affect the intermediate layers.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ExternalProcessRunner.java
```diff
@@ -0,0 +1,103 @@
...
+            process.destroy();
+        }
+
+        // if the process fails we attach the output into the exception
+        if (exitCode != 0) {
+            String fullCommandString = String.join(" ", commandWithArgs);
+            String processOutputString = String.join("\n", processOutput);
+            String errorMsg = String.format("Failed to run command '%s' with output: \n %s", fullCommandString,
+                    processOutputString);
+            throw new ProcessFailedException(errorMsg, processOutput);
```
jansu76: This stores process output in the ErrorDeviation metadata. I would like to remove that as well - it is in a sense misuse of the metadata. Maybe just add some sort of separate container property in ProcessFailedException (which may be more clear option anyways)?

We have DeviationAwareException constructors that take a DeviationAware object as a parameter and copy ErrorDeviation and metadata from there - it feels likely that the process output could start to leak to the API response by mistake. I may have written a better and longer explanation here https://github.com/nordic-institute/X-Road/pull/344#discussion_r372260209

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ExternalProcessRunner.java
```diff
@@ -0,0 +1,103 @@
...
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+@Slf4j
+@Component
+public class ExternalProcessRunner {
```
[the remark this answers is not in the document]
carohauta: It is also surprisingly hard to find a decent library to handle this. The Apache Commons one seems to be published in 2014 and is targeted for JDK 1.5 so that might be outdated. Also found other ones but they seemed overly complex.

The `ProcessBuilder` class in it's current state seems to be a really viable option for running external processes but as you mentioned there are always some common pitfalls that should be taken care of.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ExternalProcessRunner.java
```diff
@@ -0,0 +1,103 @@
...
+            process = pb.start();
+        } catch (IOException e) {
+            throw new ProcessNotExecutableException(e);
+        }
+
+        // gather output into a list of string for returning
+        List<String> processOutput = new ArrayList<>();
+        try (BufferedReader br = new BufferedReader(new InputStreamReader(process.getInputStream()))) {
+            br.lines().forEach(processOutput::add);
+        } catch (IOException e) {
```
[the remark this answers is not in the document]
carohauta: Yes the try-with-resources makes sure that all the newly instantiated `Reader`s and the underlying input stream are closed no matter what. It does look like the `UNIXProcess.destroy()` already takes care of closing all the streams. But as you said quietly closing them again does not hurt and 100% takes care of this.

But to be honest if someone in the future decides to use some custom `Process` implementation with a bad custom api that does not follow best practices (such as closing streams or doing something else nasty) I don't think we can be prepared to that.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ExternalProcessRunner.java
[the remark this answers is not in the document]
jansu76: Yeah, I tried to look at things like guava, but no luck there. It's a bit weird.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ExternalProcessRunner.java
carohauta: Yes it was not supposed to even be there because the process output is already stringified to the error message. Will be removed.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ExternalProcessRunner.java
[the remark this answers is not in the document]
jansu76: Yeah. Also on the whole this whole issue of closing streams is probably close to being over-pedantic. I just had already fallen into the rabbit hole of reading UNIXProcess sources and whatnot and figured, why not make it "bulletproof" vs public API definitions if it is not too difficult to do (plus, wanted to share and validate that I had understood the mechanisms correctly)

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ExternalProcessRunner.java
```diff
@@ -0,0 +1,103 @@
...
+            process.destroy();
+            throw new ProcessNotExecutableException(e);
+        }
+
+        int exitCode;
+
+        try {
+            exitCode = process.waitFor();
```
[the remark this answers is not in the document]
carohauta: Good thinking! Added 60s timeout here

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ExternalProcessRunner.java
carohauta: Changed this so that the `InterruptedException` is propagated after `Thread.currentThread().interrupt();`. I also discussed this with @jhyoty and we concluded that `ProcessNotExecutableException` would be incorrect information to the end user because even though the executing thread itself is interrupted the process might be able to complete.

The controller layer now converts the exception to `InternalServerErrorException` and a proper code in the error response is attached (at the moment either `internal_key_cert_interrupted` in `SystemApiController` or `wsdl_validator_interrupted` when WSDL validation fails).

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ExternalProcessRunner.java
```diff
@@ -0,0 +1,117 @@
...
+        int exitCode;
+
+        try {
+            boolean hasExited = process.waitFor(TIMEOUT, TimeUnit.MILLISECONDS);
+            // exit value cannot be asked if the process is still running after timeout - instead throw and destroy
+            if (!hasExited) {
+                throw new ProcessFailedException("Process timed out");
+            }
+            exitCode = process.exitValue();
+        } catch (InterruptedException e) {
```
jansu76: the comment "we dont want..." seems outdated

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
```diff
@@ -225,7 +225,7 @@ public ServiceDescriptionType addWsdlServiceDescription(ClientId clientId, Strin
             UnhandledWarningsException,
             ServiceAlreadyExistsException,
             InvalidUrlException,
-            WsdlUrlAlreadyExistsException {
+            WsdlUrlAlreadyExistsException, InterruptedException {
```
jansu76: Would it be good to add javadoc @throws documentation explaining where these InterruptedExceptions can originate from? At least to the public main-level service methods (addWsdlServiceDescription, updateWsdlUrl)?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ExternalProcessRunner.java
carohauta: Comment updated

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
carohauta: Fixed a few javadocs

PR #350, opened 2020-02-04
XRDDEV-718

iluwatar:
- Update op-monitoring integration test suite to work with the latest X-Road development version (6.24.x)
- Update op-monitoring testing documentation
- Fix op-monitoring not returning fault code and string in case of error


jhyoty:
In the future, it would be good to find a better solution for large test data. At least in some cases it could also be generated on-demand.

PR #351, opened 2020-02-04
XRDDEV-810 refresh OPENAPI3/REST Service Description

TJaakkola:


src/proxy-ui-api/frontend/src/global.ts
```diff
@@ -78,7 +78,9 @@ export enum Permissions {
   IMPORT_INTERNAL_SSL_CERT = 'IMPORT_INTERNAL_SSL_CERT', // Security server TLS certificate
   IMPORT_SIGN_CERT = 'IMPORT_SIGN_CERT',
   INIT_CONFIG = 'INIT_CONFIG',
-  REFRESH_WSDL = 'REFRESH_WSDL', // client > services > refresh WSDL
+  REFRESH_WSDL = 'REFRESH_WSDL', // client > services > refresh
+  REFRESH_REST = 'REFRESH_REST',
```
Riippi: Would be nice to add comment about what the permission is or where it's used. Helps to keep track on these and in the end it's easier to see if we have some unused permissions.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ServiceDescriptionsApiController.java
```diff
@@ -191,12 +191,13 @@ public ServiceDescriptionsApiController(ServiceDescriptionService serviceDescrip
...
                 | InvalidUrlException | InvalidWsdlException
-                | ServiceDescriptionService.WrongServiceDescriptionTypeException e) {
+                | ServiceDescriptionService.WrongServiceDescriptionTypeException
+                | OpenApiParser.ParsingException e) {
             throw new BadRequestException(e);
         } catch (ServiceDescriptionService.ServiceAlreadyExistsException
                 | ServiceDescriptionService.WsdlUrlAlreadyExistsException e) {
             throw new ConflictException(e);
-        } catch (ServiceDescriptionNotFoundException e) {
+        } catch (ServiceNotFoundException | ServiceDescriptionNotFoundException e) {
```
jansu76: ServiceDescriptionException is not really a HTTP 404 resource not found. API URL defines what the _resource_ is, in this case `/service-descriptions/{id}/refresh`, and that does exist. If the processing of this request fails then error is something else, like for example http 500 internal server error, http 409 conflict, or something else.

I am not 100% sure if `ServiceNotFoundException` is correct exception class to be used in this case altogether. `refreshRESTServiceDescription` does not throw it even though it claims to. `refreshOPENAPI3ServiceDescription` throws it if we had for some reason a service description without any services. So it is more a problem with a service description, not a problem of _not finding one specific service_. It could have a different exception type, or just be a (DeviationAware)RuntimeException. 

Or keep it as a ServiceNotFoundException (which is not 100% accurate IMO) but at least map it to http 500 internal server error? Isn't that what it is, something weird has happened for it to be possible, and it is not anything that client request has caused, so it cannot be http 4xx?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
```diff
@@ -534,6 +581,71 @@ public ServiceDescriptionType refreshServiceDescription(Long id, boolean ignoreW
...
+     * Refresh REST service description
+     *
+     * @param serviceDescriptionType
+     * @return {@link ServiceDescriptionType}
+     * @throws WrongServiceDescriptionTypeException if service type is not REST
+     * @throws ServiceNotFoundException             if there is no service linked to service description
+     * @throws InvalidUrlException                  if url is invalid
+     */
+    private ServiceDescriptionType refreshRESTServiceDescription(ServiceDescriptionType serviceDescriptionType)
+            throws WrongServiceDescriptionTypeException, ServiceNotFoundException, InvalidUrlException {
```
jansu76: Does not throw ServiceNotFoundException

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
```diff
@@ -534,6 +581,71 @@ public ServiceDescriptionType refreshServiceDescription(Long id, boolean ignoreW
...
+    /**
+     * Refresh REST service description
+     *
+     * @param serviceDescriptionType
+     * @return {@link ServiceDescriptionType}
+     * @throws WrongServiceDescriptionTypeException if service type is not REST
+     * @throws ServiceNotFoundException             if there is no service linked to service description
+     * @throws InvalidUrlException                  if url is invalid
+     */
+    private ServiceDescriptionType refreshRESTServiceDescription(ServiceDescriptionType serviceDescriptionType)
```
jansu76: Why does this method / functionality actually exist? Doesn't it only update refreshedDate?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
```diff
@@ -534,6 +581,71 @@ public ServiceDescriptionType refreshServiceDescription(Long id, boolean ignoreW
...
+        serviceDescriptionType.setRefreshedDate(new Date());
+
+        parseOpenapi3ToServiceDescription(serviceDescriptionType.getUrl(),
+                serviceDescriptionType.getService().get(0).getServiceCode(),
+                ignoreWarnings,
+                serviceDescriptionType);
+
+        clientRepository.saveOrUpdateAndFlush(serviceDescriptionType.getClient());
+
+        return serviceDescriptionType;
```
jansu76: This handles warnings pretty differently from what `refreshWSDLServiceDescription` does. WSDL side gives warnings **if new services were added or removed**, this does not do it AFAIK?

- I think this is how old UI works too, right? Old UI does not warn about adding / removing openapi services either?
- is this by design or accidental?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
```diff
@@ -534,6 +581,71 @@ public ServiceDescriptionType refreshServiceDescription(Long id, boolean ignoreW
...
+     * Refresh OPENAPI3 ServiceDescription
+     *
+     * @param serviceDescriptionType
+     * @param ignoreWarnings
+     * @return {@link ServiceDescriptionType}
+     * @throws WrongServiceDescriptionTypeException if service type is not openapi3
+     * @throws ServiceNotFoundException             if there is no service linked to given service description
+     * @throws UnhandledWarningsException           if unhandled warnings are found and ignoreWarnings if false
+     * @throws OpenApiParser.ParsingException       if parsing openapi3 description fails
+     * @throws InvalidUrlException                  if url is invalid
```
jansu76: `services.rb` does these checks

```
      check_duplicate_url(item)
      check_duplicate_service_codes(item)
```

refreshWSDLServiceDescription throws WsdlUrlAlreadyExistsException and ServiceAlreadyExistsException if this kind of checks fail. Are these missing from refreshOPENAPI3ServiceDescription?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
```diff
@@ -534,6 +581,71 @@ public ServiceDescriptionType refreshServiceDescription(Long id, boolean ignoreW
...
+     */
+    private ServiceDescriptionType refreshOPENAPI3ServiceDescription(ServiceDescriptionType serviceDescriptionType,
```
jansu76: `refreshOPENAPI3ServiceDescription` differs from convention used elsewhere, should probably be `refreshOpenApi3ServiceDescription`? There is also variant `parseOpenapi3ToServiceDescription` in use in several places.

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/ServiceDescriptionsApiControllerIntegrationTest.java
```diff
@@ -39,7 +39,7 @@
 import org.niis.xroad.restapi.openapi.model.ServiceDescriptionDisabledNotice;
```
jansu76: I kind of get the feeling that more tests for refreshOpenApi3Description could be good (here and in ServiceDescriptionServiceIntegrationTest)? Now `ServiceDescriptionsApiControllerIntegrationTest` tests only WSDL related operations?

Maybe "ignore warnings" feature would be good to test in ServiceDescriptionServiceIntegrationTest? Especially if we would add "add/remove services" warnings.

src/proxy-ui-api/src/main/resources/permissions.yml
```diff
@@ -33,6 +33,7 @@
 -       edit_openapi3_endpoint: [XROAD_SERVICE_ADMINISTRATOR]
 -       refresh_wsdl: [XROAD_SERVICE_ADMINISTRATOR]
 -       refresh_openapi3: [XROAD_SERVICE_ADMINISTRATOR]
+        refresh_rest: [XROAD_SERVICE_ADMINISTRATOR]
```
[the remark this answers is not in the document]
TJaakkola: I think there was no detailed reasoning for not creating refresh_rest & refresh_openapi3 (actually there is refresh_openapi3 -permission but it is required in editing openapi3 in the old ui - yes, this is a bit confusing). Old ui requires refresh_wsdl for refreshing all 3 types of servicedescriptions.

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/ServiceDescriptionsApiControllerIntegrationTest.java
TJaakkola: Yeah, I guess I could write some more tests. The reason for writing just one test for making sure refreshing works correctly is that these test would be near to identical to those for testing updating service description. Update is tested more carefully and I felt writing the same tests twice wouldn't be necessary. But it's true the implementations aren't identical and I'll write some tests.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
TJaakkola: This is what I though too, but this was clearly stated in the task description.

'when REST API URL type is "REST API base path" => "REST":
validating that the URL is syntactically correct'

src/proxy-ui-api/frontend/src/views/Clients/Services/Services.vue
```diff
@@ -24,7 +24,7 @@
         <v-btn
           v-if="showAddButton"
           color="primary"
```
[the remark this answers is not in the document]
TJaakkola: Ach good point. I completely forgot those...

src/proxy-ui-api/frontend/src/global.ts
TJaakkola: Added comments

src/proxy-ui-api/frontend/src/views/Clients/Services/Services.vue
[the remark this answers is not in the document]
TJaakkola: data-test attributes added.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
TJaakkola: Added fixes to align convention

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ServiceDescriptionsApiController.java
TJaakkola: Pushed a commit changing the exception to DeviationAwareRuntimeException in a few places related to this.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
TJaakkola: Fixed.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
TJaakkola: Also this is what old ui does.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
TJaakkola: I think these checks are unnecessary for refreshOPENAPI3ServiceDescription as Service Description of type OPENAPI3 always contains just one service (servicecode and url) which won't change on refresh.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
TJaakkola: Yes, old UI doesn't tell if endpoints are changed (refresh doesn't have effect on services).

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/ServiceDescriptionsApiControllerIntegrationTest.java
TJaakkola: Added a couple of tests.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
jansu76: Warnings will be implemented in a separate ticket https://jira.niis.org/browse/XRDDEV-918

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
jansu76: There seems to be one more still: `ServiceDescriptionService.addOpenapi3ServiceDescription`

Then some in tests, but those are not so important.

PR #355, opened 2020-02-09
Rhel 8 support for security server

samueljon:
Changed the folder structure for building both Red Hat 7 and 8 versions of RPMs. 

Added support in following files ( not a complete list, see commit logs for detail):
 
* ```src/build_packages.sh```
* ```src/packages/build-rpm.sh```
* ```src/packages/src/*/*.spec```
* ```src/systemtest/op-monitoring/zabbix_api/examples/zabbix/configure_hosts.py``` updated shebang to python2
* ```src/systemtest/op-monitoring/zabbix_api/examples/zabbix/import_sample_host.py``` updated shebang to python2
* ```rlwrap``` removed in spec file for xroad-base since it is no longer available in RHEL8

I would appreciate it if @iluwatar or @jhyoty would take this work further or assist with delegating it since I'm stuck in other projects due to organizational change at my current employer.


jhyoty:
I'm closing this, further work in https://github.com/nordic-institute/X-Road/tree/XRDDEV-911

PR #356, opened 2020-02-10
OpenAPI to ts generator from techday

raits:
https://jira.niis.org/browse/XRDDEV-759

- **npm run generate-types** generates new types based on the OpenAPI definition file in backend resources
- Fixed two locations where Service.Endpoints nullability caused issues

### DIFF
--- a/src/proxy-ui-api/frontend/src/store/modules/services.ts
+++ b/src/proxy-ui-api/frontend/src/store/modules/services.ts
@@ -61,12 +61,12 @@ export const mutations: MutationTree<ServicesState> = {
 
   setService(state, service: Service) {
     if (service.endpoints) {
-        service.endpoints = service.endpoints.sort((a: Endpoint, b: Endpoint) => {
-            const sortByGenerated = (a.generated === b.generated) ? 0 : a.generated ? -1 : 1;
-            const sortByPathSlashCount = a.path.split('/').length - b.path.split('/').length;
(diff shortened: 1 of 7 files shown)


src/proxy-ui-api/frontend/src/store/modules/services.ts
```diff
@@ -60,12 +60,14 @@ export const mutations: MutationTree<ServicesState> = {
...
 
   setService(state, service: Service) {
-    service.endpoints = service.endpoints.sort( (a: Endpoint, b: Endpoint) => {
-      const sortByGenerated = (a.generated === b.generated) ? 0 : a.generated ? -1 : 1;
-      const sortByPathSlashCount = a.path.split('/').length - b.path.split('/').length;
-      const sortByPathLength = a.path.length - b.path.length;
-      return sortByGenerated || sortByPathSlashCount || sortByPathLength;
-    });
+    if (service.endpoints) {
+      service.endpoints = service.endpoints.sort((a: Endpoint, b: Endpoint) => {
```
TJaakkola: if -clause can be removed if this optional chaining is used. 

service.endpoints = service.endpoints?.sort((a: Endpoint, b: Endpoint) => {

raits: Changed to optional chaining in services.ts

raits:
Changed services.ts to optional chaining

PR #357, opened 2020-02-10
XRDDEV-881 API for listing approved timestamping services

petkivim:
- Implement API for listing approved timestamping services.
- Update OpenAPI description.
- Add unit tests.


src/proxy-ui-api/src/main/resources/openapi-definition.yaml
```diff
@@ -3228,6 +3135,101 @@ paths:
           description: request specified an invalid format
         '500':
           description: internal server error
+
+  /timestamping-services:
```
carohauta: The indentation seems to be 4 spaces instead of 2 in this part. The openapi doc should probably be indented consistently with 2 spaces. 

I have noticed that editing yaml with IntelliJ IDEA makes it really really difficult to indent with 2 spaces so I have used other editors such as Sublime Text 3 or VSCode for that.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/converter/TimestampingServiceConverter.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Converter for timestamping services related data between openapi and service domain classes
+ */
+@Component
+public class TimestampingServiceConverter {
+
+    private final GlobalConfService globalConfService;
```
carohauta: We have tried to avoid calling services from converters to keep them nicely decoupled and the codebase consistent. Maybe change the converter to use `GlobalConfFacade` and call the required facade methods directly (`globalConfFacade.getApprovedTspName(globalConfFacade.getInstanceIdentifier(), url);`)

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
petkivim: You're right, the indentation was incorrect. That's fixed now.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/converter/TimestampingServiceConverter.java
petkivim: The implementation was updated based on your suggestion.

PR #359, opened 2020-02-10
Icelandic Certificate Profile update

samueljon:
This is an update to the Icelandic Certification profile.  The profile is more aligned to what has been implemented for FI and based on experience with working with the EJBCA Certificate Authority software solution.


src/common-util/src/main/java/ee/ria/xroad/common/certificateprofile/impl/IsAuthCertificateProfileInfo.java
```diff
@@ -39,21 +39,28 @@
...
-                                + "/" + params.getServerId().getMemberClass()).setReadOnly(true),
-
-                // Server code
                 new EnumLocalizedFieldDescriptionImpl(
-                        "CN", DnFieldLabelLocalizationKey.SERVER_DNS_NAME, "").setReadOnly(false) });
+                        "serialNumber",
+                        DnFieldLabelLocalizationKey.SERIAL_NUMBER,
+                        params.getServerId().getXRoadInstance()
+                        + "/" + params.getServerId().getServerCode()
+                        + "/" + params.getServerId().getMemberClass()
```
jhyoty: Maybe serialNumber value could be something like ` params.getServerId().getOwner().toShortString()`  if the intention is to make the DN unique (serverCode is already mapped to CN).
Note that serialNumber should be an ASN.1 PrintableString (https://en.wikipedia.org/wiki/PrintableString), and the identifiers (serverId components) can include "non-PrintableString" chars.

iluwatar:
The release window for 6.23.0 is closing soon @samueljon

gudvardur:
@samueljon gogogo ;)

samueljon:
@iluwatar @jhyoty I have updated the certificate profile for both Sign and Auth.  Please review these changes and comment on If there are any additional changes that we need to implement before we can merge. cc @petkivim

jhyoty:
In auth profile, O and OU are redundant since you already have that information in the serialNumber.  
In sign profile, the serialNumber is a mix of clientId and serverdId. I suggest that you use only serverId in constructing the serialNumber and update IsSubjectClientIdDecoder so that it retrieves the signing cert client id from the attributes (O, OU, and CN) instead of parsing serialNumber.

samueljon:
I have now changed the PR according to suggestions from Jarkko

jhyoty:
@samueljon  or @gudvardur , we noticed a possible defect in Iceland's signing certificate profile. It uses ServerId when populating the fields, which means that it is not possible to create signing certificates for other security server clients than the owner member. Is this intentional or something you would like to fix? 
The fix is to change `params.getServerId()` to `params getClientId()` when populating the `O`, `OU`, and `CN` fields in https://github.com/nordic-institute/X-Road/blob/58fc45d9a94bea6296a327421ba6955e364c07e2/src/common-util/src/main/java/ee/ria/xroad/common/certificateprofile/impl/IsSignCertificateProfileInfo.java#L58-L76

PR #364, opened 2020-02-12
XRDDEV-812 implement client.owner true/false property

jansu76:
https://jira.niis.org/browse/XRDDEV-812

ClientConverter populates all Client objects with owner (true/false) property which tells if the client is security server's owner member. This is based on a request scoped bean in a new `org.niis.xroad.restapi.cache` package.

### DIFF
--- a/src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/converter/LocalGroupConverterTest.java
+++ b/src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/converter/LocalGroupConverterTest.java
@@ -31,6 +31,7 @@ import ee.ria.xroad.common.identifier.ClientId;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
+import org.niis.xroad.restapi.cache.SecurityServerOwner;
 import org.niis.xroad.restapi.facade.GlobalConfFacade;
 import org.niis.xroad.restapi.openapi.model.LocalGroup;
 import org.springframework.boot.test.context.SpringBootTest;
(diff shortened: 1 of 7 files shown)


src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/converter/LocalGroupConverterTest.java
```diff
@@ -60,7 +61,9 @@ public String getMemberName(ClientId identifier) {
                 return MEMBER_NAME_PREFIX + identifier.getMemberCode();
             }
         };
-        clientConverter = new ClientConverter(globalConfFacade);
+        ClientId ownerId = ClientId.create("XRD2", "GOV", "M4");
```
carohauta: Not related to this task directly but it seems that this test class does not require Spring context to run -> could remove the annotations `@RunWith(SpringRunner.class)` and `@SpringBootTest` [above](https://github.com/nordic-institute/X-Road/pull/364/files#diff-61310cd629adaaaad39a3c75c48e2d6aR47-R48) to make the test run  much faster. But IIRC there already is a task on the works to improve our tests?

jansu76: Actually I don't think we have a ticket? I have just worked on the issue outside of ticketing. Maybe would be good to create one.

carohauta: I can create the task

PR #365, opened 2020-02-12
XRDDEV-913

iluwatar:
- Fix xroad-opmonitor standalone installation for Ubuntu 18

### DIFF
--- a/src/packages/src/xroad/ubuntu/generic/xroad-opmonitor.postinst
+++ b/src/packages/src/xroad/ubuntu/generic/xroad-opmonitor.postinst
@@ -35,23 +35,24 @@ configure() {
 
   db_properties=/etc/xroad/db.properties
   root_properties=/etc/xroad.properties
-
-  RET=
-  db_get xroad-common/database-host
-  db_host="$RET"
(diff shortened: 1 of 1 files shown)


src/packages/src/xroad/ubuntu/generic/xroad-opmonitor.postinst
```diff
@@ -33,24 +33,25 @@ configure() {
...
-  db_initialized=false
+  db_host=127.0.0.1:5432
 
   if  [[ -f ${root_properties} && $(get_prop ${root_properties} postgres.connection.password) != "" ]]
   then
     master_passwd=$(crudini --get ${root_properties} '' postgres.connection.password)
     export PGPASSWORD=${master_passwd}
     psql_cmd=remote_psql
+    RET=
+    db_get xroad-common/database-host
```
jhyoty: I suppose that this can still fail if opmonitor is standalone and using remote DB? This could be fixed by reverting the changes, and making the db_get not to fail (makes it also possible to use a pre-initialized remote db). Something like:
```
RET=
db_get xroad-common/database-host || true
db_host="${RET:-127.0.0.1:5432}"
db_stop
```

PR #368, opened 2020-02-13
Xrddev 817

raits:
Added the server name to the header. The server address is shown on hover.

### DIFF
--- a/src/proxy-ui-api/frontend/src/store/modules/user.ts
+++ b/src/proxy-ui-api/frontend/src/store/modules/user.ts
@@ -4,17 +4,20 @@ import _ from 'lodash';
 import { ActionTree, GetterTree, Module, MutationTree } from 'vuex';
 import { RootState } from '../types';
 import { mainTabs } from '@/global';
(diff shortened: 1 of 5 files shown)


src/proxy-ui-api/frontend/src/store/modules/user.ts
```diff
@@ -121,6 +131,20 @@ export const actions: ActionTree<UserState, RootState> = {
       });
   },
 
+  async fetchCurrentSecurityServer({ commit }) {
+    return axios.get<SecurityServer[]>('/security-servers?current_server=true')
+      .then((resp) => {
+        if (resp.data?.length !== 1) {
+          throw new Error(`Failed to set current security server. Expected the response to contain 1 item, actual was ${resp.data?.length}`);
```
Riippi: The error text should be localised?
clients.ts has an example how to use i18n  in vuex.

raits: Added localization.

PR #370, opened 2020-02-14
XRDDEV-831

raits:
Implemented the dialog functionality, also made it so saving refreshes the certificate in the parent view.

One small change also to .prettierrc to set tab width to 2.

### DIFF
--- a/src/proxy-ui-api/frontend/src/views/KeysAndCertificates/SecurityServerTlsCertificate/GenerateTlsAndCertificateDialog.vue
+++ b/src/proxy-ui-api/frontend/src/views/KeysAndCertificates/SecurityServerTlsCertificate/GenerateTlsAndCertificateDialog.vue
@@ -9,10 +9,10 @@
     :loading="loading"
   >
     <div slot="content">
-      <p>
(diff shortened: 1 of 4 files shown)


src/proxy-ui-api/frontend/src/views/KeysAndCertificates/SecurityServerTlsCertificate/GenerateTlsAndCertificateDialog.vue
```diff
@@ -0,0 +1,69 @@
...
+      type: Boolean,
+      required: true,
+    },
+  },
+  methods: {
+    save() {
+      this.loading = true;
+      api
+        .post('/system/certificate', {})
+        .then((_) => {
```
Riippi: Whi there is _  in here?

raits: Actually not needed in this instance. I am used to writing .then(_ => {}) in which case "_" marks the unused parameter, otherwise it will be an error or a warning. 

Since the linter changes it to (_) => {} then it is no longer needed.

Fixed.

PR #373, opened 2020-02-17
XRDDEV-873 Load internal tls key cert

carohauta:
`InternalKeyManager` and `InternalTrustManager` are now able to get a new `InternalSSLKey` -> this removes the need to restart the proxy process every time the internal key is rotated by an admin.

The `InternalSSLKey` is cached by default for 180 seconds to possibly avoid a file system read every time an internal request is processed.

JIRA: https://jira.niis.org/browse/XRDDEV-873


src/common-util/src/main/java/ee/ria/xroad/common/SystemProperties.java
```diff
@@ -174,6 +174,9 @@ private SystemProperties() {
     public static final String SERVER_CONF_CACHE_PERIOD =
             PREFIX + "proxy.server-conf-cache-period";
 
+    public static final String INTERNAL_KEY_CACHE_PERIOD =
+            PREFIX + "proxy.internal-key-cache-period";
```
jhyoty: Setting server-conf-cache-period to zero disables all serverconf caching, in which case this setting has no effect. Is it really necessary to have a separate cache period for the internal key?

src/proxy/src/main/java/ee/ria/xroad/proxy/util/InternalKeyManager.java
```diff
@@ -41,16 +42,14 @@
 /**
  * a KeyManager that holds the internal SSL Key
  */
-@RequiredArgsConstructor
+@NoArgsConstructor
```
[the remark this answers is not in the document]
carohauta: Removed the annotation.

src/common-util/src/main/java/ee/ria/xroad/common/SystemProperties.java
carohauta: These now use the same cache period: `proxy.server-conf-cache-period` which is 60 seconds by default. Also updated documentation accordingly

PR #380, opened 2020-02-19
XRDDEV-860 add client

jansu76:
https://jira.niis.org/browse/XRDDEV-860

New endpoint `POST /clients` for adding new clients


jansu76:
Realized there is a problem, you can add clients from other instances, and you should not be able to do so. Fixing.

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/service/ClientServiceIntegrationTest.java
```diff
@@ -114,10 +147,282 @@ public void setup() throws Exception {
...
+    }
+    private long countByType(boolean subsystems) {
+        List<ClientType> localClients = clientService.getAllLocalClients();
+        return localClients.stream()
+                .filter(client -> (client.getIdentifier().getSubsystemCode() != null) == subsystems)
+                .count();
+    }
+
+    @Test
+    public void addLocalClientSubsystemToExistingClient() throws Exception {
```
carohauta: Very thorough testing on this quite error prone feature! 👍

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ClientService.java
```diff
@@ -439,4 +464,121 @@ public ClientType deleteTlsCertificate(ClientId id, String certificateHash)
...
+        } else {
+            client.setClientStatus(ClientType.STATUS_SAVED);
+        }
+        client.setIsAuthentication(isAuthentication.name());
+        ServerConfType serverConfType = serverConfService.getServerConf();
+        client.setConf(serverConfType);
+        serverConfType.getClient().add(client);
+
+        clientRepository.saveOrUpdate(client);
+        return client;
```
carohauta: Just wondering why this does not require saving the `serverConfType` itself? Since there is `cascade=none` from `ClientType` to `ServerConf` now.

carohauta: Answering to myself here: probably the `@Transactional` annotation kicking in

jansu76: `serverConfType` is already a managed entity; if there was some change that needs to be saved to database, it would automatically be saved by the JPA implementation when `flush` happens (probably, but not guaranteed, when transaction commits, so as you said, when @Transactional kicks in). You only need to manually save entities when they are new or transient.

Also serverConfType itself is not changed, all the DB changes are related to client object and client table.

This is probably a bit inaccurate / fuzzy explanation, tried to find a more exact and concrete explanation but did not find one quickly, sorry.

PR #381, opened 2020-02-20
Xrddev 842 localgroups add e2etests

olliru:
End 2 end tests for localgroups add function
Fixed client certificate and logout tests
Updated chrome driver and configuration

### DIFF
--- a/src/proxy-ui-api/frontend/tests/e2e/pages/ssMainPage.js
+++ b/src/proxy-ui-api/frontend/tests/e2e/pages/ssMainPage.js
@@ -15,6 +15,14 @@ var navigateCommands = {
   openSettingsTab: function() {
     this.click('@settingsTab');
     return this;
+  },
+  logout: function() {
+    this.click('@userMenuButton').click('@userMenuitemLogout');
+    return this;
+  },
+  acceptLogout: function() {
+    this.click('@logoutOKButton');
+    return this;
   }
 };
 
@@ -96,7 +104,13 @@ module.exports = {
       locateStrategy: 'xpath' },
     userMenuButton: { 
       selector: 'div.v-toolbar__content button .mdi-account-circle', 
-      locateStrategy: 'css selector' }
+      locateStrategy: 'css selector' },
+    userMenuitemLogout: { 
+      selector: '#logout-list-tile', 
+      locateStrategy: 'css selector' },
+    logoutOKButton: { 
+      selector: '//div[contains(@class, "v-dialog")]//button[.//*[contains(text(), "Ok")]]', 
+      locateStrategy: 'xpath' }
   },
   sections: {
     clientsTab: {
--- a/src/proxy-ui-api/frontend/tests/e2e/pages/ssMainPage.js
+++ b/src/proxy-ui-api/frontend/tests/e2e/pages/ssMainPage.js
@@ -30,8 +30,53 @@ var clientTabCommands = {
   clickStatusHeader: function() {
     this.click('@listStatusHeader');
     return this;
+  },
+  openTestGov: function() {
+    this.click('@testGovListItem');
+    return this;
+  },
+  openTestService: function() {
+    this.click('@testServiceListItem');
+    return this;
   }
 };
+
+var clientInfoCommands = {
+  openDetailsTab: function() {
+    this.click('@detailsTab');
+    return this;
+  },
+  openServiceClientsTab: function() {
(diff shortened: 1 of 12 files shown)


src/proxy-ui-api/frontend/tests/e2e/pages/ssMainPage.js
```diff
@@ -15,35 +15,234 @@ var navigateCommands = {
...
 
-var clientTabCommands = {};
 
 module.exports = {
   url: process.env.VUE_DEV_SERVER_URL,
   commands: [navigateCommands],
   elements: {
     clientsTab: { 
-      selector: '//div[contains(@class, "v-tabs__bar")]//a[text()="clients"]', 
+      selector: '//div[contains(@class, "v-tabs-bar__content")]//a[text()="Clients"]',
```
TJaakkola: Depending on generated class-names can crash when the library generating the html is updated. This relates to the comment below. There seems to be several issues like this so I won't create more comments about those. 

No need to act on this as the situation may be getting better considering those test-data attributes?

src/proxy-ui-api/frontend/tests/e2e/specs/ss-clients-details.js
```diff
@@ -0,0 +1,85 @@
...
+  tags: ['ss', 'clients', 'clientdetails'],
+  'Security server clients list': browser => {
+    const frontPage = browser.page.ssFrontPage();
+    const mainPage = browser.page.ssMainPage();
+    const clientsTab = mainPage.section.clientsTab;
+    const clientInfo = mainPage.section.clientInfo;
+    const clientDetails = clientInfo.section.details;
+    const certificatePopup = mainPage.section.certificatePopup;
+
+    // Open SUT and check that page is loaded
```
TJaakkola: What is 'SUT'? Perhaps break it down instead :)

src/proxy-ui-api/frontend/tests/e2e/specs/ss-clients-list.js
```diff
@@ -0,0 +1,84 @@
...
+    const frontPage = browser.page.ssFrontPage();
+    const mainPage = browser.page.ssMainPage();
+    const clientsTab = mainPage.section.clientsTab
+
+    // Open SUT and check that page is loaded
+    frontPage.navigate();
+    browser.waitForElementVisible('//*[@id="app"]');
+
+    // Enter valid credentials
+    frontPage
```
TJaakkola: This block is used several times. Maybe refactor it inside a function?

src/proxy-ui-api/frontend/tests/e2e/specs/ss-clients-list.js
```diff
@@ -0,0 +1,84 @@
...
+      .signin();
+
+    // Switch to clients tab and verify
+    mainPage.openClientsTab();
+    browser.waitForElementVisible(clientsTab);
+
+    // Verify default sorting and list items
+    browser
+      .waitForElementVisible('//th[span[contains(text(),"Name")] and contains(@class, "active") and contains(@aria-sort, "ascending")]')
+      .waitForElementVisible('(//tbody/tr)[1]//span[contains(text(),"TestCom")]')
```
TJaakkola: Would it make it easier to create a function that generates these nearly identical strings. Something like this:

```
const generateSSClientTableRow = (index, text) => {
  return `(//tbody/tr)[${index}]//span[contains(text(), "${text}")]`
}
```

and then call it:
`generateSSClientTableRow(1, "TestCom")`

src/proxy-ui-api/frontend/tests/e2e/specs/ss-logout.js
```diff
@@ -0,0 +1,57 @@
...
+  'Security server logout': browser => {
+    const frontPage = browser.page.ssFrontPage();
+    const mainPage = browser.page.ssMainPage();
+
+    // Navigate to app and check that the browser has loaded the page
+    frontPage.navigate();
+    browser.waitForElementVisible('//*[@id="app"]');
+
+    // Enter valid credentials
+    frontPage
```
TJaakkola: Again this duplicated block that could be refactored to a function

jansu76:
Changes request to this PR are handled elsewhere. Closing this to clean up https://niis.slack.com/archives/GGBHVCYSY/p1595333532008600?thread_ts=1590999066.002500&cid=GGBHVCYSY

PR #385, opened 2020-02-23
XRDDEV-892 / XRDDEV-893 / XRDDEV-897 / XRDDEV-899

petkivim:
- XRDDEV-892: API for listing configuration backup files 
- XRDDEV-893: API for deleting a configuration backup file
- XRDDEV-897: API for downloading configuration backup file
- XRDDEV-899: API for backing up configuration
- Unit tests


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/repository/BackupsRepository.java
```diff
@@ -0,0 +1,130 @@
...
+    private static final int DIR_MAX_DEPTH = 1;
+
+    /**
+     * Read backup files from configuration backup path
+     * @return
+     */
+    public List<File> getBackupFiles() {
+        try (Stream<Path> walk = Files.walk(Paths.get(CONFIGURATION_BACKUP_PATH), DIR_MAX_DEPTH)) {
+            return walk.map(Path::toString)
+                    .filter(f -> f.endsWith(BACKUP_FILE_EXTENSION))
```
jansu76: Old implementation is more strict in filtering:
- https://github.com/nordic-institute/X-Road/blob/09f42ff76da5aa42da09d58fd287246e25e8bc36/src/common-ui/lib/common-ui/backup_utils.rb#L51-L52
- https://github.com/nordic-institute/X-Road/blob/09f42ff76da5aa42da09d58fd287246e25e8bc36/src/common-ui/lib/common-ui/io_utils.rb#L139-L141

Should we implement the same logic? I think this implementation would list more backups than the old UI (e.g. backup files starting with a dot)

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/repository/BackupsRepository.java
```diff
@@ -0,0 +1,130 @@
...
+            throw new RuntimeException(ioe);
+        }
+    }
+
+    /**
+     * Get the creation date/time of a backup file
+     * @param filename
+     * @return
+     */
+    public Date getCreatedAt(String filename) {
```
jansu76: Using `OffsetDateTime` here would not be wrong choice either, since that is what we will end up using after some conversions, anyway.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/repository/BackupsRepository.java
```diff
@@ -0,0 +1,130 @@
...
+        Path path = getFilePath(filename);
+        try {
+            FileTime creationTime = (FileTime) Files.getAttribute(path, "creationTime");
```
jansu76: I guess there would be also this option https://docs.oracle.com/javase/8/docs/api/java/nio/file/attribute/BasicFileAttributes.html which removes the need for attribute name String https://stackoverflow.com/a/2724009/1469083

Not very important detail though.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/repository/BackupsRepository.java
```diff
@@ -0,0 +1,130 @@
...
+            throw new RuntimeException(ioe);
+        }
+    }
+
+    /**
+     * Return configuration backup path with a trailing slash
+     * @return
+     */
+    public String getConfigurationBackupPath() {
+        return CONFIGURATION_BACKUP_PATH  + (CONFIGURATION_BACKUP_PATH.endsWith("/") ? "" : "/");
```
jansu76: Usually generic advice is that these should use `File.separator` instead of `"/"`. There is some limited use of File.separator in the codebase already.

DefaultFilepaths.CONF_BACKUP_PATH uses hard coded "/" but maybe there is some tiny change that someone will run this with overridden CONF_BACKUP_PATH on a system with different file separator.

I don't see how File.separator would break otherwise working configuration, so there should not be too much of a downside for changing. But this is not a super important detail. It's more of a style issue.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/BackupsService.java
```diff
@@ -0,0 +1,190 @@
...
+        }
+        return backup.get();
+    }
+
+    /**
+     * Check if a backup file with the given filename exists
+     * @param filename
+     * @return
+     */
+    private Optional<Backup> backupExists(String filename) {
```
jansu76: maybe more idiomatic naming is that a `backupExists()` method returns boolean, and method returning Optional is named e.g. `getBackup()`

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/BackupsService.java
```diff
@@ -0,0 +1,190 @@
...
+        backupsRepository.deleteBackupFile(filename);
+    }
+
+    /**
+     * Read backup file's content
+     * @param filename
+     * @return
+     * @throws BackupFileNotFoundException
+     */
+    public byte[] readBackupFile(String filename) throws BackupFileNotFoundException {
```
jansu76: How does this work with weird / hostile filenames? Could you use e.g. path travelsals ("../../etc/passwd")? 

Spring firewall prevents some of these on the controller layer currently, but service should be safe anyway.

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/converter/BackupsConverterTest.java
```diff
@@ -0,0 +1,87 @@
...
+import java.util.Arrays;
+import java.util.List;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * Test BackupsConverter
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest
```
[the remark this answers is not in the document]
petkivim: Fixed.

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/BackupsApiControllerTest.java
```diff
@@ -0,0 +1,209 @@
...
+
+    @MockBean
+    BackupsRepository backupsRepository;
+
+    @Autowired
+    private BackupsApiController backupsApiController;
+
+    private static final String BASE_DIR = "/tmp/backups/";
+
+    private static final String BACKUP_FILE_1_NAME = "ss-automatic-backup-2020_02_19_031502.tar";
```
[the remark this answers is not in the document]
petkivim: File removed.

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/BackupsApiControllerTest.java
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    @Test
+    @WithMockUser(authorities = { "BACKUP_CONFIGURATION" })
+    public void downloadBackup() throws Exception {
+        byte[] bytes = null;
+        try (InputStream is = getClass().getClassLoader().getResourceAsStream(BACKUP_FILE_1_NAME)) {
+            bytes = IOUtils.toByteArray(is);
+        }
+        when(backupsRepository.readBackupFile(BACKUP_FILE_1_NAME)).thenReturn(bytes);
```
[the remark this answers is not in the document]
petkivim: File removed and updated test as suggested.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/converter/BackupsConverter.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.io.File;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * Converter for backups related data between openapi and service domain classes
+ */
+@Component
+public class BackupsConverter {
```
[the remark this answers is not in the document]
petkivim: Fixed.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/BackupsService.java
```diff
@@ -0,0 +1,190 @@
...
+ */
+@Slf4j
+@Service
+@PreAuthorize("isAuthenticated()")
+public class BackupsService {
+
+    private static final String BACKUP_GENERATION_FAILED = "backup_generation_failed";
+
+    private final BackupsRepository backupsRepository;
+    private final BackupsConverter backupsConverter;
```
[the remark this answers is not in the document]
petkivim: Moved converter reference from service layer to API layer. In addition, I added a new BackupFile DTO that is used by the service layer.

https://github.com/nordic-institute/X-Road/pull/385/commits/50b21f45828a0a0deec03af392b19b01ff7c67e1#diff-4a6737d3ff19bbbfdb8b6edc282a5df2

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/repository/BackupsRepository.java
petkivim: I changed it `OffsetDateTime` which made the implementation more straightforward.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/repository/BackupsRepository.java
petkivim: You're right, the default configuration includes the trailing `/`. Since the default value can be overridden, it's possible for an admin to change the location and set the new value without the trailing `/`. Maybe not very common, but possible. However, I replaced the hardcoded `/` with `File.separator`.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/BackupsService.java
petkivim: Method name updated.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/BackupsService.java
petkivim: Before trying to read the file, the given filename is validated against existing backup files. If the given filename does not match to an existing backup filename, `BackupFileNotFoundException` exception is thrown. The existing backup file names must meet the requirements of the backup filename filter.

https://github.com/nordic-institute/X-Road/pull/385/files#diff-de66d07ae90cf7c796f633208bd7dc09R55

This should prevent accessing other than valid backup files.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/repository/BackupsRepository.java
petkivim: The new UI/API implements now the same validation rules that the old UI does. Subdirectories are excluded using `DIR_MAX_DEPTH = 1` and valid file names are limited to `"^(?!\\.)[\\w\\.\\-]+\\.tar$"` pattern.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/BackupsApiController.java
```diff
@@ -25,34 +25,77 @@
...
     }
 
     @Override
-    public Optional<NativeWebRequest> getRequest() {
-        return Optional.ofNullable(request);
+    @PreAuthorize("hasAuthority('BACKUP_CONFIGURATION')")
+    public ResponseEntity<Backup> addBackup() {
+        try {
+            Backup backup = backupsService.generateBackup();
+            return new ResponseEntity<>(backup, HttpStatus.ACCEPTED);
```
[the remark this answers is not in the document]
petkivim: Return code is changed to `CREATED`. However, I did not use ApiUtil-added details since we don't have an endpoint to access the metadata of a specific backup. This is aligned with other similar endpoints, e.g. [adding a configured timestamping service](https://github.com/nordic-institute/X-Road/blob/0626f471cf11671ce7295a782600039509c45719/src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/SystemApiController.java#L132) (`POST /api/timestamping-services`).

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/repository/BackupsRepository.java
petkivim: Changed.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/repository/BackupsRepository.java
jansu76: Very nice! Documenting what the regexp does in https://github.com/nordic-institute/X-Road/blob/7187427ed43ff2f7334fcc221c0df08a4359e57d/src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/repository/BackupRepository.java#L53-L54 is also really great, since regexps are difficult to read and adding a few lines of comments helps reviewer / modifier a lot.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/BackupsService.java
jansu76: Good, I did not fully track how readBackupFile is used - seems exploits are not possible. Also, the added more strict filename validation helps in this.

PR #387, opened 2020-02-25
XRDDEV-832: Export certificate

raits:
Implemented functionality.

Also added *.vue tab settings to .editorconf.

### DIFF
--- a/src/.editorconfig
+++ b/src/.editorconfig
@@ -19,3 +19,6 @@ indent_size=2
 
 [*.ts]
 indent_size=2
+
+[*.vue]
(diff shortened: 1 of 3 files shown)


raits:
I think the decision on whether or not to show the prompt is down to the browsers own settings. If I change mine to always ask for the location it prompts me, otherwise it always automatically downloads to the default location.

TJaakkola:
Can be merged after resolving sonarqube issues.

jhyoty:
The issues reported by sonarqube are false positives, merging is OK.

PR #389, opened 2020-02-26
XRDDEV-840 Generate internal csr

carohauta:
JIRA: https://jira.niis.org/browse/XRDDEV-840


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/SystemService.java
```diff
@@ -139,4 +153,23 @@ private String getExceptionMessage(String name, String url, String message) {
         sb.append(" ").append(message);
         return sb.toString();
     }
+
+    /**
+     * Generate internal auth cert CSR
+     * @param distinguishedName
+     * @return
+     * @throws InvalidDistinguishedNameException
```
TJaakkola: I think it would be good to add a short description about the situation(s) when this exception can occur.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/SystemService.java
```diff
@@ -139,4 +153,23 @@ private String getExceptionMessage(String name, String url, String message) {
         sb.append(" ").append(message);
         return sb.toString();
     }
+
+    /**
```
TJaakkola: Very minor thing but I think Exceptions in serviceclasses have been last thing to declare.

carohauta: Moved the exception to the bottom.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/SystemService.java
carohauta: Added javadoc description here for the `InvalidDistinguishedNameException`

PR #390, opened 2020-02-26
XRDDEV-891: Add footer with server version

raits:
JIRA ticket: https://jira.niis.org/browse/XRDDEV-891


src/proxy-ui-api/frontend/src/store/modules/user.ts
```diff
@@ -146,6 +154,15 @@ export const actions: ActionTree<UserState, RootState> = {
       });
   },
 
+  async fetchSecurityServerVersion({ commit }) {
+    return axios.get<Version>('/system/version')
+      .then((resp) => commit('setSecurityServerVersion', resp.data))
+      .catch((error) => {
+        console.error(error);
+        throw error;
```
TJaakkola: Duplicate error logging.

src/proxy-ui-api/frontend/src/views/AppLogin.vue
```diff
@@ -184,6 +185,14 @@ export default (Vue as VueConstructor<
           this.$bus.$emit('show-error', error.message);
         });
     },
+    async fetchSecurityServerVersion() {
+      this.$store
+        .dispatch('fetchSecurityServerVersion')
+        .catch((error) => {
+          console.error(error);
+          this.$bus.$emit('show-error', error.message);
```
TJaakkola: I'd say logging error is not needed here either as if control reaches catch-block there will be notification for the user and the error can be found from the network activity. Also this logs (nearly) the same error as in the previous comment

src/proxy-ui-api/frontend/src/store/modules/user.ts
raits: Removed duplicate from AppLogin.vue

src/proxy-ui-api/frontend/src/views/AppLogin.vue
raits: I think it is still good practice to log it into the console as well because that is a more obvious place to look for it rather than finding the correct request under the network tab. 

I removed the console log in this component to prevent duplication but kept it in the store level.

PR #398, opened 2020-03-02
XRDDEV-952: fixed issue with footer covering elements if the page needs to scroll

raits:
This fix involved changing the layout structure. In summary:
- The components app-toolbar and app-footer were moved from AppBase.vue to App.vue so they can be immediately under v-app (seems to be required for vuetify-s structure to work correctly)
- Changed the wrapper div around content under App.vue to be v-content instead (again, seems to be a requirement for vuetify-s layout)
- Changed AppToolbar.vue to use v-app-bar instead of v-toolbar (was recommended under console)
- Removed the "absolute" attribute from AppFooter.vue v-footer component

I manually tested as much as I knew how to and there seem to be no side-effects to these changes.

### DIFF
--- a/src/proxy-ui-api/frontend/src/App.vue
+++ b/src/proxy-ui-api/frontend/src/App.vue
@@ -1,11 +1,13 @@
 <template>
   <v-app class="xrd-app">
-    <div>
+    <app-toolbar />
+    <v-content app>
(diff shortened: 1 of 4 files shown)


raits:
Link to JIRA ticket: https://jira.niis.org/browse/XRDDEV-952

PR #399, opened 2020-03-02
XRDDEV-861 Register client

carohauta:
JIRA: https://jira.niis.org/browse/XRDDEV-861


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ManagementRequestSenderService.java
```diff
@@ -99,27 +101,32 @@ public Integer sendAuthCertDeletionRequest(byte[] authCert) throws
         }
     }
 
+    /**
+     * Sends a client register request as a normal X-Road message
+     * @param securityServer the security server id
```
jansu76: javadoc is not in sync with current interface

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ClientService.java
```diff
@@ -418,6 +416,23 @@ public ClientType deleteTlsCertificate(ClientId id, String certificateHash)
...
+     * Registers a client
+     * @param clientId client to register
+     * @throws GlobalConfOutdatedException
+     * @throws ClientNotFoundException
+     */
+    public void registerClient(ClientId clientId) throws GlobalConfOutdatedException, ClientNotFoundException {
+        ClientType client = getLocalClientOrThrowNotFound(clientId); // throws if client doesn't exist
+        try {
+            managementRequestSenderService.sendClientRegisterRequest(clientId);
+            client.setClientStatus(ClientType.STATUS_REGINPROG);
```
jansu76: This should probably verify that client can be registered, and ultimately result in http 409 if not?

It looks like client can be registered only if status = SAVED. Old implementation either

- registers right after client has been added, and hence has status SAVED https://github.com/nordic-institute/X-Road/blob/19578462600ce3ec1832e3af5ff2b0f1b1cfe01c/src/proxy-ui/public/javascripts/clients.js#L146-L187 or
- shows "register" button only if status is SAVED https://github.com/nordic-institute/X-Road/blob/19578462600ce3ec1832e3af5ff2b0f1b1cfe01c/src/proxy-ui/app/controllers/clients_controller.rb#L476-L477

There is no check in Ruby controllers, but adding this check to service layer would be consistent with how e.g. token register is now implemented.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ClientService.java
```diff
@@ -418,6 +416,23 @@ public ClientType deleteTlsCertificate(ClientId id, String certificateHash)
...
+            clientRepository.saveOrUpdate(client);
```
jansu76: Old implementation also has "cannot register owner" check that we are missing at the moment: https://github.com/nordic-institute/X-Road/blob/19578462600ce3ec1832e3af5ff2b0f1b1cfe01c/src/proxy-ui/app/controllers/clients_controller.rb#L234-L236

Should we have that? Or was it intentionally left out (for init, maybe?). 

I do not yet have clear vision of how init will eventually work, maybe one option would be to include "cannot register owner" check now, and revise that later if needed, when we implement init?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
```diff
@@ -435,4 +432,18 @@ private ClientType getClientType(String encodedId) {
         Client result = clientConverter.convert(added);
         return createCreatedResponse("/api/clients/{id}", result, result.getId());
     }
+
+    @Override
+    @PreAuthorize("hasAuthority('SEND_CLIENT_REG_REQ')")
```
jansu76: According to Sonar this method is not tested. Most of ClientsApiController other methods are tested.

But this controller method does not contain anything nontrivial, it is probably correct to not test it. So just a FYI comment.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ManagementRequestSenderService.java
carohauta: Fixed

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ClientService.java
carohauta: Added a check for this

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ClientService.java
carohauta: I added a check for this. It should be changed when implementing the init functionality – if needed.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
carohauta: Added controller tests

carohauta:
The build check fails for because of some missing dependencies?
Started a manual build: https://jenkins.niis.org/view/api-based-ui/job/rest-ui-build-pull-request/164/console

PR #400, opened 2020-03-02
XRDDEV-866 delete client

jansu76:
Do not merge before https://github.com/nordic-institute/X-Road/pull/380

Adds endpoints for client delete
- `DELETE /clients/{id}`
- `GET /clients/{id}/orphans`
- `DELETE /clients/{id}/orphans`


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenCertificateService.java
```diff
@@ -744,6 +746,35 @@ static boolean isCausedByCertNotFound(CodedException e) {
...
+     * @throws CertificateNotFoundException if certificate with given hash was not found
+     * @throws ActionNotPossibleException if delete was not possible due to cert/key/token states
+     */
+    public void deleteCertificates(List<CertificateInfo> certificateInfos)
+            throws CertificateNotFoundException, ActionNotPossibleException {
+        List<TokenInfo> tokenInfos = tokenService.getAllTokens();
+        for (CertificateInfo certificateInfo: certificateInfos) {
+            deleteCertificate(certificateInfo.getId(), tokenInfos);
+        }
+    }
```
carohauta: This is probably intentionally a non-atomic operation? Just wondering what happens in a case where some certs are deleted and some are not because in the middle of the deletion process a wild `CertificateNotFoundException` appears? 

In the case of removing orphans that shouldn't happen since the certs are not actually provided by the user. This is probably something to be looked more into if we ever need an API endpoint supporting the deletion of multiple certificates.

jansu76: Yeah, it is - kind of - intentional. If some of the certs have been deleted, and an error occurs midway, end result will be partly removed collection of certs. 

I believe the same thing would happen with the old UI implementation: https://github.com/nordic-institute/X-Road/blob/05e6443cc5212764fd68c537462a09c3750e57a9/src/proxy-ui/app/controllers/clients_controller.rb#L407-L426

I believe atomic deletion would be close to impossible since delete cert is a signer operation and core classes such as `TokenManager`, `SoftwareTokenWorker` and `HardwareTokenWorker` do not have a concepts similar to rollbackable transactions?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ClientService.java
```diff
@@ -560,6 +565,43 @@ private ClientId getPossiblyManagedEntity(ClientId transientClientId) {
         }
     }
 
+    /**
+     * Delete a local client.
+     * @param clientId
+     * @throws ActionNotPossibleException if client status did not allow delete
+     * @throws CannotDeleteOwnerException if attempted to delete the owner
+     */
+    public void deleteLocalClient(ClientId clientId) throws ActionNotPossibleException,
```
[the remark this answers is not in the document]
jansu76: added

carohauta:
Build check has failed for some weird reason. Merged 'develop' into this. This can be merged after the build check succeeds.

PR #401, opened 2020-03-03
XRDDEV-835: implemented internal CSR generation view on frontend

raits:
JIRA task: https://jira.niis.org/browse/XRDDEV-835

### DIFF
--- a/src/proxy-ui-api/frontend/src/views/KeysAndCertificates/SecurityServerTlsCertificate/GenerateInternalCsr.vue
+++ b/src/proxy-ui-api/frontend/src/views/KeysAndCertificates/SecurityServerTlsCertificate/GenerateInternalCsr.vue
@@ -51,6 +51,7 @@
   import SubViewFooter from '@/components/ui/SubViewFooter.vue';
   import LargeButton from '@/components/ui/LargeButton.vue';
   import HelpIcon from '@/components/ui/HelpIcon.vue';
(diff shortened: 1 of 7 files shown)


src/proxy-ui-api/frontend/src/views/KeysAndCertificates/SecurityServerTlsCertificate/GenerateInternalCsr.vue
```diff
@@ -0,0 +1,124 @@
...
+    methods: {
+      back(): void {
+        this.$router.go(-1);
+      },
+      generateCsr(): void {
+        this.generatingCsr = true;
+        api
+          .post('/system/certificate/csr', { name: this.distinguishedName }, { responseType: 'blob' })
+          .then((res) => {
+            const tempLink = document.createElement('a');
```
Riippi: Would it be possible to use saveResponseAsFile from helpers.ts ? Or is this a different kind of thing?

raits: Yes, that would be better. Didn't find it when initially looking for a preexisting solution. Also modified the helper function a bit:
- Made it possible to pass the default file name
- Added data-test attribute to generated link
- Remove generated link from DOM after clicking

PR #404, opened 2020-03-05
XRDDEV-900 API for uploading backup files

petkivim:
- Implement API for uploading backup files.
- Update API design.
- Add unit tests.


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/BackupService.java
```diff
@@ -147,6 +155,44 @@ public BackupFile generateBackup() throws InterruptedException {
...
+     * @param ignoreWarnings
+     * @param file
+     * @return
+     * @throws InvalidFilenameException if backup file's name is invalid and does not pass validation
+     * @throws UnhandledWarningsException if backup file with the same name already exists
+     * and ignoreWarnings is false
+     * @throws InvalidBackupFileException if backup file is not a valid tar file or the first entry of the tar file
+     * does not match to the first entry if the Security Server generated backup tar files
+     */
+    public BackupFile uploadBackup(Boolean ignoreWarnings, MultipartFile file)
```
carohauta: This method could take `String fileName` and `byte[] fileBytes` as parameters instead of a `MultipartFile`. I think that would create better separation between the controller and service layer and make this method more universal. What do you think?

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/service/BackupServiceTest.java
```diff
@@ -180,4 +190,84 @@ public void addBackupFails() throws ProcessNotExecutableException, ProcessFailed
...
+    @Test
+    public void uploadBackupFileAlreadyExistsNoOverwrite() throws InvalidFilenameException,
+            InvalidBackupFileException {
+        when(backupRepository.isFilenameValid(any(String.class))).thenReturn(true);
+        when(backupRepository.fileExists(any(String.class))).thenReturn(true);
+        try {
+            backupService.uploadBackup(false, mockMultipartFile);
+            fail("should throw UnhandledWarningsException");
+        } catch (UnhandledWarningsException expected) {
+            // success
```
carohauta: Could also assert the warning here

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/BackupsApiController.java
```diff
@@ -103,4 +108,19 @@ public BackupsApiController(BackupService backupService, BackupConverter backupC
...
+    public ResponseEntity<Backup> uploadBackup(Boolean ignoreWarnings, MultipartFile file) {
+        try {
+            BackupFile backupFile = backupService.uploadBackup(ignoreWarnings, file);
+            return new ResponseEntity<>(backupConverter.convert(backupFile), HttpStatus.CREATED);
+        } catch (InvalidFilenameException e) {
+            throw new BadRequestException(e, new ErrorDeviation(ERROR_INVALID_FILENAME));
+        } catch (UnhandledWarningsException e) {
+            throw new BadRequestException(e);
+        } catch (InvalidBackupFileException e) {
+            throw new BadRequestException(e, new ErrorDeviation(ERROR_INVALID_BACKUP_FILE));
```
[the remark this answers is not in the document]
petkivim: The metadata is meant for logging purposes only. Updated the exceptions according to the suggestions.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/BackupService.java
petkivim: Passing `String fileName` and `byte[] fileBytes` as parameters instead of `MultipartFile` is a good idea. Updated the implementation.

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/service/BackupServiceTest.java
petkivim: Fixed.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/repository/BackupRepository.java
```diff
@@ -51,7 +51,7 @@
 
     private static final String CONFIGURATION_BACKUP_PATH = SystemProperties.getConfBackupPath();
     // Criteria for valid backup file names:
-    // 1) cannot start with ".", 2) must contain one or more word characters ([a-zA-Z_0-9]), 3) must end with ".tar"
+    // 1) cannot start with ".", 2) must contain one or more word characters ([a-zA-Z_0-9.-]), 3) must end with ".tar"
     private static final String BACKUP_FILENAME_PATTERN = "^(?!\\.)[\\w\\.\\-]+\\.tar$";
```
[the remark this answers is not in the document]
petkivim: I would leave the validation in the repository layer for security reasons. In case the repository layer is accessed from multiple classes, each class must implement its own validation. When the validation is implemented on the repository layer, the validation is always guaranteed and it's done using the same rules. Since the filename comes from the client, it's extremely important that it's validated properly before it's used.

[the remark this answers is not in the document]
jansu76: Happened to see this in the github stream and got curious, so had a look. My two cents:
- validation should in general follow the layer design @carohauta linked - one reason is, that we should follow _principle of least astonishment_ and not confuse maintainers. 
- if there is a special case, where we are concerned of security, and are afraid that someone will add a service method that forgets the validation, I think that is a valid reason to have the validation **also** on repository layer. But I think it should not be missing from service layer. I was happy to see it is not!
- Of course, if we are scared of someone adding a new service method and forgetting validation, we should be scared of someone adding a new repository method and forgetting validation, too. Putting the validation on repository layer is not bulletproof (but maybe it is safer than service layer since repository layer has less modifications typically)
- I like @carohauta 's suggestion of putting the validation **logic** into somewhere else than the repository. I would still keep both validation **checks** (service and repository), just move the actual logic elsewhere. It would make the repository class a bit more consistent with other repository classes in what kind of **logic** it contains.
- repository level validation (if there is any, we should try to keep it limited) should be considered as a safety net like a database constraint is. Failing validation should be comparable to a programming error and can just throw some dirty RuntimeException, result in http 500, not care about nice error messages etc.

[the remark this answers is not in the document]
petkivim: Based on the comments, the validation logic is now moved to `FormatUtils` class, and both `BackupRepository` and `BackupService` use it from there. In addition, filename validation was added to all the methods in the repository class that read, create or delete backup files.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/repository/BackupRepository.java
```diff
@@ -90,6 +87,9 @@ public OffsetDateTime getCreatedAt(String filename) {
      * @param filename
      */
     public void deleteBackupFile(String filename) {
+        if (!FormatUtils.isValidBackupFilename(filename)) {
+            throw new RuntimeException("invalid backup filename");
```
carohauta: SonarQube requires `"invalid backup filename"` to be in a constant because it occurs at least 3 times in this class.

petkivim: Fixed

carohauta:
Merging after checks are completed

PR #409, opened 2020-03-11
XRDDEV-869

petkivim:
Add API for changing Security Server owner:

- Implement API for changing Security Server owner.
- Add API design.
- Add tests.


src/proxy-ui-api/src/main/resources/openapi-definition.yaml
```diff
@@ -898,6 +898,38 @@ paths:
           description: an existing item already exists
         '500':
           description: internal server error
+  /clients/{id}/make-owner:
```
carohauta: This endpoint could be something like `PUT /clients/owner` with the client id in the request body to be more RESTful. What do you think? I think we have already broken this pattern in a few places so this should also do just fine.

petkivim: Changed the endpoint to `PUT /clients/owner`. The new owner is defined in message body, e.g.:

```
{
  "member_class": "ORG",
  "member_code": "1234567-8"
}
```

PR #410, opened 2020-03-12
XRDDEV-638 error details

Riippi:
https://jira.niis.org/browse/XRDDEV-638

- Error snackbar has a version that stays open until user closes it. It has also an error id that user can copy.
- Snackbar doesn't use $bus anymore. It's replaced with Vuex store, which is a bit easier to understand and maintain. 
- It's possible to launch Success snackbar with localisation code or "raw text". 
- Error Snackbar has same and also possibility to use error object as a parameter.


Riippi:
I don't remember, but it probably should be discussed. 
Anyways, it's easy to adjust.

src/proxy-ui-api/frontend/src/store/modules/notifications.ts
```diff
@@ -0,0 +1,142 @@
...
+    showErrorRaw: false,
+    errorObject: undefined,
+  };
+};
+
+// Initial state. The state can be reseted with this.
+const notificationsState: NotificationsState = getDefaultState();
+
+export const getters: GetterTree<NotificationsState, RootState> = {
+  showSuccessCode(state): boolean {
```
TJaakkola: Isn't this state of type NotificationsState from above? I think types should be used always when possible - it's a lot easier for other developers to get into the implementation.

Riippi:
Added more types and solved conflicts

PR #422, opened 2020-03-19
XRDDEV-673 and 878 LocalGroup bugs

carohauta:
* new api model for adding a LocalGroup
* better API operation names and documentation
* fix error when adding rights for other clients localgroup (well throw better exception)
* add test for the latter
* change updating local group desc to patch with request body
* delete group member graceful exception

JIRA: 
https://jira.niis.org/browse/XRDDEV-878 
https://jira.niis.org/browse/XRDDEV-673


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/AccessRightService.java
```diff
@@ -222,7 +222,7 @@ public void deleteSoapServiceAccessRights(ClientId clientId, String fullServiceC
      */
     private List<AccessRightHolderDto> addSoapServiceAccessRights(ClientId clientId, String fullServiceCode,
```
TJaakkola: Lets add the missing @throws LocalGroupNotFoundException to javadoc.

TJaakkola: Also DuplicateAccessRightException is missing too. Perhaps add that too.

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
```diff
@@ -4583,6 +4591,27 @@ components:
           uniqueItems: true
           items:
             $ref: '#/components/schemas/GroupMember'
+    LocalGroupAdd: # ok
+      type: object
+      description: model for adding a new LocalGroup
+      required:
+        - code
+        - description
```
TJaakkola: Is description actually required? Just wondering it doesn't sound mandatory.

carohauta: It is required. It is also shown as the main identifier when e.g. searching for subjects (in the first column in the subjects table) -> guess it would be dull if it was just empty

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/AccessRightService.java
carohauta: Added both of them

PR #425, opened 2020-03-20
XRDDEV-998: Change updateEndpoint api to not alter service code

TJaakkola:


src/proxy-ui-api/src/main/resources/openapi-definition.yaml
```diff
@@ -4910,36 +4910,40 @@ components:
           example: false
           default: false
     Endpoint:
+      allOf:
```
carohauta: The `allOf` keyword, combined with the OpenAPI model generator that we are using, seems to generate a redundant `EndpointAllOf` object. We have mainly tried to avoid these and created for example a `ClientAdd` and `ServiceDescriptionUpdate` models that only have the allowed and required fields in them. So maybe consider separate models (without `allOf`) for these endpoint schemas.

There is a task ([XRDDEV-668](https://jira.niis.org/browse/XRDDEV-668)) about using `readOnly` keyword in those fields that aren't supposed to be sent in a request. So maybe some unifying methods should be considered and documented when that task is being investigated further. **Spoiler alert:** The `readOnly` flag has no effect on anything

TJaakkola: Removed allOf -operator and renamed EndpointPathAndMethod to EndpointUpdate.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/EndpointService.java
```diff
@@ -111,7 +111,7 @@ public void deleteEndpoint(Long id) throws EndpointNotFoundException, ClientNotF
      * @throws IllegalGeneratedEndpointUpdateException  trying to update that is generated automatically
      * @throws IllegalArgumentException                 passing illegal combination of parameters
      */
-    public EndpointType updateEndpoint(Long id, EndpointPathAndMethod.MethodEnum method, String path)
+    public EndpointType updateEndpoint(Long id, EndpointUpdate.MethodEnum method, String path)
```
carohauta: OpenAPI models shouldn't be referenced in the service layer: https://confluence.niis.org/pages/viewpage.action?spaceKey=XRDDEV&title=API+backend+implementation

Maybe map the values to a separate Method enum?

TJaakkola: I decided to just pass string version of the method to service layer extra mapping feels a bit too much here.

PR #435, opened 2020-03-26
XRDDEV-941

petkivim:
- Add API for updating existing API key.
- Add tests.
- Update documentation.


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/repository/ApiKeyRepository.java
```diff
@@ -108,6 +108,44 @@ private String createApiKey() {
...
+    /**
+     * update api key with collection of roles by key id
+     * @param id
+     * @param roleNames
+     * @return
+     * @throws InvalidRoleNameException if roleNames was empty or contained invalid roles
+     * @throws ApiKeyNotFoundException if api key was not found
+     */
+    @CacheEvict(allEntries = true, cacheNames = { LIST_ALL_KEYS_CACHE, GET_KEY_CACHE })
+    public PersistentApiKeyType update(long id, Collection<String> roleNames)
```
jansu76: Feels like this ApiKeyRepository is a bit different from the conventions used in other repositories. Other repositories mostly 
- have a saveOrUpdate method which is used for both creating and updating 
- do not deletegate simple update / delete operations to DAO, but instead just use direct calls with persistenceUtils, e.g. https://github.com/nordic-institute/X-Road/blob/735e853ba2b24871b82606d8df77478ba2283ecb/src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/repository/LocalGroupRepository.java#L104-L106

But this is earlier legacy than this ticket so nothing has to be done here. I think I will add a few lines to confluence to offer some conventions for future repository implementations. 

I think I will pick an API key related ticket next, so I might refactor this part there. Or, if you get the inspiration, you can do it also.

jansu76: Confluence updated: https://confluence.niis.org/display/XRDDEV/API+backend+conventions#APIbackendconventions-Repositories

doc/Manuals/ug-ss_x-road_6_security_server_user_guide.md
```diff
@@ -2025,7 +2027,24 @@ curl -X GET -u <user>:<password> https://localhost:4000/api/api-key -k
...
-#### 19.1.3 Revoking API keys
+#### 19.1.3 Updating API keys
+
+An existing API key is updated with a `PUT` request to `/api/api-key/{id}`. Message body must contain the roles to be
+associated with the key. Server responds with data that contains the key id and roles associated with the key.
+
+```
+curl -X PUT -u <user>:<password> https://localhost:4000/api/api-key/60 --data '["XROAD_SECURITYSERVER_OBSERVER","XROAD_REGISTRATION_OFFICER"]' --header "Content-Type: application/json" -k
+{
+  "id": 61,
```
jansu76: Looks like update targets key 60 but returns 61, is it a mistake?

doc/Manuals/ug-ss_x-road_6_security_server_user_guide.md
```diff
@@ -6,7 +6,7 @@
 
 **X-ROAD 6**
 
-Version: 2.36
+Version: 2.38
```
jansu76: Should the use case documentation also be updated, as it contains other API key related operations? https://github.com/nordic-institute/X-Road/blob/735e853ba2b24871b82606d8df77478ba2283ecb/doc/UseCases/uc-ss_x-road_use_case_model_for_security_server_management_1.4_Y-883-4.md#344-uc-ss_43-create-a-new-api-key

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/controller/ApiKeyController.java
```diff
@@ -82,6 +82,20 @@
...
+    /**
+     * update an existing api key
+     */
+    @RequestMapping(value = "/{id}", method = RequestMethod.PUT, produces = MediaType.APPLICATION_JSON_UTF8_VALUE)
+    public ResponseEntity<PublicKeyData> updateKey(@PathVariable("id") long id,
+                                                          @RequestBody List<String> roles) {
+        try {
+            PersistentApiKeyType key = apiKeyRepository.update(id, roles);
+            return new ResponseEntity<>(new PublicKeyData(key.getId(), key.getRoles()), HttpStatus.OK);
+        } catch (InvalidRoleNameException | ApiKeyRepository.ApiKeyNotFoundException e) {
```
jansu76: ApiKeyNotFoundException should probably result in ResourceNotFoundException instead (see `revoke`)

doc/Manuals/ug-ss_x-road_6_security_server_user_guide.md
petkivim: This was a mistake. It's fixed now.

doc/Manuals/ug-ss_x-road_6_security_server_user_guide.md
petkivim: Yes, it should be updated too. Editing an API key is now added to the use case document.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/controller/ApiKeyController.java
petkivim: Fixed.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/repository/ApiKeyRepository.java
petkivim: I got inspired and did some refactoring. As a result the implementation now follows the same structure with other endpoints. Repository and service layers are separated, and business logic is implemented in the service layer.

jansu76: That's great! Looks good. Very nice to have this consistent with the rest of the app now.

PR #437, opened 2020-03-26
XRDDEV-925

iluwatar:
- improve ServerConf test so that also tsp ordering is tested when reading them from database


iluwatar:
Added the ordering clause in commit https://github.com/nordic-institute/X-Road/pull/437/commits/ffb78456d3033183fcd3bde573cfb77f52dd9d40

PR #443, opened 2020-03-31
XRDDEV-947

petkivim:
Implement API for alerts that warns about:

- issues with global conf and soft token
- ongoing backup/restore processes running.


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/NotificationService.java
```diff
@@ -0,0 +1,129 @@
...
+        this.globalConfFacade = globalConfFacade;
+        this.tokenService = tokenService;
+    }
+
+    /**
+     * Checks the status of system alerts that may affect whether the system
+     * is operational or not. If alerts are disabled, the status of all alerts true.
+     * @return
+     */
+    public AlertStatus getAlerts() {
```
carohauta, line 69: Should this also check whether the Security Server has been initialized yet? That is of course still in review but since the old implementation checks the init status, maybe it should be added here also to avoid throwing "false" exceptions. If it is something that is wanted, maybe we should create a new ticket so it does not get forgotten.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/NotificationService.java
```diff
@@ -0,0 +1,129 @@
...
+
+    /**
+     * Checks if soft token pin is entered.
+     * @return
+     */
+    private boolean isSoftTokenPinEntered() {
+        Optional<TokenInfo> token = tokenService.getAllTokens().stream()
+                .filter(t -> t.getId().equals(SignerProxy.SSL_TOKEN_ID)).findFirst();
+        if (!token.isPresent()) {
+            throw new RuntimeException("soft token not found");
```
carohauta, line 102: The old implementation does not stop execution if the soft token is not found. However a RuntimeException here makes sense if the token has been properly initialized but nevertheless is not found. Adding the init check in the `getAlerts` method would actually help in this situation too

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/NotificationsApiControllerTest.java
```diff
@@ -0,0 +1,122 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.doThrow;
+import static org.mockito.Mockito.when;
+
+/**
+ * Test NotificationsApiController
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest
+public class NotificationsApiControllerTest {
```
carohauta, line 54: Nothing major but would follow other conventions if this test was in `org.niis.xroad.restapi.controller` package instead of `org.niis.xroad.restapi.openapi`

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/NotificationService.java
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * service class for handling notifications
+ */
+@Slf4j
+@Service
+@Transactional
+@PreAuthorize("isAuthenticated()")
+public class NotificationService {
+    private static OffsetDateTime backupRestoreRunningSince;
```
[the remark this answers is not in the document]
petkivim: Removed the `static` keyword from the variable and all methods accessing it.

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/NotificationsApiControllerTest.java
petkivim: Fixed.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/NotificationService.java
petkivim: Yes, you're right. It should be checked whether the Security Server is initialized or not. I'll create a new ticket about it that can be implemented once XRDDEV-904 has been completed.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/NotificationService.java
petkivim: Yes, adding the init check in the getAlerts method will solve this too. I'll create a new ticket about that.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/NotificationService.java
petkivim: The init status check will be implemented in XRDDEV-1035 (https://jira.niis.org/browse/XRDDEV-1035).

PR #449, opened 2020-04-05
XRDDEV-1015

petkivim:
Fix API key based authentication error:

- Update ApiKeyService to use method level authentication instead of class level authentication.
- Do not authenticate methods used by ApiKeyAuthenticationManager.
- Update tests.

### DIFF
--- a/src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/auth/ApiKeyAuthenticationHelper.java
+++ b/src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/auth/ApiKeyAuthenticationHelper.java
@@ -29,7 +29,6 @@ import org.niis.xroad.restapi.domain.PersistentApiKeyType;
 import org.niis.xroad.restapi.repository.ApiKeyRepository;
 import org.niis.xroad.restapi.service.ApiKeyService;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.cache.annotation.Cacheable;
 import org.springframework.security.crypto.password.PasswordEncoder;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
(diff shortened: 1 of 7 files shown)


petkivim:
Changed `ApiKeyService` back to "typical" service so that all methods require authentication, and created a helper class `ApiKeyAuthenticationHelper` that has `get(String key)` method and that does not require authentication.

In addition, moved caching from service layer to repository layer to the `ApiKeyRepository` class.  Now `ApiKeyRepository.getAllApiKeys` DB query is run when the method is invoked the first time, and the next times results are returned from cache regardless of the object that invokes the method. When the cache was implemented  on service layer, the DB query was run once when `ApiKeyService.get(String key)` was invoked  and another time when `ApiKeyAuthenticationHelper.get(String key)` was invoked. Consecutive method calls from both `ApiKeyService` and `ApiKeyAuthenticationHelper` used the cache, though.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/auth/ApiKeyAuthenticationHelper.java
```diff
@@ -0,0 +1,72 @@
...
+import org.springframework.security.crypto.password.PasswordEncoder;
+import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Transactional;
+
+import java.util.List;
+
+/**
+ * AuthenticationHelper class.
+ * This class does require authentication.
+ * Uses simple caching, using ConcurrentHashMaps in memory.
```
jansu76, line 41: Should this comment about caching be removed, since this class does not implement any caching? ApiKeyService also does not have this comment.

petkivim: Yes, it should. It's now removed.

PR #452, opened 2020-04-07
XRDDEV-960 & XRDDEV-955

TJaakkola:
- Api for fetching service clients 
- UI for listing service clients

NOTE that XRDDEV-803 must be merge to develop before this


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
```diff
@@ -113,18 +115,20 @@
     private final SubjectConverter subjectConverter;
     private final TokenCertificateConverter tokenCertificateConverter;
     private final OrphanRemovalService orphanRemovalService;
+    private final ServiceClientConverter serviceClientConverter;
 
     /**
      * ClientsApiController constructor
      */
+    @SuppressWarnings("checkstyle:ParameterNumber")
```
jansu76, line 123: Probably correct reaction to this, but parameter numbers warning is probably also an indication that this controller is getting too large. That's probably true, but there's not much we can do about the controller size & parameter numbers, since it is direct result of the API design.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/AccessRightService.java
```diff
@@ -96,6 +96,30 @@ public AccessRightService(LocalGroupRepository localGroupRepository, GlobalConfF
         this.endpointService = endpointService;
     }
 
+    public List<AccessRightHolderDto> getAccessRightHoldersByClient(ClientId clientId)
```
jansu76, line 99: Would it make sense to split AccessRightService into AccessRightService and ServiceClientService? It's quite big already.

If so, it probably makes sense to do it in XRDDEV-961 branch after refactoring, though.

jansu76, line 99: Missing javadocs, I think all public (service) methods should have ones? This one might explain the base endpoint -filtering and what it means for the caller.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/AccessRightService.java
```diff
@@ -96,6 +96,30 @@ public AccessRightService(LocalGroupRepository localGroupRepository, GlobalConfF
...
+        }
+        // Filter just acls that are set to base endpoints so they are on service code level
+        List<AccessRightType> serviceCodeLevelAcls = clientType.getAcl().stream()
+                .filter(acl -> acl.getEndpoint().isBaseEndpoint())
+                .collect(Collectors.toList());
+        List<AccessRightType> distinctAccessRightTypes = distinctAccessRightTypeByXroadId(serviceCodeLevelAcls);
+        return mapAccessRightsToAccessRightHolders(clientType, distinctAccessRightTypes);
+    }
+
+    private List<AccessRightType> distinctAccessRightTypeByXroadId(List<AccessRightType> acls) {
```
jansu76, line 113: It _is_ quite clear when you read the code, and has a descriptive name, but it _might_ make sense to add a tiny comment to explain what it does?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/AccessRightService.java
TJaakkola: I did think whether I should have created ServiceClientService already in this ticket (created in XRDDEV-961 -branch) but as access right logic for endpoints was also in this service class I decided to move it also here. But I guess it would sense to move the logic to ServiceClientService as this logic is a bit different.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
TJaakkola: Yeah, I don't know if we can do anything (else than refactoring api) with this(?)

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/AccessRightService.java
TJaakkola: Added javadoc and also a brief definition of base endpoint. Also added base endpoint definition to EndpointType as it contains isBaseEndpoint-method. Don't know if this is the best place.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/AccessRightService.java
TJaakkola: Added short description

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/AccessRightService.java
TJaakkola: Part of the service methods are moved to ServiceClientService in XRDDEV-961

PR #455, opened 2020-04-09
XRDDEV-932: restore backup frontend implementation

raits:
JIRA ticket: https://jira.niis.org/browse/XRDDEV-932

### DIFF
--- a/src/proxy-ui-api/frontend/src/views/Settings/BackupAndRestore/RestoreBackupButton.vue
+++ b/src/proxy-ui-api/frontend/src/views/Settings/BackupAndRestore/RestoreBackupButton.vue
@@ -44,12 +44,12 @@ export default Vue.extend({
   },
   data() {
     return {
-      showConfirmation: false,
-      restoring: false,
+      showConfirmation: false as boolean,
+      restoring: false as boolean,
     };
   },
   methods: {
-    async restoreBackup() {
+    restoreBackup() {
       this.restoring = true;
       api
         .put(`/backups/${this.backup.filename}/restore`, {})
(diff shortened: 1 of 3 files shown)


src/proxy-ui-api/frontend/src/views/Settings/BackupAndRestore/RestoreBackupButton.vue
```diff
@@ -0,0 +1,77 @@
...
+    },
+  },
+  data() {
+    return {
+      showConfirmation: false,
+      restoring: false,
+    };
+  },
+  methods: {
+    async restoreBackup() {
```
TJaakkola, line 52: Is async necessary here?  After all await-keyword is not used.

src/proxy-ui-api/frontend/src/views/Settings/BackupAndRestore/RestoreBackupButton.vue
```diff
@@ -0,0 +1,77 @@
...
+      required: true,
+    },
+    backup: {
+      type: Object as Prop<Backup>,
+      required: true,
+    },
+  },
+  data() {
+    return {
+      showConfirmation: false,
```
TJaakkola, line 47: Variables in data-block can be typed. At least in the case of objects it can be usefull and it works at least partially with templates and other methods.

src/proxy-ui-api/frontend/src/views/Settings/BackupAndRestore/RestoreBackupButton.vue
raits: Removed async.

src/proxy-ui-api/frontend/src/views/Settings/BackupAndRestore/RestoreBackupButton.vue
raits: Added typings, good idea to always add them.

PR #461, opened 2020-04-12
XRDDEV-849

petkivim:
Remove reflecting input in response:
- Remove `path` and `message` properties from the default JSON error response.
- Add `CustomErrorAttributes` implementation.


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/domain/CustomErrorAttributes.java
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.niis.xroad.restapi.domain;
```
jansu76, line 25: Hmmm, I would say this does not belong to `org.niis.xroad.restapi.domain` package. That package does have bit of this and that currently, and is maybe not pure "domain" even now, but I think this one fits better in `org.niis.xroad.restapi.config` (since it is configuring aspects of Spring framework like most others in that package) or maybe `org.niis.xroad.restapi.exceptions`.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/domain/CustomErrorAttributes.java
```diff
@@ -0,0 +1,51 @@
...
+
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.util.Map;
+
+/**
+ * Override the default JSON error response for all exceptions
+ */
+@Component
+public class CustomErrorAttributes extends DefaultErrorAttributes {
```
jansu76, line 39: Really nice, simple and elegant way of doing this, well done.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/domain/CustomErrorAttributes.java
petkivim: Moving the `CustomErrorAttributes` class to `config` package makes sense since it's about configuring Spring.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/domain/CustomErrorAttributes.java
petkivim: 👍

petkivim:
Successful build: https://jenkins.niis.org/view/build/job/xroad-build-packages-pipeline/1626

jansu76:
Build has failed, I don't know the reason, restarted PR build: https://jenkins.niis.org/job/pr-build-pipeline/975/

PR #462, opened 2020-04-15
XRDDEV-972: Refactor security server db migrations

jhyoty:
Make it simpler to manually fix a security server install in case migration
fails due to insufficient database rights.


iluwatar:
The SonarQube analysis failed but I'm assuming https://github.com/nordic-institute/X-Road/pull/451 where jackson-databind was updated has fixed it already

PR #463, opened 2020-04-15
XRDDEV-936 diagnostics view

Riippi:
Diagnostics view.
Also timezone fix for date time filters. They used to be zulu time and not they are local time.

https://jira.niis.org/browse/XRDDEV-936


src/proxy-ui-api/frontend/src/views/Diagnostics/Diagnostics.vue
```diff
@@ -1,16 +1,216 @@
...
+} from '@/types';
+import StatusIcon from '@/components/ui/StatusIcon.vue';
 
 export default Vue.extend({
-  data: () => ({}),
+  components: {
+    StatusIcon,
+  },
+  data: () => ({
+    timestapmingServices: [] as TimestampingServiceDiagnostics[],
```
TJaakkola, line 119: typo

src/proxy-ui-api/frontend/src/views/Diagnostics/Diagnostics.vue
```diff
@@ -1,16 +1,216 @@
...
+@import '../../assets/shared';
+@import '../../assets/colors';
+@import '../../assets/tables';
+
+.diagnostic-card {
+  &:first-of-type {
+    margin-top: 40px;
+  }
+
+  width: 100%;
```
TJaakkola, line 185: Preferably group all the style rules of a single element to same place. This width: 100%; and margin-bottom: 30px; below should stay next to each other so it'll be easier to see what rules are set for this class.

src/proxy-ui-api/frontend/src/views/Diagnostics/Diagnostics.vue
```diff
@@ -1,16 +1,216 @@
...
+          <thead>
+            <tr>
+              <th class="status-column">{{$t('diagnostics.status')}}</th>
+              <th>{{$t('diagnostics.message')}}</th>
+              <th class="time-column">{{$t('diagnostics.previousUpdate')}}</th>
+              <th class="time-column">{{$t('diagnostics.nextUpdate')}}</th>
+            </tr>
+          </thead>
+          <tbody>
+            <tr v-if="globalconf">
```
TJaakkola, line 18: This applies to timestamping and ocsp responders -cards as well. Should there be some sort if no globalconf is empty for some reason?

src/proxy-ui-api/frontend/src/views/Diagnostics/Diagnostics.vue
Riippi: Fixed

src/proxy-ui-api/frontend/src/views/Diagnostics/Diagnostics.vue
Riippi: Fixed

src/proxy-ui-api/frontend/src/views/Diagnostics/Diagnostics.vue
Riippi: It might be good to have some "no data found" text or something like that, but there should be a one consistent way that the whole app uses. Currently a "empty table" is the way it's done in all other places also. This probably should be improved at some point.
I added a loading indicator on the cards so at least the user sees if it's empty or loading.

TJaakkola: Yeah, it should be improved. I've added such infotext to a couple of table views. I guess we should make a component of that.

PR #464, opened 2020-04-16
XRDDEV-961 & XRDDEV-956

TJaakkola:
- UI for viewing service clients access rights
- Backend api for fetching service client access rights
- Backend api for fetching a single service client


src/proxy-ui-api/frontend/src/views/Clients/ServiceClients/ServiceClientAccessRights.vue
```diff
@@ -0,0 +1,187 @@
...
+      <thead>
+        <tr>
+          <th>{{$t('serviceClients.serviceCode')}}</th>
+          <th>{{$t('serviceClients.title')}}</th>
+          <th>{{$t('serviceClients.accessRightsGiven')}}</th>
+          <th></th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr v-for="accessRight in accessRights">
```
Riippi, line 50: If the accessRight has an id or other unique thing it would be good to use it with v-bind:key
[vue/require-v-for-key]
Elements in iteration expect to have 'v-bind:key' directives.

src/proxy-ui-api/frontend/src/views/Clients/ServiceClients/ServiceClients.vue
```diff
@@ -1,33 +1,39 @@
...
           <th>{{$t('serviceClients.id')}}</th>
         </tr>
-        <template v-if="serviceClients.length > 0">
-          <tr v-for="sc in this.filteredServiceClients()">
-            <td>{{sc.subject.member_name_group_description}}</td>
-            <td>{{sc.subject.id}}</td>
+      </thead>
+      <template v-if="serviceClients.length > 0">
+        <tbody>
+          <tr v-for="sc in this.filteredServiceClients()" @click="showAccessRights(sc.id)">
```
Riippi, line 30: same 'v-bind:key'  thing in this loop also

src/proxy-ui-api/frontend/src/store/modules/services.ts
```diff
@@ -1,12 +1,12 @@
 import { ActionTree, GetterTree, Module, MutationTree } from 'vuex';
 import { RootState } from '../types';
-import {Endpoint, Service, ServiceClient, Subject} from '@/types';
+import {Endpoint, Service, ServiceClient} from '@/types';
```
Riippi, line 3: We don't yet have a shared prettier setup, but before that, could these imports be adjusted in the ide to have spaces like { RootState }  ?

src/proxy-ui-api/frontend/src/views/Clients/ServiceClients/ServiceClientAccessRights.vue
TJaakkola: In this case there is no id to use. I added v-bind:index.

src/proxy-ui-api/frontend/src/store/modules/services.ts
TJaakkola: fixed

src/proxy-ui-api/frontend/src/views/Clients/ServiceClients/ServiceClients.vue
TJaakkola: Fixed

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/AccessRightService.java
```diff
@@ -78,147 +79,23 @@
...
-    private final EndpointRepository endpointRepository;
     private final EndpointService endpointService;
+    private final LocalGroupService localGroupService;
 
     @Autowired
     public AccessRightService(LocalGroupRepository localGroupRepository, GlobalConfFacade globalConfFacade,
             ClientRepository clientRepository, ServiceService serviceService, IdentifierService identifierService,
-            GlobalConfService globalConfService, EndpointRepository endpointRepository,
-            EndpointService endpointService) {
+            GlobalConfService globalConfService,
```
jansu76, line 88: private final LocalGroupRepository localGroupRepository is unused

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceClientService.java
```diff
@@ -0,0 +1,241 @@
...
+                    ? dto.getLocalGroupId().toString().equals(scDto.getLocalGroupId())
+                    : dto.getXRoadId().equals(scDto.getSubjectId()))
+            .findFirst()
+            .orElseThrow(() -> {
+                String serviceClientIdentifier = dto.isLocalGroup()
+                        ? "Localgroup id: " + dto.getLocalGroupId()
+                        : "xRoadId: " + dto.getXRoadId().toShortString();
+                return new ServiceClientNotFoundException("Service client not found for ClientId: "
+                        + clientId.toShortString() + " and " + serviceClientIdentifier);
+            });
```
jansu76, line 135: Sonar claims this is not tested, maybe consider testing?

https://sonarqube.niis.org/component_measures?id=xroad&metric=new_coverage&pullRequest=464&selected=xroad%3Aproxy-ui-api%2Fsrc%2Fmain%2Fjava%2Forg%2Fniis%2Fxroad%2Frestapi%2Fservice%2FServiceClientService.java

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/dto/ServiceClientDto.java
```diff
@@ -31,10 +31,10 @@
 import java.time.OffsetDateTime;
 
 /**
- * DTO for Service and ServiceClient access rights
+ * DTO for ServiceClient data
  */
 @Data
-public class AccessRightHolderDto {
+public class ServiceClientDto {
```
jansu76, line 37: Now that I read the code, I find it pretty confusing that this dto has localGroupDescription and memberName (which is not used in ServiceClientConverter, though) but no global group description.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
```diff
@@ -527,10 +544,70 @@ private ClientType getClientType(String encodedId) {
...
+                    convertServiceClientDtos(serviceClientService.getServiceClientsByClient(clientId));
         } catch (ClientNotFoundException e) {
             throw new ResourceNotFoundException(e);
         }
         return new ResponseEntity<>(serviceClients, HttpStatus.OK);
     }
+
+    @Override
+    @PreAuthorize("hasAuthority('VIEW_CLIENT_ACL_SUBJECTS')")
+    public ResponseEntity<ServiceClient> getServiceClient(String id, String scId) {
```
jansu76, line 556: FYI old implementation gives frontend also boolean `has_services`

```
    render_json({
      :acl_subjects => read_acl_subjects(client),
      :has_services => has_services
    })

```
Frontend has some logic that uses that, it disables the wizard, or parts of wizard based on that info. Probably needs to be taken into account when doing the frontend ticket for add service client access rights?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceClientService.java
```diff
@@ -0,0 +1,241 @@
...
+                    boolean isBaseEndpoint = acl.getEndpoint().isBaseEndpoint();
+                    return iseq && isBaseEndpoint;
+                })
+                .map(acl -> ServiceClientAccessRightDto.builder()
+                        .serviceCode(acl.getEndpoint().getServiceCode())
+                        .rightsGiven(FormatUtils.fromDateToOffsetDateTime(acl.getRightsGiven()))
+                        .title(getServiceTitle(clientType, acl.getEndpoint().getServiceCode()))
+                        .build())
+                .collect(Collectors.toList());
+    }
```
jansu76, line 214: Old implementation sorts by serviceCode, should this one do the same?

```
    services.sort! do |x, y|
      x[:service_code] <=> y[:service_code]
    end
```

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/converter/ServiceClientIdentifierConverter.java
```diff
@@ -0,0 +1,101 @@
...
+import org.niis.xroad.restapi.dto.ServiceClientIdentifierDto;
+import org.niis.xroad.restapi.openapi.BadRequestException;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
+ * converter for ServiceClient and related objects
```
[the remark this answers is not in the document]
TJaakkola: What do you mean comment is off?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/dto/ServiceClientAccessRightDto.java
```diff
@@ -0,0 +1,45 @@
...
+@Builder
+@Getter
+/**
+ * Access rights are given to a specific subject, for services owned by some client.
+ * ServiceClientAccessRightDto.clientId is id of the service owner
+ * (not the subject id)
+ */
+public class ServiceClientAccessRightDto {
+    private String id;
+    private String clientId;
```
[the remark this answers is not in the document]
TJaakkola: Yes, noticed this.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/converter/ServiceClientIdentifierConverter.java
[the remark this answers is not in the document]
jansu76: Just that it is specifically converter for service client identifiers and nothing else, not "ServiceClient and related objects". 

It is copypaste from ServiceClientConverter by yours truly, I think.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
```diff
@@ -527,10 +544,70 @@ private ClientType getClientType(String encodedId) {
...
+
+    @Override
+    @PreAuthorize("hasAuthority('VIEW_ACL_SUBJECT_OPEN_SERVICES')")
+    public ResponseEntity<List<AccessRight>> getServiceClientAccessRights(String id, String scId) {
+        ClientId clientIdentifier = clientConverter.convertId(id);
+        ServiceClientIdentifierDto serviceClientIdentifierDto = serviceClientIdentifierConverter.convertId(scId);
+        List<AccessRight> accessRights = null;
+        try {
+            accessRights = accessRightConverter.convert(
+                    serviceClientService.getServiceClientAccessRights(clientIdentifier, serviceClientIdentifierDto));
```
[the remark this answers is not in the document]
TJaakkola: Refactored as suggested. Moved the converter-method to ServiceClientService.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
```diff
@@ -527,10 +544,70 @@ private ClientType getClientType(String encodedId) {
...
+
+    @Override
+    @PreAuthorize("hasAuthority('VIEW_CLIENT_ACL_SUBJECTS')")
+    public ResponseEntity<ServiceClient> getServiceClient(String id, String scId) {
+        ClientId clientIdentifier = clientConverter.convertId(id);
+        ServiceClientIdentifierDto serviceClientIdentifierDto = serviceClientIdentifierConverter.convertId(scId);
+        ServiceClient serviceClient = null;
+        try {
+            serviceClient = serviceClientConverter.convertServiceClientDto(
+                    serviceClientService.getServiceClient(clientIdentifier, serviceClientIdentifierDto));
```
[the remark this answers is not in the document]
TJaakkola: Did the same as in the other ServiceClientIdentifierDto -> XRoadId comment

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/AccessRightService.java
TJaakkola: Removed

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/dto/ServiceClientDto.java
TJaakkola: No changes done.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceClientService.java
TJaakkola: I guess it could but it's easy to do on frontend also, don't know if it is necessary to do here.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
TJaakkola: This could also be handled in the frontend as it's only empty array -check. What do you think?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
```diff
@@ -527,10 +544,70 @@ private ClientType getClientType(String encodedId) {
...
+        } catch (ClientNotFoundException | ServiceClientNotFoundException e) {
```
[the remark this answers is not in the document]
TJaakkola: Fixed.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceClientService.java
```diff
@@ -0,0 +1,241 @@
...
+    /**
+     * Get single service client
+     *
+     * @param clientId
+     * @param dto
+     * @return
+     * @throws ClientNotFoundException if client with given id is not found
+     * @throws ServiceClientNotFoundException if service client with given parameters is not found
+     */
+    public ServiceClientDto getServiceClient(ClientId clientId, ServiceClientIdentifierDto dto)
```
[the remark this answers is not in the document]
TJaakkola: Fix'd

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceClientService.java
```diff
@@ -0,0 +1,241 @@
...
+     * Base endpoint is in other words service (code) level endpoint.
+     * Each service has one base endpoint.
+     * Base endpoint has method '*' and path '**'.
+     *
+     * @param clientId
+     * @return
+     * @throws ClientNotFoundException
+     *
+     */
+    public List<ServiceClientDto> getServiceClientsByClient(ClientId clientId)
```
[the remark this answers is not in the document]
TJaakkola: Renamed Client clientId in method declarations to Client ownerId and used ClientType owner in method implementation.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
```diff
@@ -527,10 +544,70 @@ private ClientType getClientType(String encodedId) {
...
+    @Override
+    @PreAuthorize("hasAuthority('VIEW_ACL_SUBJECT_OPEN_SERVICES')")
+    public ResponseEntity<List<AccessRight>> getServiceClientAccessRights(String id, String scId) {
+        ClientId clientIdentifier = clientConverter.convertId(id);
+        ServiceClientIdentifierDto serviceClientIdentifierDto = serviceClientIdentifierConverter.convertId(scId);
+        List<AccessRight> accessRights = null;
+        try {
+            accessRights = accessRightConverter.convert(
+                    serviceClientService.getServiceClientAccessRights(clientIdentifier, serviceClientIdentifierDto));
+        } catch (ClientNotFoundException | LocalGroupNotFoundException e) {
```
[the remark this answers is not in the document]
TJaakkola: ClientNotFoundException -> ResourceNotFoundException done

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
jansu76: Not sure if it is just empty array -check (doesn't it mean a call to `/clients/{id}/service-descriptions` and then iterating ServiceDescriptions -> Services?) - but in any case yes, it is probably better to handle it in the frontend than add some extra baggage to this endpoint.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceClientService.java
jansu76: It probably is not necessary to do that in backend. Frontend is probably just as good.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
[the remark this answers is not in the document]
jansu76: I think the fix went to a bit of wrong direction. 

`ServiceClientService.getServiceClientAccessRights` should not throw `openapi.ResourceNotFoundException`, as that is reserved for openapi layer (controllers) only. Before the fix it threw

`ClientNotFoundException, LocalGroupNotFoundException`

LocalGroupNotFoundException was thrown if local group with given ID did not exist, when service client was type LOCALGROUP. Similar thing should happen if service client is GLOBALGROUP or SUBSYSTEM type, and id was something that does not exist. To fix the `FI:hörönlörö` case and to work consistently for all service client types (same end result for bad globalgroup/subsystem as for localgroup). There's at least those options mentioned in my earlier comment, to do that.

So I think ServiceClientService.getServiceClientAccessRights should throw

`ClientNotFoundException, ServiceClientNotFoundException`

and ServiceClientNotFoundException should cover all 3 types. Both exceptions then should be wrapped as ResourceNotFoundException in the controller.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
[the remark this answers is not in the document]
jansu76: See earlier comment, I meant service should continue to throw them as ClientNotFoundException and ServiceClientNotFoundException, but controller should then wrap those in ResourceNotFoundException, instead of BadRequestException. Sorry for unclear comment.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceClientService.java
```diff
@@ -0,0 +1,252 @@
...
+        if (owner == null) {
+            throw new ClientNotFoundException("Client " + ownerId.toShortString() + " not found");
+        }
+
+        // Filter just acls that are set to base endpoints so they are on service code level
+        List<AccessRightType> serviceCodeLevelAcls = owner.getAcl().stream()
+                .filter(acl -> acl.getEndpoint().isBaseEndpoint())
+                .collect(Collectors.toList());
+        List<AccessRightType> distinctAccessRightTypes = distinctAccessRightTypeByXroadId(serviceCodeLevelAcls);
+        return accessRightService.mapAccessRightsToServiceClients(owner, distinctAccessRightTypes);
```
jansu76, line 99: Just realized one problem. When listing service client(s) for one client, I think we should not return `rights_given_at`

```
    {
        "id": "85",
        "name": "Aston Martin Red Bull Racing",
        "local_group_code": "RBR",
        "service_client_type": "LOCALGROUP",
        "rights_given_at": "2020-04-15T03:46:21.297Z"
    },
```
Reason is that we really do not know when the rights were given. Aston Martin Red Bull Racing is a service client for this client's services, but it can be a service client for service A, B and C. `distinctAccessRightTypeByXroadId` picks AccessRightType given to A, B or C **by random**. It is possible that serviceclient.rights_given_at should not be `2020-04-15T03:46:21.297Z` but maybe `2019-01-01` instead. And it is also possible that returned timestamp changes randomly from request to another.

When listing service client(s) for a client, **rights_given_at should be null** since we do not know it. On the other hand when listing service client(s) the other way (`like /services/{id}/service-clients`) we **do know** the real rights_given_at.

Alternative to nulling it would be to scan through all AccessRightTypes and pick the earliest timestamp, that would then be the correct value. But I think that is unnecessary, as we do not use that in the UI.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/converter/ServiceClientIdentifierConverter.java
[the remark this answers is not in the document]
TJaakkola: Ah yes, it is your code, yes :) 

Changed to 'Converter for ServiceClient'

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceClientService.java
TJaakkola: I found a rather easy way to populate the rights_given_at -information with the earliest timestamp. IMO it's better than mapping all those values to null after they've been first populated (which is the other option)

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceClientService.java
TJaakkola: Added a test for testing searching a service client from a client that doesn't contain it.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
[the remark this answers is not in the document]
TJaakkola: Hmm.. I think that due to the changes made to passing parameters to ServiceClientService.getServiceClientAccessRights (serviceClientId is passed as XRoadId. It should be clear by then that) there shouldn't be need to handle what type of XRoadId was passed? Instead LocalGroupNotFoundException and ServiceClientNotFoundException are thrown from ServiceClientService.convertServiceClientIdentifierDtoToXroadId. Made these changes.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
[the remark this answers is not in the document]
TJaakkola: Fixed #2.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/ClientsApiController.java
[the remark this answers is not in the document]
jansu76: I could be mistaken, but I don't think ServiceClientService.convertServiceClientIdentifierDtoToXroadId solves this yet. Now `GET  /clients/{id}/service-clients/{sc_id}/access-rights` works like this

- local group does not belong to the client - no exception, empty list
- local group with given id does not exist - LocalGroupNotFoundException -> 404
- global group with given id does not exist - no exception, empty list
- subsystem with given id does not exist - no exception, empty list

Compare that to `POST /clients/{id}/service-clients/{sc_id}/access-rights` or `POST /clients/{id}/service-clients/{sc_id}/access-rights/delete`, they work like this

- local group does not belong to the client -> 404
- local group with given id does not exist -> 404
- global group with given id does not exist -> 404
- subsystem with given id does not exist -> 404

Different service client types (local group / global group / subsystem) should work consistently.

Consistent behavior between service client types and two latter endpoints could easily be done by calling this method https://github.com/nordic-institute/X-Road/blob/XRDDEV-962-remove-service-client-access-rights/src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/AccessRightService.java#L589-L613. It's on a different branch but should not be very likely to change, could get away with just a copy paste to your branch?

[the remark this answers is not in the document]
TJaakkola: That sounds better. But shouldn't method that verifies service client XRoadIds do exist reside in ServiceClientService instead of AccessRightService?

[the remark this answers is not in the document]
jansu76: Yeah, I think that is better place for it. Good point.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceClientService.java
jansu76: OK, that's nice!

PR #468, opened 2020-04-16
XRDDEV-904 Initial configuration

carohauta:
Initializing a new Security Server

JIRA: https://jira.niis.org/browse/XRDDEV-904


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/InitializationService.java
```diff
@@ -0,0 +1,315 @@
...
+     * @param securityServerCode server code for the new Security Server
+     * @param ownerMemberClass member class of the new owner member
+     * @param ownerMemberCode member code of the new owner member
+     * @param softwareTokenPin pin code for the initial software token (softToken-0)
+     * @param ignoreWarnings
+     * @throws AnchorNotFoundException if an anchor has not been imported
+     * @throws InitializationException if prerequisite check fails: if a server conf already exists OR if a
+     * software token has already been initialized
+     * @throws UnhandledWarningsException
+     * @throws WeakPinException if the pin does not meet the length and complexity requirements (if token pin policy is
```
jansu76, line 122: Had some trouble understanding `WeakPinException` and `InvalidPinException` initially. Reading the exception names and javadocs made me think "Isn't a weak pin also an invalid PIN? And what is InvalidPinException about, if not length and complexity requirements".

I guess the idea is that frontend will be able to show proper error messages based on classification to error types `invalid_pin` and `weak_pin`? And that there will likely be two different error messages? 

The difference between invalid_pin and weak_pin does not seem too obvious. Could also do some renaming and updating of javadocs to make it more clear? To me it looks like _bad characters_ in PIN is invalid_pin and _too few characters or too few character classes_ is weak_pin. But it took a few rereads to understand that. 

There's also the possibility of using one single exception class InvalidPinException and using the same metadata-based way of specifying how exactly was it invalid as is already used for weak / METADATA_PIN_MIN_LENGTH and weak / METADATA_PIN_MIN_CHAR_CLASSES. I am not sure if it would be better or worse.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/InitializationService.java
```diff
@@ -0,0 +1,315 @@
...
+            localClient.setIdentifier(clientId);
+            localClient.setClientStatus(ClientType.STATUS_SAVED);
+            localClient.setIsAuthentication(IsAuthentication.SSLAUTH.name());
+        }
+        return localClient;
+    }
+
+    /**
+     * Verify that the initialization process can proceed and that the security server has not already been
+     * initialized. This means verifying that an anchor has been imported,
```
jansu76, line 251: This method does not verify that anchor has been imported, but javadoc days it does?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServerConfService.java
```diff
@@ -62,6 +63,22 @@ public ServerConfType getServerConf() {
...
+     * Get a server conf; an existing server conf will be returned if one exists. Otherwise
+     * a new transient instance is returned.
+     * @return
+     */
+    public ServerConfType getOrCreateServerConf() {
+        ServerConfType serverConfType = null;
+        try {
+            serverConfType = getServerConf();
+        } catch (CodedException ce) {
+            // server conf doesn't exist which is fine - let's just create one
```
jansu76, line 76: Same comment as in `isServerConfInitialized` about CodedExceptions.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/SystemApiController.java
```diff
@@ -235,14 +235,36 @@ public SystemApiController(InternalTlsCertificateService internalTlsCertificateS
...
+     * @return
+     */
+    @Override
+    @PreAuthorize("hasAuthority('INIT_CONFIG')")
+    public ResponseEntity<Void> uploadInitialAnchor(Resource anchorResource) {
+        byte[] anchorBytes = ResourceUtils.springResourceToBytesOrThrowBadRequest(anchorResource);
+        try {
+            systemService.uploadInitialAnchor(anchorBytes);
+        } catch (SystemService.InvalidAnchorInstanceException | SystemService.MalformedAnchorException
+                | SystemService.AnchorAlreadyExistsException e) {
```
jansu76, line 262: Should AnchorAlreadyExistsException be a conflict?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/InitializationService.java
```diff
@@ -0,0 +1,315 @@
...
+     * instead of InitializationException.
+     * @param securityServerCode server code for the new Security Server
+     * @param ownerMemberClass member class of the new owner member
+     * @param ownerMemberCode member code of the new owner member
+     * @param softwareTokenPin pin code for the initial software token (softToken-0)
+     * @param ignoreWarnings
+     * @throws AnchorNotFoundException if an anchor has not been imported
+     * @throws InitializationException if prerequisite check fails: if a server conf already exists OR if a
+     * software token has already been initialized
+     * @throws UnhandledWarningsException
```
jansu76, line 121: Maybe a couple of words about ignoreWarnings and UnhandledWarningsException would be good, since it is a kind of complicated concept.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/InitializationService.java
```diff
@@ -0,0 +1,315 @@
...
+        if (isSoftwareTokenInitialized) {
+            metadata.add(METADATA_SOFTWARE_TOKEN_INITIALIZED);
+        }
+        if (!metadata.isEmpty()) {
+            throw new InitializationException("Error initializing security server", metadata);
+        }
+    }
+
+    /**
+     * If something goes south with the initialization
```
jansu76, line 273: nitpick: maybe instead of "going south" something more neutral, since it is a bit conversational and codebase is sort of official and open source (this was an extreme nitpick, sorry)

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/SystemService.java
```diff
@@ -268,6 +304,44 @@ public void uploadAnchor(byte[] anchorBytes) throws InvalidAnchorInstanceExcepti
...
+     */
+    public boolean isAnchorImported() {
+        boolean isGlobalConfInitialized = false;
+        try {
+            AnchorFile anchorFile = getAnchorFile();
+            if (anchorFile != null) {
+                isGlobalConfInitialized = true;
+            }
+        } catch (AnchorNotFoundException e) {
+            log.info("SystemService#isAnchorImported: could not find Global Configuration Anchor", e);
```
jansu76, line 319: Hmm, is this a misleading log entry, will it look like an error for system admins, with the stacktrace? Could it maybe be reworded?

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
```diff
@@ -4885,6 +5049,49 @@ components:
...
           maxLength: 24
+    InitialServerConf: # ok
+      type: object
+      description: security server initial configuration
+      required:
+        - member_class
+        - member_code
+        - security_server_code
+        - software_token_pin
+        - ignore_warnings
```
jansu76, line 5060: Maybe should not be required (also default value is meaningless if it is), other endpoints do not require ignore_warnings parameters.

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/IdentifierValidationRestTemplateTest.java
```diff
@@ -184,4 +184,44 @@ private void assertUpdateServiceDescriptionValidationFailure(String restServiceC
...
+    public void initialServerConf() {
+        assertInitialServerConfValidationError(HAS_COLON, "aa", "aa");
+        assertInitialServerConfValidationError(HAS_SEMICOLON, "aa", "aa");
+        assertInitialServerConfValidationError(HAS_PERCENT, "aa", "aa");
+        assertInitialServerConfValidationError(HAS_NON_NORMALIZED, "aa", "aa");
+        assertInitialServerConfValidationError(HAS_BACKSLASH, "aa", "aa");
+        assertInitialServerConfValidationError("aa", HAS_COLON, "aa");
+        assertInitialServerConfValidationError("aa", HAS_SEMICOLON, "aa");
+        assertInitialServerConfValidationError("aa", HAS_PERCENT, "aa");
+        assertInitialServerConfValidationError("aa", HAS_NON_NORMALIZED, "aa");
```
jansu76, line 197: Nicely tested validation :+1:

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/service/InitializationServiceTest.java
```diff
@@ -0,0 +1,265 @@
...
+import static org.niis.xroad.restapi.util.DeviationTestUtils.assertErrorWithMetadata;
+import static org.niis.xroad.restapi.util.DeviationTestUtils.assertWarning;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@AutoConfigureTestDatabase
+@Slf4j
+@Transactional
+@WithMockUser
+public class InitializationServiceTest {
```
jansu76, line 69: Nice effort put into testing across the board :+1:

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/InitializationService.java
```diff
@@ -0,0 +1,315 @@
...
+    /**
+     * Verify that the initialization process can proceed and that the security server has not already been
+     * initialized. This means verifying that an anchor has been imported,
+     * server conf does not exist and a software token has not yet been initialized. This method could also be
+     * changed into throwing an UnhandledWarningsException with warning metadata instead of the current
+     * InitializationException if partial init needs to be supported in the future
+     * @throws InitializationException if server conf exists OR software token is already initialized
+     */
+    private void verifyInitializationPrerequisites() throws InitializationException {
+        boolean isServerConfInitialized = serverConfService.isServerConfInitialized();
```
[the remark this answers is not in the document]
jansu76: Based on the discussion, it seems that partial init should be implemented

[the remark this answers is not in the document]
carohauta: Partial init has been implemented. 

Primary changes:

The init endpoint itself does not have any mandatory parameters anymore because the user must be able to initialize for example only the software token (while skipping other init parts such as server code and server owner). However the parameters are required for each missing part of the initialization. For example: if server owner has not been initialized, the server owner parameters are mandatory in the request, otherwise a `MissingInitParamsException` is thrown.

Also prerequisite check now throws warnings (depending on `ignore_warnings` flag) if the Security Server already has some initialized parts.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/InitializationService.java
```diff
@@ -0,0 +1,315 @@
...
+
+    /**
+     * Check the whole init status of the Security Server. The init status consists of the following:
+     * 1. is anchor imported - whether or not a configuration anchor has been imported
+     * 2. is server conf initialized - whether or not a server conf exists
+     * 3. is software token initialized - whether or not a software token exists AND
+     * it's status != TokenStatusInfo.NOT_INITIALIZED
+     * @return
+     */
+    public InitializationStatusDto isSecurityServerInitialized() {
```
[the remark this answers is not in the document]
carohauta: Fixed. Originally this returned one boolean. In the current and newest implementation this returns four different booleans in the form of a dto.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServerConfService.java
```diff
@@ -91,4 +108,30 @@ public ClientId getSecurityServerOwnerId() {
...
+     */
+    public boolean isServerConfInitialized() {
+        boolean isServerConfInitialized = false;
+        try {
+            ServerConfType serverConfType = getServerConf();
+            if (serverConfType != null) {
+                isServerConfInitialized = true;
+            }
+        } catch (CodedException ce) { // -> this is X_MALFORMED_SERVERCONF, "Server conf is not initialized!"
+            log.info("ServerConfService#isServerConfInitialized: CodedException thrown when getting Server Conf", ce);
```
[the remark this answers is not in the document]
carohauta: Fixed. The faultCode is now checked and if it is something else than X_MALFORMED_SERVERCONF -> rethrow the CodedException.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/InitializationService.java
```diff
@@ -0,0 +1,315 @@
...
+        boolean isAnchorImported = systemService.isAnchorImported();
+        boolean isServerConfInitialized = serverConfService.isServerConfInitialized();
+        boolean isSoftwareTokenInitialized = tokenService.isSoftwareTokenInitialized();
+        InitializationStatusDto initializationStatusDto = new InitializationStatusDto();
+        initializationStatusDto.setAnchorImported(isAnchorImported);
+        initializationStatusDto.setServerConfInitialized(isServerConfInitialized && isSoftwareTokenInitialized);
```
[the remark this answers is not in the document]
carohauta: This was due to the partial init not being supported. Now that it is supported we also return each individual boolean in the dto.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/InitializationService.java
carohauta: I just renamed the `InvalidPinException` to `InvalidCharactersException` and added javadocs for further explanation

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/InitializationService.java
carohauta: The anchor is verified before this. The whole prerequisite check is now different - so fixed

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServerConfService.java
carohauta: Fixed

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/SystemApiController.java
carohauta: It probably should! Changed

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/InitializationService.java
carohauta: Added some better explanations in javadocs

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/InitializationService.java
carohauta: Did not even think about this. Fixed.

Checking from Wiktionary, the idiom seems to have quite a few different meanings - some of which have darker or otherwise inappropriate tone to them.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/SystemService.java
carohauta: Removed the logging since there is really nothing to log here. Just swallow the exception and continue

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
carohauta: Due to the partial init refactoring there are not required parameters anymore in this action (including the `ignore_warnings` flag)

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/IdentifierValidationRestTemplateTest.java
carohauta: Also seemed to (randomly, not always) exhaust the default 20 requests per second IP throttling rate. Increased the rate to 60/second for tests.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/InitializationService.java
jansu76: :)

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/IdentifierValidationRestTemplateTest.java
jansu76: Oh, did not realize that could become an issue. Good change then. Hopefully that wont randomly fail even with the increased limit. Might also think about removing throttle for the tests altogether.

carohauta:
There has been some various dependency issues with Gradle today. Started a manual build for this PR: https://jenkins.niis.org/view/api-based-ui/job/rest-ui-build-pull-request/166/console

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/IdentifierValidationRestTemplateTest.java
jansu76: throttle filter was disabled altogether from tests by @carohauta

PR #472, opened 2020-04-22
XRDDEV-17 certificate validate

martensoo:


src/serverconf/src/main/java/ee/ria/xroad/common/conf/serverconf/IsAuthentication.java
```diff
@@ -90,10 +93,18 @@ public static void verifyClientAuthentication(ClientId client,
...
-                        "Client (%s) TLS certificate does not match any"
-                                + " IS certificates", client);
+            X509Certificate isCert = isCerts.stream()
+                    .filter(c -> Objects.equals(c, auth.getCert()))
+                    .findFirst()
+                    .orElseThrow(() -> new CodedException(X_SSL_AUTH_FAILED,
+                    "Client (%s) TLS certificate does not match any"
+                                    + " IS certificates", client));
+            try {
+                isCert.checkValidity();
```
jhyoty, line 103: Since the original `isCerts.contains(auth.getCert())` already verifies that the cert _is_ one of the IS certs, iterating the list is unnecessary. One could after that test safely check the validity of the received certificate (`auth.getCert().checkValidity()`). 
I'd also argue that it would be more readable (and perhaps a bit more efficient) than using a stream.

martensoo: Done

src/serverconf/src/main/java/ee/ria/xroad/common/conf/serverconf/IsAuthentication.java
```diff
@@ -89,11 +92,13 @@ public static void verifyClientAuthentication(ClientId client,
                 throw new CodedException(X_SSL_AUTH_FAILED,
                         "Client (%s) has no IS certificates", client);
             }
-
-            if (!isCerts.contains(auth.getCert())) {
-                throw new CodedException(X_SSL_AUTH_FAILED,
-                        "Client (%s) TLS certificate does not match any"
-                                + " IS certificates", client);
```
jhyoty, line 96: The "does the cert exist" check is missing.

martensoo: Should be back now

doc/Manuals/ug-syspar_x-road_v6_system_parameters.md
```diff
@@ -262,7 +263,7 @@ This chapter describes the system parameters used by the components of the X-Roa
 | server-conf-client-cache-size                    | 100                                        |   |   | Maximum number of local clients to keep cached |
 | server-conf-service-cache-size                   | 1000                                       |   |   | Maximum number of services to keep cached |
 | server-conf-acl-cache-size                       | 100000                                     |   |   | Maximum number of access rights to keep cached in memory. |
-
+| enforce-certificate-validity-period-check        | false                                      |   |   | Whether to throw an exception about expired or not yet valid certificates. |
```
jhyoty, line 266: Perhaps the parameter name should be ~ `enforce-client-is-certificate-validity-period` because there are also other certificates this setting does not control. The explanation could also be a bit more abstract, something like `whether to reject a request when client information system certificate is expired or not yet valid`; throwing and exception is an implementation detail.

martensoo: Changed the name

PR #475, opened 2020-04-24
XRDDEV-958

TJaakkola:
- UI for adding access rights to a service client

Must not be merged before XRDDEV-963


src/proxy-ui-api/frontend/src/views/Clients/ServiceClients/ServiceClientAccessRights.vue
```diff
@@ -99,36 +108,83 @@ export default Vue.extend({
...
       // NOOP
     },
-    removeAll() {
+    showAddServiceDialog(): void {
+      this.isAddServiceDialogVisible = true;
+    },
+    removeAll(): void {
       // NOOP
     },
+    serviceCandidates(): any[] {
```
Riippi, line 166: I suppose that there is no interface in the types for the object that this method returns.
Would it make sense to create an interface that extends AccessRight just for the ui ?
If so, then this might be a good place for it
https://github.com/nordic-institute/X-Road/blob/develop/src/proxy-ui-api/frontend/src/ui-types.ts

src/proxy-ui-api/frontend/src/views/Clients/ServiceClients/ServiceClientAccessRights.vue
```diff
@@ -99,36 +108,83 @@ export default Vue.extend({
...
+        .catch( (error: any) => this.$store.dispatch('showError', error));
 
+      api
+        .get(`/clients/${this.id}/service-descriptions`)
+        .then( (response: any) => {
+          this.serviceDescriptions = response.data;
+        })
+        .catch( (error: any) => this.$store.dispatch('showError', error));
+    },
+    fetchAccessRights() {
```
Riippi, line 133: return type void ?

src/proxy-ui-api/frontend/src/views/Clients/ServiceClients/ServiceClientAccessRights.vue
```diff
@@ -99,36 +108,83 @@ export default Vue.extend({
     return {
       accessRights: [] as AccessRight[],
       serviceClient: {} as ServiceClient,
+      isAddServiceDialogVisible: false as boolean,
+      serviceDescriptions: [] as ServiceDescription[],
+      services: [] as Service[],
     };
   },
   methods: {
     fetchData() {
```
Riippi, line 117: return type void ?

src/proxy-ui-api/frontend/src/views/Clients/ServiceClients/AddServiceClientServiceDialog.vue
```diff
@@ -0,0 +1,115 @@
...
+        this.clear();
+      },
+      cancel(): void {
+        this.$emit('cancel');
+        this.clear();
+      },
+      clear(): void {
+        this.selections = [];
+      },
+      searchResults(): any {
```
Riippi, line 91: is there a reason to use any?

Riippi: also same one line down here

TJaakkola: serviceCandidates is a propvalue. I haven't found a working way to type them, is there such a way?

src/proxy-ui-api/frontend/src/views/Clients/ServiceClients/ServiceClientAccessRights.vue
TJaakkola: No such file in my branch so I can't use it. Added a ServiceCandidate-type directly to ServiceClientAcccessRights.vue

src/proxy-ui-api/frontend/src/views/Clients/ServiceClients/ServiceClientAccessRights.vue
TJaakkola: added.

src/proxy-ui-api/frontend/src/views/Clients/ServiceClients/ServiceClientAccessRights.vue
TJaakkola: added.

src/proxy-ui-api/frontend/src/views/Clients/ServiceClients/AddServiceClientServiceDialog.vue
TJaakkola: Just marked the return type to ServiceCandidate[]

Riippi: Good question. I don't know a good way to do that either.

raits: You can use Prop<T> so this would be:
type: Array as Prop<ServiceCandidate[]>

TJaakkola: Hmm, I think I tried at some point. Don't remember how but it didn't work as well typings in data-block. Anyway I'll give it another shot...

TJaakkola: Ah I guess I tried something similar but it didn't work. But this does 👍

PR #480, opened 2020-04-29
XRDDE-957: Remove access rights from a service client

TJaakkola:
DO NOT MERGE BEFORE XRDDEV-962


src/proxy-ui-api/frontend/src/views/Clients/ServiceClients/ServiceClientAccessRights.vue
```diff
@@ -110,24 +122,47 @@ export default Vue.extend({
...
         .catch( (error: any) =>
           this.$store.dispatch('showError', error));
-
     },
     close() {
       this.$router.go(-1);
     },
-    remove() {
-      // NOOP
+    remove(accessRight: AccessRight) {
```
Riippi, line 138: Add missing return types to methods. 
In other respects frontend is ok.

TJaakkola: Fixed

PR #481, opened 2020-04-29
Xrddev 850 Csrf improvement

carohauta:
Adds additional CSRF token validation. Also changes some filter ordering.

JIRA: https://jira.niis.org/browse/XRDDEV-850


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/auth/securityconfigurer/ApiWebSecurityConfigurerAdapter.java
```diff
@@ -92,42 +93,12 @@ protected void configure(HttpSecurity http) throws Exception {
...
-                cookieCsrfTokenRepository.saveToken(token, request, response);
-            }
-        }
-
-        @Override
-        public CsrfToken loadToken(HttpServletRequest request) {
-            return cookieCsrfTokenRepository.loadToken(request);
-        }
+    private static boolean isSessionCookieAuthenticated(HttpServletRequest request) {
+        return request.getSession(false) != null;
```
jansu76, line 101: I don't remember why I implemented the original `isSessionCookieAuthenticated` as `WebUtils.getCookie(request, "JSESSIONID") != null;` instead. It is a little bit different, and I think you said original one did not work in some situations properly, which seems curious. 

I had some thoughts about renaming the method since it is not 100% about cookies (and session id could be in path params) but maybe it is more clear the way it already is.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/auth/securityconfigurer/CookieAndSessionCsrfTokenRepository.java
```diff
@@ -0,0 +1,153 @@
...
+import javax.servlet.http.Cookie;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpSession;
+
+import java.lang.reflect.Method;
+import java.util.UUID;
+
+/**
+ * Implements (and copies) functionalities from two csrf repositories:
```
jansu76, line 42: It would be nice to avoid copying `HttpSessionCsrfTokenRepository` and `CookieCsrfTokenRepository` and instead create / inject instances of those, and delegate method calls to them appropriately.

I am not sure if that can be achived. There was some problem about `if (!StringUtils.isEmpty(tokenValue)) {` in `saveToken` in some situation X (null tokenValue or something). Both `HttpSessionCsrfTokenRepository.saveToken` and  `CookieCsrfTokenRepository.saveToken` do handle situation X (null tokenValue? or are they not called in situation X?) when used by themselves alone, so maybe it would be possible to implement `CookieAndSessionCsrfTokenRepository.saveToken` so that it calls (or does not call) both existing tokenRepositories with correct parameters in situation X?

Leveraging existing Spring classes would be great since there would be less code for us to maintain, and probably less risk of having updates to HttpSessionCsrfTokenRepository and CookieCsrfTokenRepository which could be out of sync with our copied code.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/auth/securityconfigurer/CookieAndSessionCsrfTokenRepository.java
```diff
@@ -0,0 +1,153 @@
...
+            response.addCookie(cookie);
+        }
+    }
+
+    /**
+     * The de facto token that gets loaded from the session
+     */
+    @Override
+    public CsrfToken loadToken(HttpServletRequest request) {
+        HttpSession session = request.getSession(false);
```
jansu76, line 113: We need our custom `CsrfValidationFilter` since the original `CsrfFilter` only verifies the one CsrfToken returned from `loadToken` here. So CsrfFilter only checks the value from **session attribute** and compares it to value from **request header**.

If we changed loadToken so that it 
- also loads value from **cookie**
- compares it to value from session attribute
- fails with `accessDeniedHandler.handle()` if those do not match

That way CsrfFilter + CookieAndSessionCsrfTokenRepository would be verifying that cookie == header == session attribute.

That way we could eliminate the custom CsrfValidationFilter and just use the original CsrFilter?

It is maybe a slight deviation of what you could expect from the players (repository stores and loads token, CsrfFilter validates token), but I did not see any hard rules in javadocs that we would be breaking, and it is also kind of logical; if repository stores tokens into two places, and loads only one value, it makes sense that it validates that value 1 == value 2 when loading one of the values.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/auth/securityconfigurer/CookieAndSessionCsrfTokenRepository.java
```diff
@@ -0,0 +1,153 @@
...
+import java.lang.reflect.Method;
+import java.util.UUID;
+
+/**
+ * Implements (and copies) functionalities from two csrf repositories:
+ * {@link org.springframework.security.web.csrf.HttpSessionCsrfTokenRepository HttpSessionCsrfTokenRepository}
+ * and {@link org.springframework.security.web.csrf.CookieCsrfTokenRepository CookieCsrfTokenRepository}.
+ * This way we get the same token in session and csrf cookie
+ */
+public class CookieAndSessionCsrfTokenRepository implements CsrfTokenRepository {
```
jansu76, line 47: Since this is changing some "core" aspects of CSRF handling in Spring, it would be great to have a test that verifies that CSRF handling works as expected (checking positive and negative cases for header / cookie / session match). I'm mainly thinking about protection against updates to Spring CSRF handling that could be somehow incompatible with our custom implementation, and e.g. stop CSRF checks from working, and it might be hard to detect (if it just e.g. silently ignores all CSRF headers or something)

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/auth/securityconfigurer/ApiWebSecurityConfigurerAdapter.java
carohauta: Renamed the method

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/auth/securityconfigurer/CookieAndSessionCsrfTokenRepository.java
carohauta: Tests added

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/auth/securityconfigurer/CookieAndSessionCsrfTokenRepository.java
carohauta: Basically all the validation logic was moved from the filter to this repository. If the validation fails, `loadToken` just returns `null` which causes 403 Forbidden.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/auth/securityconfigurer/CookieAndSessionCsrfTokenRepository.java
carohauta: I was able to change the implementation to just a wrapper class for both repositories.

jansu76: Very nice! :+1: :1st_place_medal:

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/config/LimitRequestSizesFilter.java
```diff
@@ -48,12 +47,13 @@
 import java.io.InputStreamReader;
 
 @Configuration
-@Order(Ordered.HIGHEST_PRECEDENCE + 1)
+@Order(LimitRequestSizesFilter.LIMIT_REQUEST_SIZES_FILTER_ORDER)
```
jansu76, line 50: Nice improvements :+1:

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/auth/CsrfWebMvcTest.java
```diff
@@ -0,0 +1,181 @@
...
+                .andExpect(content().json(expectedUserJsonString));
+    }
+
+    /**
+     * Test getting user data for the currently logged in user. Uses mocked authentication in a mock session.
+     * XSRF header value should not match
+     * @throws Exception
+     */
+    @Test
+    public void getUserCsrfHeaderFail() throws Exception {
```
jansu76, line 133: Nice to see you found a good way to implement these tests :+1: Might even be a topic worth of a short developer blog, informal "share the knowledge" video session, or whatever....

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/auth/securityconfigurer/CookieAndSessionCsrfTokenRepository.java
jansu76: Nice :+1:

PR #485, opened 2020-04-30
XRDDEV-1058

petkivim:
- Update error handling when status of a sign cert is resolved.
- Catch errors and set sign cert status to false.


petkivim:
Successful build: https://jenkins.niis.org/job/pr-build-pipeline/971/

PR #497, opened 2020-05-12
XRDDEV-944: Fixes to service description detail views

TJaakkola:
JIRA: https://jira.niis.org/browse/XRDDEV-944


src/proxy-ui-api/frontend/src/views/Clients/Services/Services.vue
```diff
@@ -153,6 +153,7 @@ import ServiceIcon from '@/components/ui/ServiceIcon.vue';
 
 import { cloneDeep } from 'lodash';
 import {Service, ServiceDescription} from '@/types';
+import {ServiceTypeEnum} from '@/domain';
```
Riippi, line 156: Seems that it's been here before this PR, but it looks a bit ugly when there are two different styles of spacing. Should be { xx } not {xx}.

src/proxy-ui-api/frontend/src/views/ServiceDescriptionDetails/ServiceDescriptionDetails.vue
```diff
@@ -130,6 +140,7 @@ import ConfirmDialog from '@/components/ui/ConfirmDialog.vue';
 import WarningDialog from '@/components/service/WarningDialog.vue';
 import LargeButton from '@/components/ui/LargeButton.vue';
 import { ServiceDescription } from '@/types';
+import {ServiceTypeEnum} from '@/domain';
```
Riippi, line 143: same { xx } thing here

TJaakkola: Fixed.

src/proxy-ui-api/frontend/src/views/Clients/Services/Services.vue
TJaakkola: I did the 'fix'. But in general I disagree with this change. I think this should be done by prettier and if we're missing it for now(?) currently it will fix it when it is configured, right? So basically this manual work will anyways be done automatically later on.

Riippi: Ok, if it's manual then I'll skip these in the future. I assumed that there would be a setting or "format all" kind of solution for this the itelliJ.

PR #500, opened 2020-05-14
XRDDEV-1027 adding duplicate access rights for a localgroup is prevented.

TJaakkola:
This was actually done already in a refactoring, but I added a couple of tests just for this.

https://jira.niis.org/browse/XRDDEV-1027

### DIFF
--- a/src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/ServicesApiControllerIntegrationTest.java
+++ b/src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/ServicesApiControllerIntegrationTest.java
@@ -516,13 +516,13 @@ public class ServicesApiControllerIntegrationTest {
         // try adding two identical localgroups -> no exception, only one added
         List<ServiceClient> itemsBefore =
                 servicesApiController.getServiceServiceClients(TestUtils.SS1_GET_RANDOM_V1).getBody();
-        ServiceClient localGroup = new ServiceClient().id(TestUtils.DB_LOCAL_GROUP_ID_1).serviceClientType(
+        ServiceClient localGroup = new ServiceClient().id(TestUtils.DB_LOCAL_GROUP_ID_2).serviceClientType(
                 ServiceClientType.LOCALGROUP);
         ServiceClients duplicateLocalGroups = new ServiceClients().addItemsItem(localGroup).addItemsItem(localGroup);
         servicesApiController.addServiceServiceClients(TestUtils.SS1_GET_RANDOM_V1, duplicateLocalGroups);
         List<ServiceClient> itemsAfter =
                 servicesApiController.getServiceServiceClients(TestUtils.SS1_GET_RANDOM_V1).getBody();
-        assertEquals(itemsBefore.size(), itemsAfter.size());
+        assertEquals(itemsBefore.size() + 1, itemsAfter.size());
     }
 
     @Test(expected = BadRequestException.class)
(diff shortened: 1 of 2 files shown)


src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/ServicesApiControllerIntegrationTest.java
```diff
@@ -484,22 +484,45 @@ public void addAccessRights() {
         assertEquals(calculatePrimeClientsBefore + 3, updatedServiceClients.size());
     }
 
-    @Test(expected = ConflictException.class)
```
carohauta, line 487: Missing `@Test` annotation so the test does not actually run

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/ServicesApiControllerIntegrationTest.java
```diff
@@ -484,22 +484,45 @@ public void addAccessRights() {
...
 
+        // add subsystem TestUtils.CLIENT_ID_SS2 as duplicate
         ServiceClients clientsToAdd = new ServiceClients()
                 .addItemsItem(new ServiceClient().id(TestUtils.DB_LOCAL_GROUP_ID_2).serviceClientType(
                         ServiceClientType.LOCALGROUP))
                 .addItemsItem(new ServiceClient().id(TestUtils.CLIENT_ID_SS2).serviceClientType(
                         ServiceClientType.SUBSYSTEM));
+        try {
+            servicesApiController.addServiceServiceClients(TestUtils.SS1_GET_RANDOM_V1, clientsToAdd);
+            throw new Exception("Should throw Conflict exception in stead of this");
```
carohauta, line 503: Could use `fail()` assertion here instead of throwing a plain Exception. It would better indicate that there was an actual assertion error and not an application error.

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/ServicesApiControllerIntegrationTest.java
```diff
@@ -484,22 +484,45 @@ public void addAccessRights() {
...
+        } catch (ConflictException e) { }
 
-        servicesApiController.addServiceServiceClients(TestUtils.SS1_GET_RANDOM_V1, clientsToAdd);
+        // try adding duplicate local group
+        ServiceClients existingLocalGroup = new ServiceClients()
+                        .addItemsItem(new ServiceClient().id(TestUtils.DB_LOCAL_GROUP_ID_1).serviceClientType(
+                        ServiceClientType.LOCALGROUP));
+        try {
+            servicesApiController.addServiceServiceClients(TestUtils.SS1_GET_RANDOM_V1, existingLocalGroup);
+            throw new Exception("Should throw Conflict exception in stead of this");
```
carohauta, line 512: Same as above 👆

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/ServicesApiControllerIntegrationTest.java
TJaakkola: Oops. Added.

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/ServicesApiControllerIntegrationTest.java
TJaakkola: Changed.

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/ServicesApiControllerIntegrationTest.java
TJaakkola: Changed.

PR #518, opened 2020-05-26
XRDDEV-1102: fix anchor preview and upload to use new API endpoints

raits:
Bug ticket: https://jira.niis.org/browse/XRDDEV-1102


raits:
Rebuild succeeded

PR #521, opened 2020-05-27
XRDDEV-68

martensoo:
Initial version


src/common-verifier/src/main/java/ee/ria/xroad/common/ocsp/OcspVerifier.java
```diff
@@ -184,11 +192,12 @@ private synchronized SingleResp verifyResponseValidityCached(OCSPResp response,
                                                                  X509Certificate issuer)
             throws Exception {
         String key = SINGLE_RESP + response.hashCode() + subject.hashCode() + issuer.hashCode();
-        if (!CACHE.isValid(key)) {
-            CACHE.setValue(key, verifyResponseValidity(response, subject, issuer));
+        SingleResp responseValidity = RESPONSE_VALIDITY_CACHE.getIfPresent(key);
+        if (responseValidity == null) {
+            responseValidity = verifyResponseValidity(response, subject, issuer);
+            RESPONSE_VALIDITY_CACHE.put(key, responseValidity);
```
jhyoty, line 198: Suggestion: use the cache.get(key, loader). It also makes it possible to remove the `synchronized` modifier. something like
```
try {
   final String key = ...
   return RESPONSE_VALIDITY_CACHE.get(key, () -> verifyResponseValidity(response, subject, issuer));
 } catch (ExecutionException | UncheckedExecutionException e) {
   throw (Exception)e.getCause();
 }
```

src/common-verifier/src/main/java/ee/ria/xroad/common/conf/globalconfextension/GlobalConfExtensions.java
```diff
@@ -37,25 +37,22 @@
...
-    private static final ThreadLocal<GlobalConfExtensions> INSTANCE = new ThreadLocal<GlobalConfExtensions>() {
-        @Override
-        protected GlobalConfExtensions initialValue() {
-            return new GlobalConfExtensions();
-        }
-    };
-
     private OcspNextUpdate ocspNextUpdate;
     private OcspFetchInterval ocspFetchInterval;
+    private static GlobalConfExtensions instance = null;
```
iluwatar, line 42: I think it should be kept `volatile`. It's similar case in class GlobalConf. See also https://stackoverflow.com/questions/16542402/lazy-initialization-singleton-class-with-volatile-variable

src/common-verifier/src/main/java/ee/ria/xroad/common/conf/globalconfextension/GlobalConfExtensions.java
```diff
@@ -37,25 +37,22 @@
...
-        GlobalConfExtensions configuration = INSTANCE.get();
-        return configuration;
+        if (instance == null) {
+            synchronized (GlobalConfExtensions.class) {
+                if (instance == null) {
+                    instance = new GlobalConfExtensions();
+                }
+            }
+        }
+        return instance;
```
jhyoty, line 55: These changes are not sound since the GlobalConfExtensions instance is not thread-safe, and now it is potentially shared by several concurrent threads. Since this task has been open quite some time, I suggest reverting these changes and creating a separate task of fixing GlobalConfExtensions.

PR #523, opened 2020-05-28
XRDDEV-1001 Validation error structure

carohauta:
JIRA: https://jira.niis.org/browse/XRDDEV-1001


src/proxy-ui-api/src/main/resources/openapi-definition.yaml
```diff
@@ -4875,7 +4875,7 @@ components:
         - ERROR_CODE_MISSING_PRIVATE_PARAMS # The downloaded global configuration did not contain private parameters
         - ERROR_CODE_UNINITIALIZED # The configuration client is initializing.
         - UNKNOWN
-    CodeWithMetadata:
+    CodeWithDetails:
       type: object
       description: object that contains a code identifier and possibly collection of
```
jansu76, line 4880: maybe add mention about validation errors here, now it mentions only code and metadata?

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
```diff
@@ -4895,6 +4895,19 @@ components:
           items:
             type: string
             format: text
+        validation_errors:
+          description: object that contains validation errors bound to their respected fields - the object's key
```
jansu76, line 4899: would mentioning "dictionary" help to make this structure more clear? I was a bit befuddled at first, not understanding how this results in `Map<String, List<String>> validationErrors` (granted, due to my lack of knowledge in openapi)

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
```diff
@@ -4895,6 +4895,19 @@ components:
...
+            represents the field where the validation error has happened and the key's value is the validation error
+          type: object
+          additionalProperties:
+            type: array
+            description: array containing the validation errors of a field
+            items:
+              type: string
+              format: text
+          example:
+            clientAdd.client.memberCode: ["NoPercent"]
```
jansu76, line 4909: I think that we should document the different validation error codes for the end users. Both our own custom errors (NoPercent, NoBackslashes, etc) and standard bean validation errors (NotNull, Size, etc - this should probably be just a link to documentation elsewhere, possible some section of Bean Validation API specification?). 

Not sure what is a good way to do this, in openapi specification, or maybe somewhere in https://github.com/nordic-institute/X-Road/blob/develop/doc/Manuals/ug-ss_x-road_6_security_server_user_guide.md#19-management-rest-apis ?

Similar documentation is missing for error codes and metadata, but we already have a ticket about that https://jira.niis.org/browse/XRDDEV-665

carohauta: I think it could be a good idea to add them into the official documents (e.g. the one you linked or https://github.com/nordic-institute/X-Road/blob/develop/doc/Manuals/ug-ss_x-road_6_security_server_user_guide.md#192-executing-rest-calls). Then we can just add a link to the document in the `validation_errors` description field.

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
carohauta: Added a mention of validation errors

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
carohauta: Tried to be more verbose about this. Also added a link to OAS3 Dictionary docs right after the `additionalProperties` keyword

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
carohauta: Added a [chapter](https://github.com/nordic-institute/X-Road/blob/XRDDEV-1001-validation-error-structure/doc/Manuals/ug-ss_x-road_6_security_server_user_guide.md#194-validation-errors) to the X-Road User Guide and a mention about it into the API definition.

PR #530, opened 2020-06-01
XRDDEV-446 eslint+prettier

Riippi:
https://jira.niis.org/browse/XRDDEV-446

- Fixed 60+ linter errors
- Most of the error fixes were trivial, but some needed a bit of work
- There are still a lot of warnings. Some of those should be fixed later and rest we need to think about.
- Also removed some unused things that came up when going through the code

These files need attention: 
src/views/Clients/ServiceClients/ServiceClientAccessRights.vue
src/views/Service/Endpoints/Endpoints.vue
src/views/ServiceDescriptionDetails/ServiceDescriptionDetails.vue


src/proxy-ui-api/frontend/src/views/Service/Endpoints/Endpoints.vue
```diff
@@ -77,8 +82,16 @@ export default Vue.extend({
   },
   computed: {
     ...mapGetters(['service']),
+
+    endpoints(): Endpoint[] {
+      const filtered = this.service.endpoints.filter((endpoint: Endpoint) => {
+        return endpoint.method === '*' && endpoint.path === '**';
```
TJaakkola, line 88: Currently endpoints view lists only base endpoint. I think this should be 

`return endpoint.method !== '*' && endpoint.path !== '**';`

src/proxy-ui-api/frontend/src/views/Clients/ServiceClients/ServiceClientAccessRights.vue
```diff
@@ -189,23 +221,42 @@ export default Vue.extend({
...
+    keyedServiceClientAccessRights() {
+      let index = 0;
+      const arrWithKeys = [];
+      for (const accessRight of this.serviceClientAccessRights) {
+        arrWithKeys[index] = accessRight as UiAccessRight;
+        arrWithKeys[index].uiKey = index;
+        index++;
+      }
+
+      return arrWithKeys;
```
TJaakkola, line 254: i think this could be simplified a bit with just a map-function something like this:

`
keyedServiceClientAccessRights(): UiAccessRight[] {
      return this.serviceClientAccessRights.map( (sca: AccessRight, index: number) => {
        return { ... sca, uiKey: index };
      }) as UiAccessRight[];
    },
`

Riippi: fixed

src/proxy-ui-api/frontend/src/views/Service/Endpoints/Endpoints.vue
```diff
@@ -77,8 +82,16 @@ export default Vue.extend({
   },
   computed: {
     ...mapGetters(['service']),
+
+    endpoints(): Endpoint[] {
+      const filtered = this.service.endpoints.filter((endpoint: Endpoint) => {
+        return !this.isBaseEndpoint(endpoint);
+      });
+
+      return filtered;
```
TJaakkola, line 91: one more thing. there's no need to create that filtered -variable. Just return the value from expression.

Riippi: fixed

PR #536, opened 2020-06-05
XRDDEV-1046 csr dropdown fix

Riippi:
https://jira.niis.org/browse/XRDDEV-1046

Fixes also a bug in saving a csr file


src/proxy-ui-api/frontend/src/store/modules/certificateSignRequest.ts
```diff
@@ -263,6 +270,27 @@ export const actions: ActionTree<CsrState, RootState> = {
...
+      .then((res) => {
+        const idArray: string[] = [];
+        res.data.forEach((client: Client) => {
+          // Create id:s without possible subsystem code
+          const partialId = `${client.instance_id}:${client.member_class}:${client.member_code}`;
+
+          if (!idArray.includes(partialId)) {
+            idArray.push(partialId);
+          }
+        });
```
TJaakkola, line 286: As we have the latest javascript & typescript language features in use here I would like to see we embrace the current way of writing idiomatic javascript. I mean that although initializing a new array with in for loop works correctly in this case I urge using arrays map-function and Set -data structure to solve creating 1:1 transformations on array of objects & removing duplicates.

TJaakkola: Current implementation is better although I though mapping id-strings would have been a separate step and that why mentioned the map-function. Adding items to set in loop makes using map futile. This is good but you could use forEach with instead of map in there.

PR #539, opened 2020-06-08
XRDDEV-1020 Error snackbar update

Riippi:
https://jira.niis.org/browse/XRDDEV-1020

- Snackbar shows the metadata if it exists
- Error code is used to pull localised text from en.json with error_code. prefix. Currently only a few exists and adding the localisations for rest of the error codes should be done later.

### DIFF
--- a/src/proxy-ui-api/frontend/src/components/ui/Snackbar.vue
+++ b/src/proxy-ui-api/frontend/src/components/ui/Snackbar.vue
@@ -62,30 +62,30 @@
       multi-line
     >
       <div class="row-wrapper">
-        <div>
-          {{ $t('error_code.' + errorObject.response.data.error.code) }}
+        <div v-if="errorCode">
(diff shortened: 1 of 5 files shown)


src/proxy-ui-api/frontend/src/components/ui/Snackbar.vue
```diff
@@ -48,36 +52,50 @@
...
       </template>
 
-      <v-btn
-        data-test="snackbar-yes-button"
-        text
-        @click="closeError()"
-        outlined
-        >{{ $t('action.close') }}</v-btn
-      >
+      <v-btn icon color="white" @click="closeError()">
```
TJaakkola, line 96: data-test missing

Riippi: fixed

PR #548, opened 2020-06-15
XRDDEV-1132

petkivim:
Fix timezone issue in diagnostics timestamps:

- Convert local time returned by admin APIs to OffsetDateTime using default system timezone as source timezone.
- Set UTC timezone for tests since otherwise default timezone of JVM is used.

JIRA issue: https://jira.niis.org/browse/XRDDEV-1132

### DIFF
--- a/src/proxy-ui-api/build.gradle
+++ b/src/proxy-ui-api/build.gradle
@@ -228,3 +228,8 @@ dependencyManagement {
         }
     }
 }
+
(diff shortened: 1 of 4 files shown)


carohauta:
Started manual PR build in Jenkins: https://jenkins.niis.org/view/api-based-ui/job/rest-ui-build-pull-request/168/console

PR #550, opened 2020-06-15
XRDDEV-1119: fix clipboard for correlation ID copy functionality

raits:
JIRA ticket: https://jira.niis.org/browse/XRDDEV-1119

Also replaced the external dependency on vue-clipboard2 with a custom function.


carohauta:
Jenkins PR build failure not related to this fix

PR #558, opened 2020-06-17
XRDDEV-1122

petkivim:
- Fix Add subsystem query
   - Add missing query parameter: `internal_search=false`
- Fix Add client query
   - Filter out all clients that have local relatives.
   - If the member owning the client or another subsystem of the same member is already present locally, the client is excluded.
- Fix bug in client search with `local_valid_sign_cert` param
  - Compare only the member part of the client identifier to member identifiers in the certificates.

JIRA issues: https://jira.niis.org/browse/XRDDEV-1122

### DIFF
--- a/src/proxy-ui-api/frontend/src/store/modules/addClient.ts
+++ b/src/proxy-ui-api/frontend/src/store/modules/addClient.ts
@@ -143,7 +143,7 @@ export const actions: ActionTree<AddClientState, RootState> = {
     // Fetch clients from backend that can be selected
     return api
       .get(
-        '/clients?exclude_local=true&member_missing_sign_cert=true&internal_search=false&show_members=false',
+        '/clients?exclude_local=true&internal_search=false&show_members=false',
       )
       .then((res) => {
         commit('storeSelectableClients', res.data);
--- a/src/proxy-ui-api/frontend/src/store/modules/addClient.ts
+++ b/src/proxy-ui-api/frontend/src/store/modules/addClient.ts
@@ -158,24 +158,18 @@ export const actions: ActionTree<AddClientState, RootState> = {
   },
 
   fetchSelectableClients({ commit }, id: string) {
-    // Fetch clients from backend that can be selected
-    return api
-      .get(
-        '/clients?exclude_local=true&internal_search=false&show_members=false',
-      )
-      .then((globalClientsRes) => {
-        // Fetch list of local clients and filter out global clients
-        // that have local relatives
-        api.get('/clients').then((localClientsRes) => {
-            commit('storeSelectableClients',
-            excludeClientsWithLocalRelatives(globalClientsRes.data, localClientsRes.data));
-          })
-          .catch((localClientsError) => {
-            throw localClientsError;
-          });
+    const globalClientsPromise = api.get('/clients?exclude_local=true&internal_search=false&show_members=false');
+    const localClientsPromise = api.get('/clients');
+    // Fetch list of local clients and filter out global clients
+    // that have local relatives
+    return Promise.all([globalClientsPromise, localClientsPromise])
+      .then((response) => {
+        const globalClients = response[0];
+        const localClients = response[1];
+        commit('storeSelectableClients', excludeClientsWithLocalRelatives(globalClients.data, localClients.data));
       })
-      .catch((globalClientsError) => {
-        throw globalClientsError;
+      .catch((error) => {
+        throw error;
       });
   },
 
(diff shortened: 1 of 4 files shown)


src/proxy-ui-api/frontend/src/store/modules/addClient.ts
```diff
@@ -143,13 +161,21 @@ export const actions: ActionTree<AddClientState, RootState> = {
...
+          })
+          .catch((localClientsError) => {
+            throw localClientsError;
+          });
       })
-      .catch((error) => {
-        throw error;
+      .catch((globalClientsError) => {
+        throw globalClientsError;
       });
```
TJaakkola, line 179: Making an async call inside anothers success-block creates what is called a callback hell. This also introduces an extra period of waiting as the latter call is created after the first has been successfully completed. 

With Promise.all we can send multiple rest calls and wait for ALL of them have completed successfully. The code sample of how this is done is below. I didn't test it but it's at least nearly there...

``` 
// Fetch list of local clients and filter out global clients
    // that have local relatives
    return Promise.all([globalClientsPromise, localClientsPromise])
      .then( (response) => {
        const globalClients = response[0];
        const localClients = response[1];
        commit('storeSelectableClients', excludeClientsWithLocalRelatives(globalClients.data, localClients.data));
      })
      .catch( error => {
        throw error;
      });

```

src/proxy-ui-api/frontend/src/store/modules/addClient.ts
```diff
@@ -134,6 +134,24 @@ export const mutations: MutationTree<AddClientState> = {
...
+    let showClient = true;
+    localClients.forEach((localClient: Client) => {
+      if (localClient.member_class === client.member_class &&
+          localClient.member_code === client.member_code &&
+          localClient.instance_id === client.instance_id) {
+        showClient = false;
+      }
+    });
+    return showClient;
+  });
```
TJaakkola, line 152: Although this code works correctly I think it could be optimized and made in more currently idiomatic javascript way by using array operation some (which is a for loop which will quit looping through if true is returned inside it. Also some will return a false or true value, false is true is not returned). 

Here's an example showing how I would try to make this to be easier to read:

`

const memberEquals = (client: Client, other: Client): boolean =>
  client.member_class === other.member_class &&
  client.member_code === other.member_code &&
  client.instance_id === other.instance_id;

// Filters out clients that have local relatives.
// If the member owning the client or another subsystem
// of the same member is already present locally,
// the client is excluded.
const excludeClientsWithLocalRelatives = (clients: Client[], localClients: Client[]): Client[] => {
  return clients.filter((client: Client) => {
    return !localClients.some( (localClient: Client) => memberEquals(localClient, client))
  });
}
`

If memberEquals (not sure if this is a good function name here) returns true then loop is short circuited and false (negated) is returned to the wrapping filter function. 

I didn't test whether this works by just replacing the existing implementation but it should.

src/proxy-ui-api/frontend/src/views/AddSubsystem/AddSubsystem.vue
```diff
@@ -255,7 +255,7 @@ export default Vue.extend({
       // Fetch selectable subsystems
       api
         .get(
-          `/clients?instance=${this.instanceId}&member_class=${this.memberClass}&member_code=${this.memberCode}&show_members=false&exclude_local=true`,
+          `/clients?instance=${this.instanceId}&member_class=${this.memberClass}&member_code=${this.memberCode}&show_members=false&exclude_local=true&internal_search=false`,
```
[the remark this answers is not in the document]
petkivim: By default `internal_search` is actually `true`:

https://github.com/nordic-institute/X-Road/blob/bc20e9a2c5a271f24dc4b7668e96234b7f615717/src/proxy-ui-api/src/main/resources/openapi-definition.yaml#L657

The original problem was that the parameter was missing and the default value (`true`) was used which led to incorrect query.

src/proxy-ui-api/frontend/src/store/modules/addClient.ts
petkivim: Thanks for the suggestion! I updated the implementation accordingly. It worked almost directly - just to had to define `globalClientsPromise` and `localClientsPromise`.

src/proxy-ui-api/frontend/src/store/modules/addClient.ts
petkivim: Thanks for the suggestion. I updated the implementation accordingly.

I think that `memberEquals` is a descriptive name for the function. However, I added some comments for clarification just in case.

PR #561, opened 2020-06-18
Disable message body logging for the Icelandic X-Road instance

gudvardur:
As per request from Icelandic X-Road, message body logging will be disabled on all security servers.


raits:
See: https://github.com/nordic-institute/X-Road/pull/632

Closing this PR.

PR #579, opened 2020-07-03
XRDDEV-1178: Add sorting for services and service clients views

TJaakkola:


carohauta:
This seems to add sorting only to frontend but the task itself is marked as a backend task. Is this still work in progress?

src/proxy-ui-api/frontend/src/views/Clients/ServiceClients/AddServiceClientAccessRightsWizard.vue
```diff
@@ -92,13 +93,13 @@ export default Vue.extend({
...
-              (curr: Service[], next: ServiceDescription) =>
-                curr.concat(...next.services),
-              [],
-            )
-            .map((service: Service) => ({
+          this.serviceCandidates = serviceDescriptions.reduce(
+            (curr: Service[], next: ServiceDescription) =>
+              curr.concat(...next.services),
+            [],
+          )
```
carohauta, line 100: Could this be simplified to:
```
...
this.serviceCandidates = serviceDescriptions.flatMap(serviceDescription => serviceDescription.services)
...
```

carohauta:
Added sorting to backend. Waiting for checks to finish

PR #580, opened 2020-07-06
XRDDEV-1179 Add clients and service clients sorting

petkivim:
- Sort clients by name (primary) and client id (secondary).
- Sort service clients by name / group description (primary) and id (secondary).
- Update unit tests.

JIRA issue: https://jira.niis.org/browse/XRDDEV-1179


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/controller/ServiceClientHelper.java
```diff
@@ -104,4 +106,27 @@ public BadRequestException wrapInBadRequestException(
...
+    /**
+     * Sort a list of ServiceClient objects using member name / group description as the primary sort key, and client id
+     * as the secondary sort key.
+     * @param clients
+     */
+    public void sortServiceClientsList(List<ServiceClient> clients) {
+        clients.sort(new Comparator<ServiceClient>() {
+            @Override
+            public int compare(ServiceClient c1, ServiceClient c2) {
+                if (c1.getName() == null) {
```
petkivim, line 119: Added null checks since it's possible that in some circumstances a client's name is not available, e.g., global conf has expired, client has been removed from global conf.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/controller/ServiceClientHelper.java
```diff
@@ -104,4 +105,27 @@ public BadRequestException wrapInBadRequestException(
         return new BadRequestException(INVALID_SERVICE_CLIENT_ID + e.getServiceClientIdentifier(),
                 new ErrorDeviation(ERROR_INVALID_SERVICE_CLIENT_ID));
     }
+
+    /**
+     * Sort a list of ServiceClient objects using member name / group description as the primary sort key, and client id
+     * as the secondary sort key.
+     * @param clients
+     */
+    public void sortServiceClientsList(List<ServiceClient> clients) {
```
jansu76, line 114: Instead of this concrete list-sorting method (and `ClientUtils.sortClientsList`) a `Comparator` implementing the comparison logic would be more reusable (can sort arrays, sorted sets etc), and maybe even a bit cleaner design. Both Comparators could then be in same package (probably org.niis.xroad.restapi.controller or org.niis.xroad.restapi.openapi - or maybe even a new org.niis.xroad.restapi.converter.comparator to signal this is part of the "converting" domain...) and so be more consistent with each other and "discoverable".

For the actual implementation, there's also alternatives https://stackoverflow.com/a/20093642/1469083 such as 
- using Comparator.comparing+thenComparing+nullsFirst builders (end result may be cleaner or messier)
- using Guava or Apache commons helpers

Although the actual implementation is pretty simple so I don't know if alternative ways are worth it.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/controller/ServiceClientHelper.java
```diff
@@ -104,4 +105,27 @@ public BadRequestException wrapInBadRequestException(
...
+        clients.sort((c1, c2) -> {
+                if (c1.getName() == null && c2.getName() == null) {
+                    return c1.getId().compareToIgnoreCase(c2.getId());
```
jansu76, line 117: I would add some tests to verify just the ordering (without any controllers, services, etc) and the logic for different null / not null cases etc. 

In fact, I would usually start with the test and then write the implementation, I find that the easiest way to reach correct result in details like "should null here be +1 or -1...?". Especially since the test should be easy to implement (building a few `ServiceClient` objects with wanted name & id fields).

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/controller/ServiceClientHelper.java
```diff
@@ -104,4 +105,27 @@ public BadRequestException wrapInBadRequestException(
...
                 new ErrorDeviation(ERROR_INVALID_SERVICE_CLIENT_ID));
     }
+
+    /**
+     * Sort a list of ServiceClient objects using member name / group description as the primary sort key, and client id
+     * as the secondary sort key.
+     * @param clients
+     */
+    public void sortServiceClientsList(List<ServiceClient> clients) {
+        clients.sort((c1, c2) -> {
```
jansu76, line 115: I am happy to see that sorting is a separate feature, and **not** e.g. included in `ServiceClientConverter.convertServiceClientDtos`. I think it keeps the role of converters more clear and retains the possibility of only converting without changing the order.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/EndpointsApiController.java
```diff
@@ -156,6 +156,7 @@ public EndpointsApiController(
         }
         List<ServiceClient> serviceClients = serviceClientConverter
```
jansu76, line 157: There are some APIs that return unsorted service client collections:

- `EndpointsApiController.addEndpointServiceClients` (response from adding a new service client to an endpoint)
- `ServicesApiController.addServiceServiceClients` (same for a service)

That's probably OK, although could consider adding the sort to them also, for consistency.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/controller/ServiceClientHelper.java
petkivim: Created a new package `org.niis.xroad.restapi.converter.comparator` and added both comparators implementing the `Comparator` interface under it. In this way they are more discoverable. However, the implementation was left as it was before.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/controller/ServiceClientHelper.java
petkivim: Added tests for both comparators that verify the comparison logic.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/EndpointsApiController.java
petkivim: Thanks, I added sorting to those endpoints too. I think it's good to try to keep the ordering consistent between different endpoints.

petkivim:
Successful build: https://jenkins.niis.org/job/pr-build-pipeline/1277/

PR #582, opened 2020-07-07
XRDDEV-639 Implement access logging

petkivim:
- Access log is written to /var/log/xroad/proxy_ui_api_access.log.
- Access log is rotated daily.
- Log entry contains X-Fowarded-For headers if present in the request.
- Otherwise the log format is aligned with the current Nginx access log.

JIRA issue: https://jira.niis.org/browse/XRDDEV-639


petkivim:
I added the `max-days` property with the value of 30 days - the access logs are retained for 30 days before they are removed. I think it's better to remove them automatically so that they do not accumulate indefinitely and reserve disk space.

carohauta:
Merged develop into this one. Waiting for checks to finish

PR #583, opened 2020-07-07
XRDDEV-1017: Make SignerClient more robust

jhyoty:
In certain conditions, SignerClient can observe timeout when Signer is restarted while a request is being made. Make SignerClient
to watch Signer's RequestProcessor actor, and in case of termination, try to re-establish connection as soon as possible. Alter the Signer startup to remove unnecessary delays.

Also update Akka to the latest patch version (2.6.6).

### DIFF
--- a/src/common-util/src/main/resources/akka-global.conf
+++ b/src/common-util/src/main/resources/akka-global.conf
@@ -1,40 +1,55 @@
 akka {
-    stdout-loglevel = "OFF"
-    loggers = ["akka.event.slf4j.Slf4jLogger"]
-    loglevel = "DEBUG"
-    logging-filter = "akka.event.slf4j.Slf4jLoggingFilter"
+  stdout-loglevel = "OFF"
+  loggers = ["akka.event.slf4j.Slf4jLogger"]
+  loglevel = "DEBUG"
+  logging-filter = "akka.event.slf4j.Slf4jLoggingFilter"
 
-    actor {
-        # for now, using java serialization
-        allow-java-serialization = true
-        warn-about-java-serializer-usage = false
+  actor {
+    # for now, using java serialization
+    allow-java-serialization = true
+    warn-about-java-serializer-usage = false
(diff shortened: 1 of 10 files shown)


src/signer-protocol/src/main/java/ee/ria/xroad/signer/protocol/SignerClient.java
```diff
@@ -65,95 +72,191 @@ private SignerClient() {
...
+        @Override
+        public void postStop() {
+            if (signerRef != null) {
+                context().unwatch(signerRef);
+            }
+            resetRequestProcessor(null);
+        }
+
+        @Override
+        public void onReceive(final Object message) {
```
jansu76, line 202: The code related to SignerClient, SignerWatcher and signer requestProcessor tracking is pretty complex. I read it for some time and was barely able to grasp what is happening, and would be a bit scared if I had to update it.

I think ideally some sort of diagram would be great aid to help understanding the implementation, so that people can later update the implementation with better confidence. Perhaps it could show things like the actual client, SignerClient, SignerWatcher, Signer, and the different messages that flow between these (Watch, Identify, ActorIdentity, Terminated).

If not a diagram, then maybe consider adding some textual comments to make it easier to understand the mechanism.

Also behavior of `correlationId` and `signerRef` might benefit from more explanation.

src/common-util/src/main/resources/akka-global.conf
```diff
@@ -1,40 +1,55 @@
...
-            }
+      advanced {
+        # Maximum serialized message size, including header data.
+        maximum-frame-size = 256 KiB
 
-            advanced {
-                # Maximum serialized message size, including header data.
-                maximum-frame-size = 256 KiB
-            }
+        # Disable compression:
```
jansu76, line 26: After the branch has been merged, it would be good to follow how long term testing, if it shows any changes in memory usage?

PR #586, opened 2020-07-08
XRDDEV-316 Ignore couple low severity vulnerabilities

carohauta:
* lodash transitive dev dependency - ignored indefinitely
* parser-yargs transitive dev dependency - ignored indefinitely


carohauta:
Waiting for checks to finish. Nothing to review here

PR #588, opened 2020-07-10
XRDDEV-1170 Any logged in user can check init status

carohauta:


carohauta:
Waiting for checks to succeed. Can be merged after that

PR #591, opened 2020-07-13
XRDDEV-651 API version

carohauta:
* Add basepath `/api/v1` into the API servers `object` ~~to all API endpoints individually and group endpoints by tags~~
* Update Swagger Parser

JIRA: https://jira.niis.org/browse/XRDDEV-651

### DIFF
--- a/src/proxy-ui-api/src/main/resources/openapi-definition.yaml
+++ b/src/proxy-ui-api/src/main/resources/openapi-definition.yaml
@@ -12,8 +12,46 @@ info:
     name: MIT
     url: https://opensource.org/licenses/MIT
 tags:
-  - name: security server
-    description: <br>Security Server admin API design
+  - name: backups
+    description: backup endpoints
+  - name: certificate-authorities
+    description: CA endpoints
+  - name: clients
+    description: clients endpoints
+  - name: diagnostics
+    description: diagnostics endpoints
+  - name: endpoints
+    description: services' endpoints related endpoints
+  - name: initialization
+    description: initialization endpoints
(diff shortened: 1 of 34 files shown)


src/proxy-ui-api/src/main/resources/openapi-definition.yaml
```diff
@@ -12,13 +12,51 @@ info:
...
+    description: token certificates endpoints
+  - name: tokens
+    description: tokens endpoints
+  - name: user
+    description: user endpoints
+  - name: xroad-instances
+    description: xroad-instances endpoints
 paths:
-  /backups:
+  /api/v1/backups:
```
jansu76, line 56: Could also consider using servers-object to hold the `api/v1` prefix? https://swagger.io/docs/specification/api-host-and-base-path/

In that case some small test would be good to verify, that we have some means of having both `api/v1` and `api/v2 `endpoints coexist in the same definition.

carohauta: Removed the `/api/v1` prefix from the individual endpoints and added it into the `servers` object. The current OpenAPI Code Generator that we are using only checks that the `servers` object exists but it does not do anything with it. The API base path is now maintained in the code instead of the API specification, so the prefix was moved into each controller class as a `@RequestMapping("/api/v1")` annotation. 

Maintaining multiple versions of an endpoint is possible by adding a new controller with `@RequestMapping("/api/v2")` annotation. In such cases one should also remember to add an overriding `servers` object into the endpoint in question in order to follow best practices.

PR #594, opened 2020-07-14
XRDDEV-1190

petkivim:
Fix error code 500 when REST endpoint is updated:

- Return error code 409 with metadata when updating endpoint fails because of another endpoint with the same signature already exists.
- `DataIntegrityViolationException` is  thrown when another endpoint with the same code, method and path already exists.
- `DataIntegrityViolationException` is catched on the controller layer because the exception is thrown when the transaction commits and therefore, it must be catched outside of the transaction's scope. It cannot be catched on the service layer that would be more logical place.
- Remove a duplicate, unsused file (`/views/Service/Endpoints/EndpointDetails.vue`)
  -  The correct file is: `/views/Service/Endpoints/Endpoint/EndpointDetails.vue` 

JIRA issue: https://jira.niis.org/browse/XRDDEV-1190


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/EndpointsApiController.java
```diff
@@ -137,6 +139,10 @@ public EndpointsApiController(
             throw new ResourceNotFoundException(NOT_FOUND_ERROR_MSG + " " + id);
         } catch (EndpointService.IllegalGeneratedEndpointUpdateException e) {
             throw new BadRequestException("Updating is not allowed for generated endpoint " + id);
+        } catch (DataIntegrityViolationException e) {
```
jansu76, line 142: Should this maybe be a ConflictException instead of bad request?

In any case I think controller should not catch DataIntegrityViolationException here

- DataIntegrityViolationException is a low level data access layer concept and _ideally_ controller layer should not be aware of such things https://confluence.niis.org/display/XRDDEV/API+backend+implementation#APIbackendimplementation-Layering
- DataIntegrityViolationException is generic in principle, so we do not know for sure that it was caused by this specific problem "Endpoint with equivalent service code, ...". OK, now it is probably 99% sure, but this is a relevant issue when you factor in future changes and nontrivial DB constraints
- it deviates from the conventions used in other controllers - they mostly catch service layer exceptions and map those to controller layer / openapi exceptions. ServicesApiController.addEndpoint handles this by

```
        } catch (EndpointAlreadyExistsException e) {
```

But, handling this on service layer is probably not as easy as catch DataIntegrityViolationException -> throw EndpointAlreadyExistsException. I believe constraints only kick in when transaction commits, and this only happens when service call ends, so we can't catch DataIntegrityViolationException in the service method? I believe the constraint needs to be validated in a similar way as in addEndpoint https://github.com/nordic-institute/X-Road/blob/df9eef12face9053deef7ed5f836f707ef34a6ff/src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceService.java#L227-L230

It has to take into account endpoint ID probably, to distinguish between updated endpoint from the rest of them.

This adds some extra work to implement the constraints, which is unfortunate. If the service layer checks were too difficult to implement then some compromises or some generic handlers could be made. But I think in this case it should not be too costly to implement the check and be more consistent with the rest of the app.

petkivim: Refactored the PR based on the comments. `ConflictException` is now thrown instead of `BadRequest`.

PR #595, opened 2020-07-14
XRDDEV-1124: fix linter warnings

raits:
Fixed type usage all over the place, added a deepClone utility method for the JSON cloning and a common FileUpload component to handle uploading files in a common manner.

JIRA ticket: https://jira.niis.org/browse/XRDDEV-1124


src/proxy-ui-api/frontend/src/store/modules/general.ts
```diff
@@ -82,8 +82,9 @@ export const actions: ActionTree<State, RootState> = {
   },
 
   fetchMemberName({ commit }, { memberClass, memberCode }) {
+    // this is currently an inline schema and is not automatically generated to a typescript type
     return api
-      .get(
+      .get<{ member_name: string }>(
```
carohauta, line 87: All OpenAPI objects now get generated into TypeScript types. For example this one is `MemberName`.

This was done in PR (already merged in develop): https://github.com/nordic-institute/X-Road/pull/563/files

raits: Fixed this

PR #604, opened 2020-07-17
XRDDEV-1186 Service desc refresh bug

carohauta:
Changes the refreshing of an OpenAPI service description so that only the acl and endpoints that are related to the _selected **service description** are affected_. 

This bug occurred because the refresh affected on all acl and endpoints of the _owning **client** of the service description_

JIRA: https://jira.niis.org/browse/XRDDEV-1186


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ServiceDescriptionService.java
```diff
@@ -772,7 +774,7 @@ public ServiceDescriptionType updateOpenApi3ServiceDescription(Long id, String u
      * @param ignoreWarnings
      * @param serviceDescription
      * @throws OpenApiParser.ParsingException if there are errors in the openapi3 description document
-     * @throws UnhandledWarningsException     if ignoreWarnings is false and parser returns warnings from openapi
+     * @throws UnhandledWarningsException if ignoreWarnings is false and parser returns warnings from openapi
      */
     private void parseOpenApi3ToServiceDescription(String url, String serviceCode, boolean ignoreWarnings,
```
jansu76, line 779: parseOpenApi3ToServiceDescription is really quite complex, it would be really great to make it simpler somehow. Could it be split into smaller parts, etc? Now it feels like a risky part that could lead to some bugs.

The fact that you can mix generated and non-generated endpoints makes it kind of unavoidable that the aggregate implementation will be complex, unfortunately.

It would have been good to add a test for the failing case (before fixing it)

PR #612, opened 2020-07-23
XRDDEV-1183 Client flow improvement

Riippi:
https://jira.niis.org/browse/XRDDEV-1183

- Add client flow handles existing csr / certificate scenarios in similar way as add member flow
- Some cleanup for add member flow

Merge PR 606 before this
https://github.com/nordic-institute/X-Road/pull/606

### DIFF
--- a/src/proxy-ui-api/frontend/src/views/AddClient/ClientDetailsPage.vue
+++ b/src/proxy-ui-api/frontend/src/views/AddClient/ClientDetailsPage.vue
@@ -120,10 +120,14 @@ import { mapGetters } from 'vuex';
 import FormLabel from '@/components/ui/FormLabel.vue';
 import LargeButton from '@/components/ui/LargeButton.vue';
 import SelectClientDialog from '@/components/client/SelectClientDialog.vue';
+import { debounce, isEmpty, containsClient } from '@/util/helpers';
 import { Client } from '@/openapi-types';
-import { containsClient } from '@/util/helpers';
-
 import { ValidationProvider, ValidationObserver } from 'vee-validate';
+import { AddMemberWizardModes } from '@/global';
+
+// To provide the Vue instance to debounce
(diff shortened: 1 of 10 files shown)


src/proxy-ui-api/frontend/src/views/AddClient/FinishPage.vue
```diff
@@ -115,21 +163,66 @@ export default Vue.extend({
...
+        .finally(() => {
           this.disableCancel = false;
           this.submitLoading = false;
-        },
+        });
+    },
+
+    registerClient(): void {
+      const clientId = createClientId(
+        this.currentSecurityServer.instanceId,
```
raits, line 206: This doesn't seem to work, API call that results from this is "/api/v1/clients/undefined:ORG:TEST/register" (instance is undefined), resulting in failure to register the client.

I think this identifier should also include the subsystem that is being added.

src/proxy-ui-api/frontend/src/views/AddClient/ClientDetailsPage.vue
```diff
@@ -180,9 +185,9 @@ export default Vue.extend({
   data() {
     return {
       disableDone: false,
-      certificationService: undefined,
       filteredServiceList: [],
```
raits, line 188: This doesn't seem to be used anymore, maybe it should be removed?

src/proxy-ui-api/frontend/src/views/AddClient/FinishPage.vue
Riippi: Fixed

src/proxy-ui-api/frontend/src/views/AddClient/ClientDetailsPage.vue
Riippi: Removed it and also another unused value

raits:
Looks good to me.

PR #623, opened 2020-07-28
XRDDEV-1060 session expired > log out user

guycalledguy:


raits:
JIRA ticket: https://jira.niis.org/browse/XRDDEV-1060

PR #628, opened 2020-07-30
XRDDEV-1160 autogenerate centerui password

iluwatar:


src/packages/src/xroad/ubuntu/generic/xroad-center.postinst
```diff
@@ -97,6 +101,8 @@ GRANT USAGE ON SCHEMA public to "${db_plain_user}";
 EOF
     fi
 
+    crudini --set "${db_properties}" '' password "${db_password}"
```
jhyoty, line 104: There should be no need to modify the db.properties in case the database is not created by the install script  --> should be moved in the previous else block (lines 85-101).

iluwatar: Yes I agree. Pushed the update.

PR #631, opened 2020-08-05
Frontend unit test fix

Riippi:
- Fixed failing test for clients vuex store
- Remove the snapshot test which was more of an example than actual test
- Refactored the vuex store imports


Riippi:
Merged without review as there are no changes to actual code.

PR #635, opened 2020-08-06
XRDDEV-1235 Fix wrong date

Riippi:
https://jira.niis.org/browse/XRDDEV-1235

Use getDate() instead of getDay()


Riippi:
Merged by myself because only two lines were changed

PR #636, opened 2020-08-07
XRDDEV-559 remove old ui

jansu76:
https://jira.niis.org/browse/XRDDEV-559


src/proxy-ui-api/build.gradle
```diff
@@ -187,6 +187,20 @@ jar {
...
+        eachFile { fcd ->
+            fcd.relativePath = new RelativePath(true, fcd.relativePath.segments.drop(2))
+        }
+        includeEmptyDirs = false
+    }
+    into "$buildDir/unpacked-libs"
+}
+
+jar.finalizedBy(unpackLibsFromBootJar)
+
```
jhyoty, line 203: suggestion (a bit simpler and also avoids copying unnecessary libs):
```
task copyDeps(type: Copy) {
    into "$buildDir/unpacked-libs"
    from configurations.compileClasspath.find {it.name.startsWith("postgresql") }
}

build.dependsOn copyDeps
```

src/proxy-ui-api/build.gradle
```diff
@@ -208,6 +222,7 @@ dependencies {
     implementation('wsdl4j:wsdl4j:1.6.3')
     implementation('com.github.vladimir-bukhtoyarov:bucket4j-core:4.10.0')
     implementation("io.swagger.parser.v3:swagger-parser:${swaggerParserVersion}")
+    implementation("org.postgresql:postgresql:42.2.14")
```
jhyoty, line 225: should be unnecessary (already a transitive dependency)

doc/Manuals/ug-syspar_x-road_v6_system_parameters.md
```diff
@@ -276,14 +276,7 @@ limits the configuration parameters usage:
...
 
-### 3.3 Proxy User Interface parameters: `[proxy-ui]`
-
-| **Parameter**                                    | **Vanilla value**                          | **Description** |
-|--------------------------------------------------|--------------------------------------------|-----------------|
-| wsdl-validator-command                           |                                            | The command to validate the given X-Road service WSDL. The command script must:<br/>a) read the WSDL from the URI given as an argument,<br/>b) return exit code 0 on success,<br/>c) return exit code 0 and write warnings to the standard error (*stderr*), if warnings occurs,<br/>d) return exit code other then 0 and write error messages to the standard error (*stderr*), if errors occurs.<br/>Defaults to no operation. |
-| auth-cert-reg-signature-digest-algorithm-id      | SHA-512                                    | Signature digest algorithm used for generating authentication certificate registration request.<br/>Possible values are<br/>-   SHA-256,<br/>-   SHA-384,<br/>-   SHA-512. |
-
-### 3.4 Signer parameters: `[signer]`
+### 3.3 Signer parameters: `[signer]`
```
jhyoty, line 279: Maybe it would be safer to keep the numbering (just leave an empty 3.3 section "proxy-ui removed in version 6.24" or something) in case there are some external links to the parameter sections.

jansu76: Fixed

src/proxy-ui-api/build.gradle
jansu76: Fixed

src/proxy-ui-api/build.gradle
jansu76: Good suggestion, fixed, much simplified.

Gradle was a bit picky about the place where `copyDeps` should be put. That might explain why I ended up with the original clumsy implementation. `copyDeps` needed to be after `dependencyManagement`, it seems.

PR #641, opened 2020-08-10
XRDDEV-1234: encode path and query parameters

raits:
JIRA ticket: https://jira.niis.org/browse/XRDDEV-1234

Added encodePathParameter utility function under api. It currently just wraps encodeUriComponents but if we ever need to change the implementation it can be done in a central place in the future. All path parameters should now be wrapped with this function call so that they are properly encoded.

Changed instances of query parameter use to use the axios configuration params object instead of template literals so that axios would handle encoding for us. The encoding method can be overwritten in the axios instance configuration if needed.

Removed usages of template literals where no values were used inside them. Fixed a few instances where path variables could have been undefined

**NB!** Definitely sanity check the fetchCsrForm() method implementation change inside src/store/modules/certificateSignRequest.ts.


TJaakkola:
src/store/modules/certificateSignRequest.ts fetchCsrForm -method seems identical to the previous one to me.

PR #646, opened 2020-08-12
Xrddev 1087 manage internal and external configurations

guycalledguy:


raits:
Will handle use-case documentation separately.

PR #650, opened 2020-08-15
XRDDEV-1248/XRDDEV-1249

petkivim:
Add warning dialog when auth key with registered cert is deleted:

- Warning dialog is shown when an auth key with a registered certificate is deleted.
- Auth cert is unregistered and key + cert deleted if the dialog is approved.
- Add loading state to the Delete key button.
- Throw an UnhandledWarningsException if auth key has a registered certificate when it's deleted.
- Add ignore_warnings URL param to delete key endpoint.
- Update OpenAPI description.
- Add unit test.

Frontend JIRA issue: https://jira.niis.org/browse/XRDDEV-1249
Backend JIRA issue: https://jira.niis.org/browse/XRDDEV-1248


Riippi:
Frontend looks ok and is good to merge, but someone else should take a look at the backend.

jansu76:
> Frontend looks ok and is good to merge, but someone else should take a look at the backend.

I will review backend

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/KeyService.java
```diff
@@ -202,15 +206,27 @@ private static String signerFaultCode(String detail) {
 
     static final String KEY_NOT_FOUND_FAULT_CODE = signerFaultCode(X_KEY_NOT_FOUND);
 
+    public void deleteKey(String keyId) throws KeyNotFoundException, ActionNotPossibleException,
+            GlobalConfOutdatedException {
+        try {
+            deleteKey(keyId, true);
+        } catch (UnhandledWarningsException e) {
+            // Since "ignoreWarnings = true", the exception is never thrown
```
jansu76, line 214: Swallowing exceptions should _almost_ never be done. Even if UnhandledWarningsException should indeed be impossible, wrapping it as RuntimeException is preferrable, in case there's some weird bug that _does_ cause UnhandledWarningsException, or if someone later changes code so that our current assumptions are not valid anymore. Even if unlikely, I've seen those two things happen, and then it may mean a long bughunt since from outside everything looks like the key delete was successful.

Adding `throw new RuntimeException(e);` here also does not have any real downside, in case we are correct and this never happens. Comment is a good one to keep IMO, to explain this to updater.

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
```diff
@@ -2154,6 +2154,14 @@ paths:
             type: string
             format: text
             minLength: 1
+        - in: query
+          name: ignore_warnings
+          description: if true, any ignorable warnings are ignored. if false (or missing),
```
jansu76, line 2159: Hmmm. I feel we might need to better explain users what the warnings related to this operation could be? Like in this case, give specifics so that user understands the warnings are about registered auth certificates?

You could also instead (or, also) explain these things in the markdown documentation, but maybe the openapi definition is a pretty good place for this?

Client add says this in definition-part:

```
      description: >
        Adds new client to the system.
        Note that with this endpoint it is possible to add an unregistered member as a client.
        Attempt to add an unregistered member with ClientAdd.ignore_warnings = false causes the operation to fail with a warning in response's ErrorInfo object.
        Attempt to add an unregistered member with ClientAdd.ignore_warnings = true succeeds.
```

But unfortunately this is not consistent, other endpoints with an `ignore_warnings` possibility do not have the same type of explanation.

This ticket could add some explanation of warnings for this endpoint, or maybe documentation for all endpoints could be updated and made consistent in a separate ticket....

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/KeysApiController.java
```diff
@@ -225,14 +226,14 @@ private Key getKeyFromService(String keyId) {
     @Override
     @PreAuthorize("hasAnyAuthority('DELETE_KEY', 'DELETE_AUTH_KEY', 'DELETE_SIGN_KEY')")
     @AuditEventMethod(event = DELETE_KEY)
-    public ResponseEntity<Void> deleteKey(String keyId) {
+    public ResponseEntity<Void> deleteKey(String keyId, Boolean ignoreWarnings) {
```
jansu76, line 229: Would be good to add javadocs to old public void `deleteKey(String keyId) {` so that it explains the ignoreWarnings = true deal. (weird, I thought checkstyle requires javadocs for public methods. It is good practice to have those anyhow)

Maybe also consider changing old method name to something like deleteKeyAndIgnoreWarnings, or removing it altogether and using deleteKey(keyId, true) instead. Does having a shorthand version, without boolean parameter, make things better or worse? Maybe without it code might be more clear and more explicit about what is the intended action. Well, not sure about that one really.

jansu76 (changes requested):

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/KeyService.java
petkivim: Good points, `RuntimeException` is now thrown in case `UnhandledWarningsException` is catched.

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
petkivim: I updated the OpenAPI definition for this endpoint. I think it's better to have all the related documentation in one place to make consuming the API easier for the users. In my opinion OpenAPI definition is the right place for this kind of documentation. I also think that it's a good idea to create a separate ticket about updating the OpenAPI definition and making it more consistent. More example messages should be added to the documentation too since some endpoints  have parameters which values are difficult to get right, e.g. endpoints related to CSR creation.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/KeysApiController.java
petkivim: Changed the old method name to `deleteKeyAndIgnoreWarnings` and updated javadocs.

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
jansu76: Created https://jira.niis.org/browse/XRDDEV-1272

jansu76 (approved):

PR #653, opened 2020-08-17
XRDDEV-1268

petkivim:
Fix error that caused alerts endpoint (`GET /api/v1/notifications/alerts`) to return HTTP 500 when connection to signer fails.

Add two new properties to Alerts endpoint: `global_conf_valid_check_success` and `soft_token_pin_entered_check_success`. After the update Alerts endpoint's response is:

```
{
  "global_conf_valid":true,
  "global_conf_valid_check_success": true,
  "soft_token_pin_entered":true,
  "soft_token_pin_entered_check_success":true
}
```

The `_check_success` properties indicate whether the check was successful and whether `global_conf_valid` / `soft_token_pin_entered` tell the real status. When `_check_success` property is `true`, we know that the current status is correct. When `_check_success` property is `false`, we know that the check failed and we cannot be sure what’s the current status.

For example, connection to signer fails:

```
{
   "global_conf_valid":true,
   "global_conf_valid_check_success":true,
   "soft_token_pin_entered":false,
   "soft_token_pin_entered_check_success":false
}
```

For example, connection to signer succeeds, but PIN code has not been entered:

```
{
   "global_conf_valid":true,
   "global_conf_valid_check_success":true,
   "soft_token_pin_entered":false,
   "soft_token_pin_entered_check_success":true
}
```

Background information: https://niis.slack.com/archives/GGBHVCYSY/p1597732119019400?thread_ts=1597402894.008800&cid=GGBHVCYSY

JIRA issue: https://jira.niis.org/browse/XRDDEV-1268


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/NotificationService.java
```diff
@@ -79,35 +73,35 @@ public AlertStatus getAlerts() {
...
      */
     private boolean isGlobalConfValid() {
-        try {
-            globalConfFacade.verifyValidity();
-            return true;
-        } catch (CodedException e) {
-            return false;
-        }
+        globalConfFacade.verifyValidity();
+        return true;
```
jansu76, line 104: This method now never returns false, since `globalConfFacade.verifyValidity` either returns true or throws CodedException? 

a `private boolean isGlobalConfValid` that either returns true or throws exception is pretty nonstandard for a Java method, would be better to return true/false in un-exceptional cases. Maybe the old implementation was better after all?

I may have posted misleading comments in the slack thread `I guess the implementation should then inspect the exception to distinguish which case it is` , as it looks like the implementation already did that (well, it did not inspect the error code vs X_OUTDATED_GLOBALCONF, so it was not completely flawless)?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/NotificationService.java
```diff
@@ -79,35 +73,35 @@ public AlertStatus getAlerts() {
...
-        } else {
+            alertStatus.setGlobalConfValidCheckSuccess(true);
+        } catch (CodedException e) {
+            log.error("getting global conf status failed");
             alertStatus.setGlobalConfValid(false);
+            alertStatus.setGlobalConfValidCheckSuccess(false);
+        }
+
+        try {
+            alertStatus.setSoftTokenPinEntered(isSoftTokenPinEntered());
```
jansu76, line 87: Should we just `setSoftTokenPinEnteredCheckSuccess(false)` if we know that backup restore is running? Since doing `isSoftTokenPinEntered` will likely fail doing restore, and possibly cause the "could not list all tokens" error to appear with some delay?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/NotificationService.java
```diff
@@ -79,35 +73,35 @@ public AlertStatus getAlerts() {
...
             alertStatus.setCurrentTime(OffsetDateTime.now(ZoneOffset.UTC));
         }
-        if (isSecurityServerFullyInitialized()) {
+
+        try {
             alertStatus.setGlobalConfValid(isGlobalConfValid());
-            alertStatus.setSoftTokenPinEntered(isSoftTokenPinEntered());
-        } else {
+            alertStatus.setGlobalConfValidCheckSuccess(true);
+        } catch (CodedException e) {
```
jansu76, line 80: Could just catch any `Exception`, probably only `CodedExceptions` should occur now, but we might not have realized a possibility of some other exceptions like RuntimeExceptions or some future changes may make that possible - and we probably want those to also result in setGlobalConfValidCheckSuccess(false)?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/NotificationService.java
```diff
@@ -47,23 +45,19 @@
...
     @Autowired
-    public NotificationService(GlobalConfFacade globalConfFacade, TokenService tokenService,
-            InitializationService initializationService) {
+    public NotificationService(GlobalConfFacade globalConfFacade, TokenService tokenService) {
         this.globalConfFacade = globalConfFacade;
         this.tokenService = tokenService;
-        this.initializationService = initializationService;
     }
 
     /**
```
jansu76, line 63: > If alerts are disabled, the status of all alerts true.

Is this comment up to date? Does "alerts are disabled" refer to `backupRestoreRunningSince`?

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/NotificationService.java
petkivim: Yes, you're right, the old implementation was better. That is now restored. The logic of the global conf check now works so that when the check throws `CodedException` the response is:

```
global_conf_valid: false
global_conf_valid_check_success: true
```

And the the check throws any other exception the response is:

```
global_conf_valid: false
global_conf_valid_check_success: false
```

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/NotificationService.java
petkivim: That makes sense. Both `soft_token_pin_entered` and `soft_token_pin_entered_check_success` are now set to `false` always when backup / restore is running, and the actual check is skipped.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/NotificationService.java
petkivim: The implementation was updated so that `Exception` is catched. It's possible that `globalConfFacade.verifyValidity()` [throws other exceptions](https://github.com/nordic-institute/X-Road/blob/develop/src/common-verifier/src/main/java/ee/ria/xroad/common/conf/globalconf/GlobalConf.java#L64) too (e.g. `RuntimeException`). Please see the first comment regarding the updated logic.

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/NotificationService.java
petkivim: The comment was updated.

jansu76 (approved):

PR #655, opened 2020-08-18
XRDDEV-1257: User with only xroad-security-officer role doesn't have …

TJaakkola:
…access to add service clients button

JIRA: https://jira.niis.org/browse/XRDDEV-1257


raits (approved):
Looks good to me

PR #656, opened 2020-08-19
XRDDEV-1267

jansu76:
Do not merge....should not go to develop

https://jira.niis.org/browse/XRDDEV-1267

- GlobalConfChecker skips checking while backup restore is running
- restore script kills open serverconf database connections before restoring
- application clears connection pool after restore


jansu76:
tried to merge to wrong branch, so closing this

PR #659, opened 2020-08-20
XRDDEV-1269 Login fail signer bug

carohauta:
Fix a bug which caused login to fail if tokens could not be queried from the signer module when checking the server init status. Also includes a change to the frontend init phase routing logic

Notice that this PR is targeted for the `beta-6.24.0` branch.

JIRA: https://jira.niis.org/browse/XRDDEV-1269


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenService.java
```diff
@@ -328,12 +328,19 @@ public TokenInfoAndKeyId getTokenAndKeyIdForCertificateHash(String hash) throws
...
-    public boolean isSoftwareTokenInitialized() {
+    public Boolean isSoftwareTokenInitialized() {
         boolean isSoftwareTokenInitialized = false;
-        List<TokenInfo> tokens = getAllTokens();
+        List<TokenInfo> tokens = null;
+        try {
+            tokens = getAllTokens();
+        } catch (Exception e) {
+            log.error("Could not get software token status from signer", e);
+            return null;
```
jansu76, line 342: This would cause NPE in InitializationService if null, is that intentional?

        boolean isSoftwareTokenInitialized = tokenService.isSoftwareTokenInitialized();

Rather than return Boolean.TRUE, Boolean.FALSE or null, maybe consider returning an enum with 3 values. Or just catch exceptions in upper layer and handle them there. I think it might be more explicit and make it less likely that someone misuses the return value.

This seems to come down to the question you asked 

> should the getAllTokens method actually throw a checked exception instead of a runtime exception if signer is unresponsive

If we want to handle the exception properly on upper layers, then the checked exception would help this a lot. I guess the reasoning behind `throw new RuntimeException("could not list all tokens")` was assumption that no one really wants to do anything else than http 500 when that happens. And now we do. 

So there's also the possibility of changing signer communication problems into checked exceptions. But not sure if that's in scope of this ticket. Or, just keep those as RuntimeExceptions and fix the NPE by maybe changing the return value to enum?

jansu76 (changes requested):

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/TokenService.java
carohauta: Created a new method that returns the state of a token as an enum value in the `TokenService`. The login now works fine and the init status api returns the SoftwareTokenInitStatus as `UNKNOWN`.

There is still one problem when getting clients from the api while signer is down: the user only gets a `{ status: 500 }` response. This happens because the clients api uses request scoped beans and if creating those beans fails the `SignerNotReachableException` gets nested in Spring's internal exceptions such as `BeanCreationException` and `BeanInstantiationException`. I'll try to include a fix for that in this PR

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/exceptions/ApplicationExceptionHandler.java
```diff
@@ -93,4 +99,24 @@
         log.error("exception caught", e);
         return exceptionTranslator.toResponseEntity(e, HttpStatus.FORBIDDEN);
     }
+
+    /**
+     * check for nested SignerNotReachable exception when BeanCreationException is caught
```
jansu76, line 104: I think it would be good to explain _why_ we do this. What is the special bond between BeanCreationExceptions and SignerNotReachableExceptions? Otherwise some future developer will be confused.

jansu76 (changes requested):

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/exceptions/ApplicationExceptionHandler.java
carohauta: Added better documentation here

jansu76 (approved):

PR #665, opened 2020-08-24
XRDDEV-1262

TJaakkola:


raits (commented):
I think that this isn't a complete fix. Under the "created" lifecycle hook, there is the old check for fetching the approved certificate authorities:
```
if (this.hasPermission(Permissions.GENERATE_AUTH_CERT_REQ)) {
   this.fetchApprovedCertificateAuthorities();
}
```

This probably needs to be changed to also check for the new permission. However, when we change this, we get a 403 error, because the same permission isn't being used on the backend.

I suspect that the permission check on this endpoint also needs to be updated:
https://github.com/nordic-institute/X-Road/blob/a28e73114dda404da1b85b490ec9ad968b1cc71f/src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/CertificateAuthoritiesApiController.java#L102

TJaakkola:
Yes, I missed that fetching part of the logic. Now in addition to initial roles that can fetch certificate authorities also users with the created permission can fetch the data.

raits (approved):
Looks good to me.

PR #673, opened 2020-08-26
XRDDEV-1288 better file upload max size configuration

jansu76:
https://jira.niis.org/browse/XRDDEV-1288

Also fixes https://jira.niis.org/browse/XRDDEV-1289


carohauta (approved):
🦄

PR #674, opened 2020-08-27
XRDDEV-401 Input validation in api

carohauta:
JIRA: https://jira.niis.org/browse/XRDDEV-401

This only includes the minor findings such as 
* increase service id max length
* endpoint id max length
* security-server id max and min lengths

New tickets will be created out of the other findings; check the JIRA task comments


jansu76 (approved):
looks good

PR #678, opened 2020-09-01
XRDDEV-680 OpenAPI validation

carohauta:
Adds an OpenAPI validator Gradle Plugin to the project. Check confluence for additional info about the validator and the reasons/tech behind it: https://confluence.niis.org/display/XRDDEV/OpenAPI+3+Validation

JIRA: https://jira.niis.org/browse/XRDDEV-680


src/buildSrc/src/main/java/org/niis/xroad/oasvalidatorplugin/Oas3ValidatorGradlePlugin.java
```diff
@@ -0,0 +1,78 @@
...
+                    extension.getApiDefinitionPaths().forEach(path -> {
+                        try {
+                            ApiValidationResults validationResults = Oas3Validator.validate(path);
+                            allValidationResults.add(validationResults);
+                        } catch (ResolutionException e) {
+                            throw new GradleException("API definition not found", e);
+                        }
+                    });
+                    for (ApiValidationResults apiValidationResults : allValidationResults) {
+                        ApiValidationResult apiSpecValidationResult = apiValidationResults
```
jansu76, line 65: There seem to be some validation error cases (duplicate property or required -tag in openapi-authentication.yaml) that result in an exception that is printed out, but validation result is SUCCESS and build continues.

Possibly caused by a JsonParseException that is thrown out instead of GradleException. Not sure, but would it be a good idea to catch all exceptions that happen during validation and wrapping them in GradleException("API definition validation failed", rootCause), or something?

jansu76 (changes requested):

src/buildSrc/src/main/java/org/niis/xroad/oasvalidatorplugin/Oas3ValidatorGradlePlugin.java
carohauta: Pushed a fix for this. All Exceptions are now caught in the Plugin. The JsonParseException was not actually thrown but only logged by one of the parser libraries. But the data was still recoverable and is now added to the validation failure data.

jansu76 (approved):

PR #683, opened 2020-09-03
XRDDEV-1245 fix internal TLS cert related permissions, remove duplicates

jansu76:
https://jira.niis.org/browse/XRDDEV-1245


jansu76:
wrong target

PR #686, opened 2020-09-04
XRDDEV-1315 Fix router duplicate navigation console alerts

Riippi:
https://jira.niis.org/browse/XRDDEV-1317

### DIFF
--- a/src/proxy-ui-api/frontend/src/components/layout/AppToolbar.vue
+++ b/src/proxy-ui-api/frontend/src/components/layout/AppToolbar.vue
@@ -80,7 +80,17 @@ export default Vue.extend({
   },
   methods: {
     home(): void {
-      this.$router.replace({ name: RouteName.Clients });
+      this.$router
+        .replace({
+          name: this.$store.getters.firstAllowedTab.to.name,
+        })
+        .catch((err) => {
+          // Ignore the error regarding navigating to the same path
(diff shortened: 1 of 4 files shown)


src/proxy-ui-api/frontend/src/components/layout/AppToolbar.vue
```diff
@@ -80,7 +80,17 @@ export default Vue.extend({
   },
   methods: {
     home(): void {
-      this.$router.replace({ name: RouteName.Clients });
+      this.$router
+        .replace({
+          name: this.$store.getters.firstAllowedTab.to.name,
+        })
+        .catch((err) => {
```
TJaakkola, line 87: The fix in itself is ok. I was thinking whether we should not completely swallow the expection but instead add somekind of console.log notice etc. I wonder if there could now be an endless routing loop and we wouldn't know why the app would fail?

TJaakkola (changes requested):

src/proxy-ui-api/frontend/src/components/layout/AppToolbar.vue
Riippi: I added a console.log there. Now there is some clue if something weird starts happening.

TJaakkola (approved):

PR #690, opened 2020-09-07
XEDDEV-1303 Fix router permissions

Riippi:
https://jira.niis.org/browse/XRDDEV-1303
https://jira.niis.org/browse/XRDDEV-1304
https://jira.niis.org/browse/XRDDEV-1308


TJaakkola (changes requested):

TJaakkola:
In addition to code comments. I tested this behaviour locally and I could access the Security server TLS certificate -page in contrary to how it is described to work in the old ui.

TJaakkola (approved):

PR #693, opened 2020-09-08
XRDDEV-1321

TJaakkola:
Populate read-only field in key details with token data.

https://jira.niis.org/browse/XRDDEV-1321


Riippi (approved):
LGTM

PR #694, opened 2020-09-09
XRDDEV-1091 CS installation manual update

iluwatar:


jhyoty (commented):
- Images could be more compact. Now they take a lot of space but provide relatively little information.
- from the CS point of view "remote database" and "remote cloud database" deployment options are more or less the same, could be combined with the mention that also managed cloud databases can be used if CS is deployed in a cloud.
- Clustered CS with a remote database (clustered or not) is also an option
- The database setup does not provide instructions how to set up the system without the database super-user credentials stored on the CS host. However, that is possible if the database (and users) are created prior to the installation.

petkivim:
- In the images, please use the Central Server logo instead of X-Road logo inside the CS instances.
  - The CS logo is available in the diagram toolkit.

iluwatar:
Addressed the review comments and left a comment to the Jira ticket.

petkivim:
* In my opinion it would be good to change the order of chapters 2.6 and 2.7 so that chapters related to database configuration follow each other and package repository  / installation chapters follow each other. In that way the order of the chapters is more logical. Now the installation guide is jumping between the topics.
  * 2.7  -> 2.5.2
  * 2.8  ->  2.7
*  The diagrams should have a higher resolution. Their current size in the document is OK, but it would be good to be able see a bigger version when opening the image file directly.
    * I usually use the advanced export feature and set DPI to `400dpi`.
* It might be good to mention in section 2.5.1 that it's not required to define all the parameter values in `/etc/xroad/db.properties`. It's enough to define the values that are customized, and the parameters that are not defined in the file use the default values instead.

jhyoty:
About the images: Maybe the scalable vector graphics (SVG) format could be used instead (well supported by browsers: https://www.caniuse.com/svg-img).

iluwatar:
Pushed changes based on review comments. The images are still in png-format, but let me know if they need to be converted.

petkivim:
Thanks, looks good to me now. For me it's OK to use `png` images. We should probably define the image format to be used in the documentation so that it would be consistent between different documents.

jhyoty:
Converted images to SVG format to improve readability. Especially the text in the small png images was quite hard to read. 
I'd suggest that we use SVG in diagrams (it is even possible to embed the drawio source into the svg file so that one can directly edit that).

jhyoty (approved):

Issue #691, opened 2020-09-08
Jetty download fails when building packages


mbert:
In the script `./src/packages/download_jetty.sh` the `wget` command is used to download the jetty distribution. This no longer works, because the build docker image from `./src/packages/docker-compile/Dockerfile` which is based on `ubuntu:18.04` does not contain `wget` anymore. 

Hence a fix is needed: either install `wget` in the build image build or use `curl` instead in the jetty download script.

I guess that this problem has remained unnoticed because the build images have probably been built a long time ago, so that the package list change in `ubuntu:18.04` had no effect for your builds. 

Could you please check?

I was trying to build from tag `6.24.0`.

petkivim:
How did you try to build the installation packages? You should use the `./src/build_packages.sh` script which takes care of compiling the code and building the packages. The build instructions are documented [here](https://github.com/nordic-institute/X-Road/blob/develop/src/BUILD.md).

The Docker file that you're referring to (`./src/packages/docker-compile/Dockerfile`) is meant for our CI tools only, and it's not meant to be used manually by developers.

mbert:
I am using a shell script mimicking the build instructions from the Jenkinsfile. This seems to make sense, because we need to produce packages for all the various distributions. 

But if I'm right, you should get the same problem as I did once you simply delete the build docker image in your CI environment so that it needs to be rebuilt.

petkivim:
We tested the scripts in a clean environment and unfortunately, we were not able to reproduce the issue. The Docker image that you're referring to is for compiling the code, and the script `./src/packages/download_jetty.sh` is not needed until packaging. And the image that is used for packaging (`./src/packages/docker/deb-bionic/Dockerfile`) already [installs](src/packages/docker/deb-bionic/Dockerfile) `wget`. Therefore, the current images should be OK.

In case you try to run `build_packages.sh` inside a container created from `./src/packages/docker-compile/Dockerfile `, the process probably fails because `wget` is missing. However, that's not the way how the image is meant to be used.

mbert:
OK, thank you, I will need to go through the process again and compare the Jenkinsfile and my script step by step. It used to work with previous X-Road releases, now it no longer does. So something must have changed.

petkivim:
You're welcome! The most simple solution is to install Docker on your Jenkins and run the `build_packages.sh` script. Since Docker is installed, Jenkins will build packages for all the supported distributions.

mbert:
[...]

PR #695, opened 2020-09-09
Issue #691 - improvement to build script: if docker is available, the…

mbert:
Changes to the build script and the build instructions:

- if docker is available, the whole process is performed in docker containers, no need to install further build dependencies on the build host
- if docker is not available, perform some minimal checking on the build dependencies on the host
- implement early fail error handling
- changed the build instructions accordingly

This PR should not make any changes to existing build environments necessary.

### DIFF
--- a/src/BUILD.md
+++ b/src/BUILD.md
@@ -4,7 +4,10 @@ Running the X-Road software requires Linux (Ubuntu or RHEL). As a development en
 
 **Tools**
 
-*Required for building*
+*Required for building in Docker (recommended)*
+* Docker (for deb/rpm packaging)
+
+*Required for building natively (without Docker)*
 * OpenJDK / JDK version 8
 * Gradle
 * JRuby and rvm (ruby version manager)
@@ -19,7 +22,15 @@ Running the X-Road software requires Linux (Ubuntu or RHEL). As a development en
 
 The development environment should have at least 8GB of memory and 20GB of free disk space (applies to a virtual machine as well), especially if you set up a local X-Road instance.
 
-## Dependency installation and building instructions
+## Dependency installation and instructions for building in Docker
+
+* Install Docker
+
+* Build the software and installation packages:
+
+    `./build_packages.sh`
+
+## Dependency installation and instructions for building natively (without Docker)
 
 * Requires Ubuntu >=16.04, 18.04 is recommended.
 
--- a/src/build_packages.sh
+++ b/src/build_packages.sh
@@ -2,18 +2,27 @@
 set -e
 export XROAD=$(cd "$(dirname "$0")"; pwd)
 
-./compile_code.sh "$@"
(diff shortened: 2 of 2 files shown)


petkivim:
Hi @mbert 

Thank you for your contribution! Before approving it we would need to have the NIIS Contributor License Agreement (CLA) document signed by you. More information about the NIIS CLA and instructions for signing and delivering it can be found at:

https://github.com/nordic-institute/X-Road-development/#contributor-license-agreement

Best regards,
Petteri

mbert:
Can. Unfortunately I am under immense time pressure at the moment (I just squeezed this PR in during my breakfast), and I won't be able to read the document on e-signatures. Is there a quick and simple way? Maybe print out the agreement, sign it, scan it in and send it to some email address?

Sorry to bother you with this, but I really need to move on now.

mbert:
Otherwise I'm totally fine if you take my change, commit it under your own name (may mention me in the commit message) and submit the change, if that's easier for you. I am really not that important :)

petkivim:
The easiest way to proceed with the CLA is to print it out, signg manually and sent to NIIS by post:

```
MTÜ Nordic Institute for Interoperability Solutions
Hobujaama 4
10151 Tallinn
Estonia
```

The evaluation of the pull request will be added to our backlog, and implemented during the upcoming sprints. In other words, the process of reviewing the pull request will take some time in any case, and the CLA probably arrives before it.

mbert:
To be honest this procedure isn't exactly going to encourage a vast community to contribute ;)
You might want to take a look at the way the CNCF are doing this.

Nevertheless, printed, signed, will send this out the coming days.

petkivim:
Thanks! Yes, I know that the process could be more straight forward. Since NIIS is an Estonian organisation, we must have the CLA and its signature in the format that's required by Estonian legislation or otherwise it has no legal value. Fortunately, you only need to submit the CLA once and it covers all you future contributions too.

petkivim:
Hi @mbert

We have reviewed your pull request. Unfortunately, we have decided to reject it, because the current implementation does not serve different needs that X-Road developers may have. For example, the containers that are used to compile the code and build the installation packages are removed after every run. It means that all the dependencies must be downloaded every time when the code is compiled which increases the total build time and adds overhead. Also, developers may prefer to do the build locally on the host even if Docker is installed. 

We agree that doing the whole build process using Docker is something that should be supported. However, the support should be implemented in a more flexible and configurable manner that enables developers to choose between available alternatives. Therefore, we will create a new backlog item about implementing Docker support for the build process and implement it by ourself. 

Thank you for the CLA. We have received it and it's valid for any future contributions.

Best regards,
Petteri

mbert:
I understand where you're coming from. I have amended my commit, and I think it can now cater for what you need. 
I'll open a new PR for it.

PR #721, opened 2020-09-21
XRDDEV-1266 Error snackbar update

Riippi:
https://jira.niis.org/browse/XRDDEV-1266
https://jira.niis.org/browse/XRDDEV-1236

Updates for snackbars.
Updated version of vuetify.


src/proxy-ui-api/frontend/src/store/modules/notifications.ts
```diff
@@ -28,41 +28,104 @@ import { RootState } from '../types';
...
+    successNotifications: [],
   };
 };
 
+// Finds if an array of notifications contains a similar notification.
+function containsNotification(
+  errorNotifications: Notification[],
+  notification: Notification,
+): number {
+  if (!notification || !errorNotifications || errorNotifications.length === 0) {
```
TJaakkola, line 47: I think you could use the custom isEmpty -utility method from util/helpers.ts here.

So the clause could be written like this (double check!):

if (isEmpty(notification) || isEmpty(errorNotifications)) {

TJaakkola (changes requested):

src/proxy-ui-api/frontend/src/store/modules/notifications.ts
Riippi: isEmpty accepts strings or arrays. And seems that it's currently used only for strings. I'm not sure if it would be good to expand it for Notifications or some more general things.

TJaakkola: Ah yes, I've that notification is an object. But I guess it could be used for errorNotifications

Riippi: I suppose the line would look like this  `if (!notification || isEmpty(errorNotifications as [])) {`  
and import of the isEmpty would add one line. So it seems a bit boilerplated and I would rather keep it like it is now.

Related to the isEmpty method, I think there is a case to make it to accept string only, or somehow define the array type smarter.

TJaakkola (approved):

PR #723, opened 2020-09-22
XRDDEV-1244: Add fixes to ui breaking when there's long identifiers

TJaakkola:
https://jira.niis.org/browse/XRDDEV-1244


Riippi (changes requested):
Run the linter. Looks good otherwise.

Riippi (approved):
LGTM

PR #725, opened 2020-09-23
XRDDEV-1213 Remove duplicate code

Riippi:
https://jira.niis.org/browse/XRDDEV-1213


carohauta (commented):
Also [AddClient.vue](https://github.com/nordic-institute/X-Road/blob/XRDDEV-1213-remove-duplicate-code/src/proxy-ui-api/frontend/src/views/AddClient/AddClient.vue) and [AddMember.vue](https://github.com/nordic-institute/X-Road/blob/XRDDEV-1213-remove-duplicate-code/src/proxy-ui-api/frontend/src/views/AddMember/AddMember.vue) look very similar. Did you look if there aren't any reasons to merge those into one component too?

Riippi:
Yes I did. Even thought they are quite similar a shared component would have needed some if statements or splitting into multiple pieces. I felt it would not be worth of the effort.

carohauta (approved):
🦄

PR #726, opened 2020-09-23
Issue #691 - improvement to build script: if docker is available, the…

mbert:
Changes to the build script and the build instructions:

- if docker is available, the whole process can be performed in docker containers if passing the `-d` or `--docker-build` as a command line option, so that there is no need to install further build dependencies on the build host
- if docker is not available, perform some minimal checking on the build dependencies on the host
- implement early fail error handling
- changed the build instructions accordingly 

This PR should not make any changes to existing build environments necessary.

### DIFF
--- a/src/BUILD.md
+++ b/src/BUILD.md
@@ -1,10 +1,13 @@
 # Building X-Road
 
-Running the X-Road software requires Linux (Ubuntu or RHEL). As a development environment,  only Ubuntu (>=16.04, 20.04 recommended) is currently supported. It should be possible to use some other Linux distribution for development, but the instructions and helper scripts assume Ubuntu. If you are using some other operating system (e.g. Windows or macOS), the easiest option is to first install Ubuntu into a virtual machine.
+Running the X-Road software requires Linux (Ubuntu or RHEL). As a development environment, only Ubuntu (>=16.04, 20.04 recommended) is currently supported. It should be possible to use some other Linux distribution for development, but the instructions and helper scripts assume Ubuntu. Alternatively the software can be built entirely inside docker containers (see below) making the build host distribution agnostic but also a bit slower. If you are using some other operating system (e.g. Windows or macOS), the easiest option is to first install Ubuntu into a virtual machine.
 
 **Tools**
 
-*Required for building*
+*Required for deb/rpm packaging and/or building in Docker*
+* Docker
+
+*Required for building natively (without Docker)*
 * OpenJDK / JDK version 8
 * Gradle
 * JRuby and rvm (ruby version manager)
@@ -19,7 +22,18 @@ Running the X-Road software requires Linux (Ubuntu or RHEL). As a development en
 
 The development environment should have at least 8GB of memory and 20GB of free disk space (applies to a virtual machine as well), especially if you set up a local X-Road instance.
 
-## Dependency installation and building instructions
+## Dependency installation and instructions for building in Docker
+
+* Install Docker
+
+* Build the software and installation packages:
+
+    `./build_packages.sh -d`
+
+    Note that the `build_package.sh` script's behaviour is controlled by its first command line argument.
+    All subsequent arguments will simply be passed to whatever build script it calls to perform the actual build.
+
+## Dependency installation and instructions for building natively (without Docker)
 
 * Requires Ubuntu >=16.04, 20.04 is recommended.
 
--- a/src/build_packages.sh
+++ b/src/build_packages.sh
@@ -2,27 +2,69 @@
 set -e
 export XROAD=$(cd "$(dirname "$0")"; pwd)
 
-if [[ ! " $* " =~ " --packageonly " ]]; then
-  ./compile_code.sh "$@"
-fi
+HAS_DOCKER=""
+
+errorExit() {
+    echo "*** $*" 1>&2
+    exit 1
+}
+
+usage () {
+    echo "Usage: $0 [option for $0...] [other options]"
+    echo "Options for $0:"
+    echo " -p, --package-only    Skip compilation, just build packages'"
+    echo " -d, --docker-compile  Compile in docker container instead of native gradle build"
+    echo " -h, --help            This help text."
+    echo "The option for $0, if present, must come fist, before other options."
+    echo "Other options are passed on to compile_code.sh"
+    test -z "$1" || exit "$1"
+}
+
+buildInDocker() {
+    test -n "$HAS_DOCKER" || errorExit "Error, docker is not installed/running."
+    echo "Building in docker..."
+    # check if running attached to terminal
+    # makes it possible to stop build with Ctrl+C
+    if [ -t 1 ]; then OPT="-it"; fi
+
+    docker build -q -t xroad-build --build-arg uid=$(id -u) --build-arg gid=$(id -g) $XROAD/packages/docker-compile || errorExit "Error building build image."
+    docker run --rm -v $XROAD/..:/workspace -w /workspace/src -u builder ${OPT} xroad-build  bash -c "./update_ruby_dependencies.sh && ./compile_code.sh -nodaemon" || errorExit "Error running build of binaries."
+}
+
+buildLocally() {
+    echo "Building locally..."
+    cd $XROAD || errorExit "Error 'cd $XROAD'."
+    ./compile_code.sh "$@" || errorExit "Error running build of binaries."
+}
 
 if command -v docker &>/dev/null; then
-    docker build -q -t xroad-deb-bionic "$XROAD/packages/docker/deb-bionic"
-    docker build -q -t xroad-deb-focal "$XROAD/packages/docker/deb-focal"
-    docker build -q -t xroad-rpm "$XROAD/packages/docker/rpm"
-    docker build -q -t xroad-rpm-el8 "$XROAD/packages/docker/rpm-el8"
+    HAS_DOCKER=true
+fi
+
+case "$1" in
+    --package-only|-p) shift;;
+    --docker-compile|-d) shift; buildInDocker "$@";;
+    --help|-h) usage 0;;
+    *) buildLocally "$@";;
+esac
+
+if [ -n "$HAS_DOCKER" ]; then
+    docker build -q -t xroad-deb-bionic "$XROAD/packages/docker/deb-bionic" || errorExit "Error building deb-bionic image."
+    docker build -q -t xroad-deb-focal "$XROAD/packages/docker/deb-focal" || errorExit "Error building deb-focal image."
+    docker build -q -t xroad-rpm "$XROAD/packages/docker/rpm" || errorExit "Error building rpm image."
+    docker build -q -t xroad-rpm-el8 "$XROAD/packages/docker/rpm-el8" || errorExit "Error building rpm-el8 image."
 
     OPTS=("--rm" "-v" "$XROAD/..:/workspace" "-u" "$(id -u):$(id -g)" "-e" "HOME=/workspace/src/packages")
     # check if running attached to terminal
     # makes it possible to stop build with Ctrl+C
     if [[ -t 1 ]]; then OPTS+=("-it"); fi
 
-    docker run "${OPTS[@]}" xroad-deb-bionic /workspace/src/packages/build-deb.sh bionic
-    docker run "${OPTS[@]}" xroad-deb-focal /workspace/src/packages/build-deb.sh focal
-    docker run "${OPTS[@]}" xroad-rpm /workspace/src/packages/build-rpm.sh
-    docker run "${OPTS[@]}" xroad-rpm-el8 /workspace/src/packages/build-rpm.sh
+    docker run "${OPTS[@]}" xroad-deb-bionic /workspace/src/packages/build-deb.sh bionic || errorExit "Error building deb-bionic packages."
+    docker run "${OPTS[@]}" xroad-deb-focal /workspace/src/packages/build-deb.sh focal || errorExit "Error building deb-focal packages."
+    docker run "${OPTS[@]}" xroad-rpm /workspace/src/packages/build-rpm.sh || errorExit "Error building rpm packages."
+    docker run "${OPTS[@]}" xroad-rpm-el8 /workspace/src/packages/build-rpm.sh || errorExit "Error building rpm-el8 packages."
 else
     echo "Docker not installed, building only .deb packages for this distribution"
     cd "$XROAD/packages"
-    ./build-deb.sh "$(lsb_release -sc)"
+    ./build-deb.sh "$(lsb_release -sc)" || errorExit "Error building deb packages."
 fi


mbert:
@petkivim here's the new PR for #691 (replacing PR #695).

petkivim:
@mbert Thank you for the updated contribution! I have created a backlog item about reviewing it:

https://jira.niis.org/browse/XRDDEV-1367

src/build_packages.sh
```diff
@@ -2,27 +2,58 @@
...
+}
+
+buildInDocker() {
+    test -n "$HAS_DOCKER" || errorExit "Error, docker is not installed/running."
+    echo "Building in docker..."
+    BUILD_IMAGE="`docker images | grep xroad-build | awk '{ print $1 }'`"
+    if [ "$BUILD_IMAGE" != "xroad-build" ]; then
+        echo "Build image 'xroad-build' not found, building it."
+        docker build -q -t xroad-build --build-arg uid=$(id -u) --build-arg gid=$(id -g) $XROAD/packages/docker-compile || errorExit "Error building build image."
+    fi
```
jansu76, line 19: Is this "check if image is already built" logic actually needed?

The regular docker image builds (such as ones used for packaging, https://github.com/nordic-institute/X-Road/blob/develop/src/build_packages.sh#L10-L13) already skip building if there are no changes and use build cache https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#leverage-build-cache instead.

Using the regular build also has the advantage that it notices if image **does** need to be rebuild (e.g. changes to Dockerfile), which this does not.

src/build_packages.sh
```diff
@@ -2,27 +2,58 @@
...
-    docker build -q -t xroad-rpm "$XROAD/packages/docker/rpm"
-    docker build -q -t xroad-rpm-el8 "$XROAD/packages/docker/rpm-el8"
+    HAS_DOCKER=true
+fi
+
+case "$1" in
+    --packageonly|-p) shift;;
+    --docker-build|-d) shift; buildInDocker "$@";;
+    *) shift; buildLocally "$@";;
+esac
```
jansu76, line 37: Maybe I would not use shift to remove the parameter. The scripts `build_packages.sh` and `compile_code.sh` seem to support parameter carrying so that you can e.g. call `build_packages.sh param1 param2` where param1 param2 are passed on as `compile_code.sh` parameters.

shift assumes that build-mode parameter (packageonly, docker-build, p, d, and any other parameter which means "build locally") is the first one. If order is different, a wrong parameter can get removed. Old script also meant calling `build_packages.sh param1` resulted in execution if `compile_code.sh param1`, but now it executes `compile_code.sh` which could be a breaking change for some. And also code compile_code.sh does not care about parameter order, maybe this script could use consistent approach.

I would probably make build-style parameter not required to be the first one, and just pass all parameters to compile_code without shifting. Either that, or document the behavior.

jansu76 (changes requested):
Thanks for the PR @mbert ! And sorry for the long delay. Ticket for reviewing this took its time travelling from our backlog into a development sprint.

We think this fully dockerized build is a useful tool for enabling builds easily without disturbing your host computer. It can be valuable in setting up CI pipelines, verifying the build against a known good configuration, etc. 

It does not work well with continuous iterative development, which is the kind of thing we do mostly. For that it is too slow, as it starts up with installing Ruby gems, downloading gradle and required dependencies, etc. We could optimize this, but that would add some complexity, and after some discussion we decided that is not worth pursuing.

Current plan is to use the two modes for what they work best for:

1. native gradle build for faster iterative development
2. fully dockerized build for things like CI or just running the build once, maybe doing a sanity check against a known good configuration

I have added some review comments on the PR. We already do have a ticket about, more or less, fixing those and integrating this feature to develop: https://jira.niis.org/browse/XRDDEV-1363. If you want you can work on those comments yourself, or if you don't they will be fixed by ourselves after the ticket gets in a development sprint.

src/BUILD.md
```diff
@@ -4,7 +4,10 @@ Running the X-Road software requires Linux (Ubuntu or RHEL). As a development en
 
 **Tools**
 
-*Required for building*
+*Required for building in Docker (recommended)*
```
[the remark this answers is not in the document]
mbert: I have weaved in some text into the introduction. Actually building in docker is the only option for me because I cannot afford "taining" my machine with stuff not related to my work. But we know from experience that longer a README gets the less people will read it :)

src/BUILD.md
```diff
@@ -4,7 +4,10 @@ Running the X-Road software requires Linux (Ubuntu or RHEL). As a development en
...
+* Docker (for deb/rpm packaging)
```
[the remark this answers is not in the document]
mbert: I've made it "Required for deb/rpm packaging and/or building in Docker".

src/build_packages.sh
mbert: That's fine for me.

src/build_packages.sh
mbert: You're right, the last shift should not be there. I have removed it. The first two are necessary, because the `--packageonly` or `--docker-build` arguments are solely for `build_packages.sh` and should not be passed to any other scripts.

mbert:
@jansu76 I have rebased and uploaded another revision to this PR. Can you please review again?

jansu76:
> @jansu76 I have rebased and uploaded another revision to this PR. Can you please review again?

Sure, thanks for the updates! I have kind of a challenging schedule for Thursday and Friday so it might take a few days.

mbert:
> > @jansu76 I have rebased and uploaded another revision to this PR. Can you please review again?
> 
> Sure, thanks for the updates! I have kind of a challenging schedule for Thursday and Friday so it might take a few days.

Same here, so no problem. But this one should not take long, the changes are homeopathic ;)

src/build_packages.sh
```diff
@@ -2,27 +2,54 @@
...
-    docker build -q -t xroad-rpm "$XROAD/packages/docker/rpm"
-    docker build -q -t xroad-rpm-el8 "$XROAD/packages/docker/rpm-el8"
+    HAS_DOCKER=true
+fi
+
+case "$1" in
+    --packageonly|-p) shift;;
+    --docker-build|-d) shift; buildInDocker "$@";;
+    *) buildLocally "$@";;
+esac
```
jansu76, line 33:
```suggestion
for i in "$@"; do
shift
case "$i" in
    "--package-only"|"-p")
        PACKAGE_ONLY=1
        continue
        ;;
    "--docker-compile"|"-d")
        DOCKER_COMPILE=1
        continue
        ;;
    "--help"|"-h")
         usage
        ;;
    *) set -- "$@" "$i";;
esac
done

if [[ -n "$PACKAGE_ONLY" && -n "$DOCKER_COMPILE" ]]; then
    echo "Can't use both package-only and docker-compile options at the same time"
    usage
fi

if [[ -n "$PACKAGE_ONLY" ]]; then
    echo "Skipping compilation..."
elif [[ -n "$DOCKER_COMPILE" ]]; then
    buildInDocker "$@"
else
    buildLocally "$@"
fi
```

Suggestion: parameter handling in the same fashion as `compile_code.sh`: https://github.com/nordic-institute/X-Road/blob/develop/src/compile_code.sh#L6-L18

- docker-compile, --package-only, -d, -p can appear in any parameter positions, do not have to be the first ones
- those parameters are removed, and retained parameters are carried over to e.g. compile_code.sh
- also minor rewording of parameter names and added -h, --help parameter and usage display

src/build_packages.sh
```diff
@@ -2,27 +2,54 @@
...
+    echo "Building in docker..."
+    docker build -q -t xroad-build --build-arg uid=$(id -u) --build-arg gid=$(id -g) $XROAD/packages/docker-compile || errorExit "Error building build image."
+    docker run --rm -v $XROAD/..:/workspace -w /workspace/src -u builder xroad-build  bash -c "./update_ruby_dependencies.sh && ./compile_code.sh -nodaemon" || errorExit "Error running build of binaries."
+}
+
+buildLocally() {
+    echo "Building locally..."
+    cd $XROAD || errorExit "Error 'cd $XROAD'."
+    ./compile_code.sh "$@" || errorExit "Error running build of binaries."
+}
```
jansu76, line 24:
```suggestion
usage () {
  echo "Usage: $0 [options for $0...] [other options]"
  echo "Options for $0:"
  echo " -p, --package-only    Skip compilation, just build packages'"
  echo " -d, --docker-compile  Compile in docker container instead of native gradle build"
  echo " -h, --help            This help text."
  echo "Other options are passed on to compile_code.sh"

  exit 2
}
```

jansu76 (changes requested):

src/build_packages.sh
jansu76: I made some suggestions that would make build-mode parameter order more flexible and make the handling implementation more consistent with how `compile_code.sh` does it:

- https://github.com/nordic-institute/X-Road/pull/726#pullrequestreview-555514097
- https://github.com/nordic-institute/X-Road/pull/726#discussion_r545856375

src/build_packages.sh
mbert: I find the construct with `set -- "$@" "$i"` rather confusing. The `continue` statements are unnecessary. 
I really prefer to keep things simple (that's why I personally would be fine with allowing those parameters only in the first position). 

The example from `compile_code.sh` looks better to me because it does not mix a `for` loop with `shift` statements (this should really be avoided). 

If you think a more complete parameter handling is necessary I can dive a bit deeper into it and make a proposal.

src/build_packages.sh
mbert: Can do.

src/build_packages.sh
jansu76: That set & shift approach is inspired by https://unix.stackexchange.com/a/258514. Not sure why for loop with shift statements should be avoided in general? 

Those `continue`s look unnecessary, I think you are right. 

I did test the proposed solution with different parameter combinations, and all parameters were evaluated if I am not mistaken.

mbert:
@jansu76  OK, here's another go. I avoid rebuilding `$@` but rather enforce the script-specific options to come first. That keeps the code simple and comprehensible. 

Apart from that it should be pretty close to what you proposed.

src/build_packages.sh
mbert: I avoid using `shift` in `for` loops altogether because it is very confusing. Shell code needs to be treated with care anyway because scripts are not parsed but rather simply executed line by line (and hence tend to abort in the middle of a file system operation leaving a complete mess). 

That's why I usually recommend to use patterns as clear and simple as possible. When parsing command line args, it's usually either `while` + `shift` or it's `for`. Only in this case it's getting a bit inconvenient because we want to pass on command line args preserving white spaces. Rather than trying to build a new argument list I rather try to avoid having to do so in the first place :)

jansu76: Thanks for the updates!

Looks like this still expects build-mode parameter to be in first position. If we were to do so, there is no reason for the for-loop over all parameters. The earlier syntax of

```
case "$1" in
    --packageonly|-p) shift;;
    --docker-build|-d) shift; buildInDocker "$@";;
    *) shift; buildLocally "$@";;
esac
```

would be simpler and better.

**I will try to argue for the freely order build-mode (d/p) parameters**. The build modes operate a bit weirdly now, as there are two modes that trigger a different-from-default build path, and absence of those triggers the default build path. Everything might a bit more straightforward if there was symmetry and buildLocally was also a mandatory parameter (but I am not very convinced that it would be optimal either). Now, putting the build-mode parameter into wrong index by mistake triggers wrong build path which may be an error that is possibly not immediately easy to detect. Especially when combined with how the sister-script `compile_code.sh` behaves (parameters do have free order) this is a recipe for confusion IMO. 

I know the free-order parameters are not perfect either (potential for masking sharing parameters - but shared parameters should be avoided in any case IMO), but this is still the way I would prefer.

As I put in my suggestion in https://github.com/nordic-institute/X-Road/pull/726#discussion_r543144343, requiring build-mode parameters to be the first ones is also an acceptable solution IMO. I would prefer the free-order solution (especially after I've already spent the time to write and test it), but fixed order with improved documentation would be merged as well.

I agree 100% with the point about shell code needing to be treated with care, and to be as clear and simple as possible, I don't think we have a disagreement there. However things like `shift` in `for` loops _is confusing_ or _is not confusing_ is more subjective, and also down to what solutions each developer is used to. IMO using shift in the for-loop is fine here. Maybe some additional comments would make it more clear:

```
# remove all build-mode params (--package-only,-p,--docker-compile,-d), maintain others
for i in "$@"; do
    # remove current parameter i from all params
    shift
    case "$i" in
        --package-only|-p)
            PACKAGE_ONLY=1
            ;;
        --docker-compile|-d)
            DOCKER_COMPILE=1
            ;;
        --help|-h)
            usage 0
            ;;
        *)
            # bring back removed parameter i
            set -- "$@" "$i"
            ;;
    esac
done
```

mbert: > I agree 100% with the point about shell code needing to be treated with care, and to be as clear and simple as possible, I don't think we have a disagreement there. However things like `shift` in `for` loops _is confusing_ or _is not confusing_ is more subjective, and also down to what solutions each developer is used to.

We can of course happily agree to disagree on that one :)
Just one last remark, why I don't really favour the construction with the `for` loop: Usually `shift` is used when you always look at `$1`, and that's what you'd do in a `while` loop:

```
while [ $# -gt 0 ]; do
    case "$1" in
        *) ...;; 
    esac
    shift
done
```
In that case I consume the very parameter you're looking at. When using a `for` loop, `shift` does not work the same way, because it's always consuming the first parameter regardless which one is in `$i`. These are code blocks that people writing a lot of shell script code usually have in their heads and just put them in when necessary. Of course, in your case you're always consuming before looking, so the behaviour is consistent and defined, but still it takes some time to grasp. 

> Looks like this still expects build-mode parameter to be in first position. If we were to do so, there is no reason for the for-loop over all parameters. The earlier syntax of
> 
> ```
> case "$1" in
>     --packageonly|-p) shift;;
>     --docker-build|-d) shift; buildInDocker "$@";;
>     *) shift; buildLocally "$@";;
> esac
> ```
> 
> would be simpler and better (copy-pasted from old, uses parameter names no longer in use).
> 
> **I will try to argue for the freely ordered build-mode (d/p) parameters**. The build modes operate a bit weirdly now, as there are two modes that trigger a different-from-default build path, and absence of those triggers the default build path. Everything might a bit more straightforward if there was symmetry and buildLocally was also a mandatory parameter (but I am not very convinced that it would be optimal either). Now, putting the build-mode parameter into wrong index by mistake triggers wrong build path which may be an error that is possibly not immediately easy to detect. Especially when combined with how the sister-script `compile_code.sh` behaves (parameters do have free order) this is a recipe for confusion IMO.

Actually I do not think that the parameters to `build_package.sh` should be freely placable for two reasons:
1. The same parameters (e.g. `-p` or `-d`) could be legal for the script called afterwards. They'd get filtered out here which could cause problems.
2. For me as a user a clear distinction between parameters to the outer and parameters to the inner script makes sense. The simple method is just enforce parameters to the outer script at position 1, a more sophisticated would be to introduce a parameter `--` after which everything will be ignored by the outer script.

For a build script as simple as this one I would prefer not to use the `--` variant, because it makes the script longer than necessary and/or forces me to do "dirty" things like rebuilding `$@`. My original version's charm was its simplicity, but I accept that error handling (`-p` and `-d`) was missing. 

In general I try to keep shell scripts short. My personal convention is to scrutinise everything significantly longer than an editor page and get rid of unnecessary stuff if possible (more code - more bugs, fatal in a programming language without a parser). 

To be honest, I still like my first version best, but the second would still be OK. I don't like the `set -- "$@" "$i"` construct, because basically everybody needs to look up what it does (that's by the way also my standard criticism on `bash`isms), and few people will be able to understand _why_ it works. 

But long story short (I do enjoy this kind of debates): if you really want the variant you proposed I'd be - reluctantly - willing to go along. I'd still prefer the simple one :)

jansu76: Heh, looks like we have created quite a long debate about this detail :) Thanks for explaining your thinking, appreciated. I guess we understand the related factors mostly the same way, but maybe put different weights to different factors when summing up the overall preferred approach.

To recap, I think both options are fine for this PR

1. build-mode params (--package-only,-p,--docker-compile,-d) in fixed first position - close to your 1st version
2. build-mode params in any positions, will get removed for inner script - close to what I proposed

For (1) I would like to see a bit of added documentation, making this as clear as possible for the user, so that she does not `build_packages.sh param1 --docker-build param2` by accident. Usage() at least could use couple of additional words about it, maybe also BUILD.md?

I still prefer (2) but it is not a must.

mbert:
I have uploaded another revision. Hope we're getting there now :)

src/center-service/Gemfile.lock
```diff
@@ -99,7 +99,7 @@ GEM
       rake (>= 0.8.7)
       rdoc (~> 3.4)
       thor (>= 0.14.6, < 2.0)
-    rake (13.0.1)
+    rake (13.0.3)
```
jansu76, line 102: Seems unrelated to the PR?

src/center-ui/Gemfile.lock
```diff
@@ -99,7 +99,7 @@ GEM
       rake (>= 0.8.7)
       rdoc (~> 3.4)
       thor (>= 0.14.6, < 2.0)
-    rake (13.0.1)
+    rake (13.0.3)
```
jansu76, line 102: Seems unrelated to the PR?

jansu76 (changes requested):

jansu76:
> I have uploaded another revision. Hope we're getting there now :)

Thanks, the documentation and usage seem spot on.

What happened with the Gemfile.lock updates, they seem unrelated to this PR, are they included by mistake? Could you revert those please, if they are not related?

mbert:
Yes, give me a sec, I'll quickly fix this.

mbert:
One of the reasons why I really dislike the gradle / ruby combination :)
New patch uploaded, hope it's OK now.

src/build_packages.sh
```diff
@@ -2,27 +2,65 @@
...
+    echo "The option for $0, if present, must come fist, before other options."
+    echo "Other options are passed on to compile_code.sh"
+    test -z "$1" || exit "$1"
+}
+
+buildInDocker() {
+    test -n "$HAS_DOCKER" || errorExit "Error, docker is not installed/running."
+    echo "Building in docker..."
+    docker build -q -t xroad-build --build-arg uid=$(id -u) --build-arg gid=$(id -g) $XROAD/packages/docker-compile || errorExit "Error building build image."
+    docker run --rm -v $XROAD/..:/workspace -w /workspace/src -u builder xroad-build  bash -c "./update_ruby_dependencies.sh && ./compile_code.sh -nodaemon" || errorExit "Error running build of binaries."
```
jansu76, line 27: Should include a conditional `-it` parameter to make it possible to interrupt build with Ctrl+C, as in packaging https://github.com/nordic-institute/X-Road/blob/develop/src/build_packages.sh#L16-L18

jansu76 (changes requested):

jansu76:
> One of the reasons why I really dislike the gradle / ruby combination :)
> New patch uploaded, hope it's OK now.

Noticed that docker compile does not make it possible to interrupt build, this can be a bit of a pain. Docker packaging has this option, added a comment https://github.com/nordic-institute/X-Road/pull/726#discussion_r547346605

I have a refactored build_packages.sh with this feature in the works, with a `COMMON_DOCKER_RUN_OPTS` for both build and packaging. I could take this PR and finalize that change, to wrap things up?

A heads up, I'm just off to christmas vacation and this will be on hold for about a week. Happy holidays!

mbert:
OK, repushed again, Ctrl-C now works. 

Have a nice christmas and happy new year!

src/build_packages.sh
```diff
@@ -2,27 +2,69 @@
...
 if command -v docker &>/dev/null; then
-    docker build -q -t xroad-deb-bionic "$XROAD/packages/docker/deb-bionic"
-    docker build -q -t xroad-deb-focal "$XROAD/packages/docker/deb-focal"
-    docker build -q -t xroad-rpm "$XROAD/packages/docker/rpm"
-    docker build -q -t xroad-rpm-el8 "$XROAD/packages/docker/rpm-el8"
+    HAS_DOCKER=true
+fi
+
+case "$1" in
+    --packageonly|-p) shift;;
```
jansu76, line 45: Actual expected parameters `packageonly` and `docker-build` do not match now what usage says (`package-only` and `docker-compile`).

I would use the names from usage. Change from packageonly to package-only is in principle a breaking change, but has been an undocumented feature and by a quick check should not cause trouble for our CI jobs, at least.

jansu76 (changes requested):

jansu76:
> OK, repushed again, Ctrl-C now works.
> 
> Have a nice christmas and happy new year!

Thank you! I noticed a small issue with parameter names: https://github.com/nordic-institute/X-Road/pull/726#discussion_r550066260

I can fix that myself too, whichever is the easiest option.

jansu76:
@niis-jenkins test this please

src/build_packages.sh
mbert: Sure, that had slipped my attention.

mbert:
Rebased and pushed again.

jansu76 (approved):

jansu76:
PR tests (build and Sonar) are at https://github.com/nordic-institute/X-Road/pull/819

PR #735, opened 2020-10-01
XRDDEV-1365: Stricter constraints for fetching service client candi…

TJaakkola:
…dates

Use equalsIgnoreCase when comparing instance and memberclass parameters while searching service client candidates

https://jira.niis.org/browse/XRDDEV-1365

### DIFF
--- a/src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/AccessRightService.java
+++ b/src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/AccessRightService.java
@@ -815,7 +815,7 @@ public class AccessRightService {
             if (xRoadId instanceof LocalGroupId) {
                 return true;
             } else {
-                return instance.equalsIgnoreCase((dto.getSubjectId().getXRoadInstance()));
+                return instance.equalsIgnoreCase(dto.getSubjectId().getXRoadInstance());
             }
         };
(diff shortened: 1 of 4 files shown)


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/AccessRightService.java
```diff
@@ -811,11 +811,11 @@ public DuplicateAccessRightException(String msg) {
         return dto -> {
             XRoadId xRoadId = dto.getSubjectId();
             // In case the Subject is a LocalGroup: LocalGroups do not have explicit X-Road instances
-            // -> always return
+            // -> always return true
             if (xRoadId instanceof LocalGroupId) {
                 return true;
             } else {
-                return StringUtils.containsIgnoreCase(dto.getSubjectId().getXRoadInstance(), instance);
+                return instance.equalsIgnoreCase((dto.getSubjectId().getXRoadInstance()));
```
carohauta, line 818: Double parenthesis not needed here 👀

TJaakkola: Oopsie. Fixed.

carohauta (approved):
🦄

PR #737, opened 2020-10-07
XRDDEV-1391: Fixes to UI breaking when there are long identifiers

TJaakkola:
https://jira.niis.org/browse/XRDDEV-1391


carohauta (approved):
🦄

PR #743, opened 2020-10-11
XRDDEV-1366 Add defensive checks against bad client configuration

petkivim:
Add defensive checks against bad client configuration: 

- Check member class when new local client is added.
- Check instance identifier when registering a client.
- Check member class when registering a client.
- Add a client with invalid instance identifier to the test configuration.
- Add a client with invalid member class to the test configuration.
- Add new tests for invalid member class.
- Add new tests for invalid instance identifier.

JIRA issue: https://jira.niis.org/browse/XRDDEV-1366


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/ClientService.java
```diff
@@ -493,13 +495,27 @@ public ClientType deleteTlsCertificate(ClientId id, String certificateHash)
...
+
+        String instanceIdentifier = client.getIdentifier().getXRoadInstance();
+        if (!instanceIdentifier.equals(globalConfFacade.getInstanceIdentifier())) {
+            throw new InvalidInstanceIdentifierException(INVALID_INSTANCE_IDENTIFIER + instanceIdentifier);
+        }
+
+        String memberClass = client.getIdentifier().getMemberClass();
+        if (!isValidMemberClass(memberClass)) {
+            throw new InvalidMemberClassException(INVALID_MEMBER_CLASS + memberClass);
+        }
```
carohauta, line 517: There are three methods in _ManagementRequestSenderService_ that take in a _ClientId_ parameter
* sendClientRegisterRequest
* sendClientUnregisterRequest
* sendOwnerChangeRequest

Would it make sense to check the validity of the _memberClass_ in all of these cases? Maybe moving the whole checking logic into _ManagementRequestSenderService_ methods listed above.

Or another way would be to just add the check to _unregisterClient_ and _changeOwner_ methods within _ClientService_.

petkivim: Originally I didn't include the member class check in `sendClientUnregisterRequest` and `sendOwnerChangeRequest`, because both requests require that the client has already been registered on the Security Server. And since the `sendClientRegisterRequest` has the check, registering a client with invalid member class should not be possible. So instead of including the check in all the three management requests, it's only included in the request that is a precondition for the two other requests.

carohauta (approved):
🦄

PR #747, opened 2020-10-14
XRDDEV-1233: Replace dtsgenerator with openapi-typescript-codegen

TJaakkola:
https://jira.niis.org/browse/XRDDEV-1233


src/proxy-ui-api/frontend/package.json
```diff
@@ -12,7 +12,7 @@
     "test:coverage": "vue-cli-service test:unit --coverage",
     "check-audit": "check-audit",
     "resolve-audit": "resolve-audit",
-    "generate-types": "./node_modules/typescript/bin/tsc --esModuleInterop scripts/dtsgen-wrapper.ts && node scripts/dtsgen-wrapper.js",
+    "generate-types": "openapi --input ../src/main/resources/openapi-definition.yaml --output ./src/openapi-types",
```
carohauta, line 15: Seems to generate a lot of stuff we don't need (like mock server and such). Maybe add these flags to only generate the models `--exportCore false --exportServices false`

carohauta: Also the [serve and build scripts](https://github.com/nordic-institute/X-Road/blob/a0a783742352287bdfbc44e58f3a565c5834f8e2/src/proxy-ui-api/frontend/package.json#L6-L7) try to remove `src/openapi-types.ts`. Should probably be removing `src/openapi-types` folder.

carohauta (changes requested):
Nice change! Only couple issues in `package.json`.

Not a concern of this task but it would be awesome to utilize the [runtime schema objects](https://github.com/ferdikoomen/openapi-typescript-codegen#runtime-schemas---exportschemas) this library is able to generate. With the schemas all user inputs could be validated against the constraints we have defined in the API spec. I tried it out of curiosity but there were some errors because of all the enums we have in the API spec.

src/proxy-ui-api/frontend/package.json
TJaakkola: Yeah nicely spotted. I wasn't sure whether those core and services folders could turn out usefull.

src/proxy-ui-api/frontend/package.json
```diff
@@ -3,16 +3,16 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "serve": "rm -f src/openapi-types.ts && npm run generate-types && vue-cli-service serve",
-    "build": "rm -f src/openapi-types.ts && npm run generate-types && vue-cli-service build",
+    "serve": "rm -f src/openapi-types && npm run generate-types && vue-cli-service serve",
+    "build": "rm -f src/openapi-types && npm run generate-types && vue-cli-service build",
```
carohauta, line 7:
```suggestion
    "serve": "rm -rf src/openapi-types && npm run generate-types && vue-cli-service serve",
    "build": "rm -rf src/openapi-types && npm run generate-types && vue-cli-service build",
```

carohauta (changes requested):

carohauta (approved):
🦄

PR #751, opened 2020-10-16
XRDDEV-1237: Set autofocus to several places in ui for better user experience

TJaakkola:
https://jira.niis.org/browse/XRDDEV-1237


Riippi (approved):
LGTM

PR #756, opened 2020-10-22
XRDDEV-742 rest transport protocol

jansu76:
https://jira.niis.org/browse/XRDDEV-742


jhyoty:
I find the "REST request without body" and "REST request body" naming a bit confusing. For example, in the pr-messtransport-message diagram, one can get the impression that those can be alternatives. Maybe "REST request header" and "REST request body" would convey the meaning better.

Also, I'd suggest that the diagram displayed SOAP and REST payloads as alternatives (it is explained in the text, but without prior knowledge and reading the text the diagram can give a wrong impression.  In the request, there has to be either SOAP request or REST request header. 

The response message diagram is also a bit misleading -- If there is a fault originating from the producer security server, there are none of the other parts. If it is not a Fault, then there must be either SOAP response or REST response header and at least the Signature. Now the diagram suggests that some of the optional parts can be present even in the case of Fault.

jansu76:
> I find the "REST request without body" and "REST request body" naming a bit confusing. For example, in the pr-messtransport-message diagram, one can get the impression that those can be alternatives. Maybe "REST request header" and "REST request body" would convey the meaning better.
> 
> Also, I'd suggest that the diagram displayed SOAP and REST payloads as alternatives (it is explained in the text, but without prior knowledge and reading the text the diagram can give a wrong impression. In the request, there has to be either SOAP request or REST request header.

Good suggestions, made updates based on these

jhyoty (approved):

PR #759, opened 2020-10-27
XRDDEV-1424 Update the ignore list of the frontend license check

Riippi:
https://jira.niis.org/browse/XRDDEV-1424

Add new type files created from openapi specifications to the ignore list of the frontend license check


Riippi:
This is a small change so I merged it by myself

PR #762, opened 2020-10-29
XRDDEV-1405: Save user module data to session storage instead of local storage

TJaakkola:
https://jira.niis.org/browse/XRDDEV-1405


Riippi (approved):
LGTM

PR #763, opened 2020-10-29
XRDDEV-1425 limit HSM CSR creation to SIGNING usage

jansu76:
https://jira.niis.org/browse/XRDDEV-1425

### DIFF
--- a/src/proxy-ui-api/frontend/src/store/modules/certificateSignRequest.ts
+++ b/src/proxy-ui-api/frontend/src/store/modules/certificateSignRequest.ts
@@ -35,6 +35,7 @@ import {
   KeyLabelWithCsrGenerate,
   KeyUsageType,
   CsrFormat,
+  TokenType,
 } from '@/openapi-types';
 import * as api from '@/util/api';
 import { encodePathParameter } from '@/util/api';
@@ -50,6 +51,7 @@ export interface CsrState {
   form: CsrSubjectFieldDescription[];
   keyLabel: string | undefined;
   tokenId: string | undefined;
+  tokenType: string | null;
   memberIds: string[];
   isNewMember: boolean;
 }
@@ -66,6 +68,7 @@ const getDefaultState = () => {
     form: [],
     keyLabel: undefined,
     tokenId: undefined,
+    tokenType: null,
     memberIds: [],
     isNewMember: false,
   };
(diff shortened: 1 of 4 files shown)


src/proxy-ui-api/frontend/src/store/modules/certificateSignRequest.ts
```diff
@@ -103,6 +106,10 @@ export const csrGetters: GetterTree<CsrState, RootState> = {
     return state.keyLabel;
   },
   isUsageReadOnly(state): boolean {
+    // if creating CSR for a hardware token, only sign CSRs can be created
+    if (state.tokenType == TokenType.HARDWARE) {
```
TJaakkola, line 110: In this case it doesn't matter but it's a good practise to always use three equal signs when doing a comparison in javascript.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Equality

For example. 

1 == '1' // true 
1 === '1' // false

src/proxy-ui-api/frontend/src/store/modules/certificateSignRequest.ts
```diff
@@ -194,6 +204,13 @@ export const actions: ActionTree<CsrState, RootState> = {
   setCsrTokenId({ commit }, tokenId: string) {
     commit('storeCsrTokenId', tokenId);
   },
+  setCsrTokenType({ commit }, tokenType: string) {
+    if (tokenType == TokenType.HARDWARE) {
```
TJaakkola, line 208: Strict equality like above

TJaakkola (changes requested):
A couple of minor changes required

TJaakkola (approved):

src/proxy-ui-api/frontend/src/store/modules/certificateSignRequest.ts
jansu76: fixed

src/proxy-ui-api/frontend/src/store/modules/certificateSignRequest.ts
jansu76: fixed

PR #765, opened 2020-11-01
XDDEV-1396 Frontend library

Riippi:
https://jira.niis.org/browse/XRDDEV-1396
https://confluence.niis.org/display/XRDDEV/Front-end+UI+framework

Merging to frontend-ui-library so it's possible to review the changes even though this cannot be added to develop branch at this point.


src/shared-ui/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "vue-template-compiler": "^2.6.11",
+    "vuetify-loader": "^1.3.0",
+    "webpack-node-externals": "^2.5.2"
+  },
+  "peerDependencies": {
+    "core-js": "^3.6.5",
+    "vue": "^2.6.11",
+    "vuetify": "^2.2.11",
+    "vue-i18n": "^8.21.0",
+    "vee-validate": "^3.4.1"
```
TJaakkola, line 66: I doubt these dependencies should be both in dependencies and peerDependencies?

TJaakkola (changes requested):

src/shared-ui/package.json
Riippi: They need to be in both because peerDependencies are not installed. Those are demands for the project that includes the library.

TJaakkola: I did some more investigation and got the meaning of peerDependencies better now. So this is ok.

TJaakkola (approved):

PR #768, opened 2020-11-04
XRDDEV-1403

jhyoty:


jansu76 (approved):
Pretty complex stuff and hard to understand in-depth in a casual review. As far as I could understand, looks good.

PR #770, opened 2020-11-04
XRDDEV-1397 Ui components

Riippi:
https://jira.niis.org/browse/XRDDEV-1397
Components that can be easily share between projects are moved to UI library


TJaakkola:
Seems there are some linter issues, those could be fixes before merging.

TJaakkola:
Unit tests for the shared-ui doesn't seem to pass. That should be fixed

Riippi:
Fixed unit tests and linter issues.

TJaakkola (approved):

PR #773, opened 2020-11-05
XRDDEV-1368: Improve token status indicators

TJaakkola:
https://jira.niis.org/browse/XRDDEV-1368


src/proxy-ui-api/frontend/src/components/ui/Expandable.vue
```diff
@@ -49,15 +56,20 @@
...
 import Vue from 'vue';
+import { Prop } from 'vue/types/options';
 
 export default Vue.extend({
   name: 'expandable',
   components: {},
   props: {
     isOpen: {
-      type: Boolean,
+      type: Boolean as Prop<boolean>,
```
Riippi, line 66: As far as I know there is no benefit on adding the "as Prop" for basic types like boolean or string. 
At least the vscode knows the prop types without this.  
For the interfaces like Token this is useful.

Riippi (changes requested):

src/proxy-ui-api/frontend/src/views/KeysAndCertificates/SignAndAuthKeys/TokenExpandable.vue
```diff
@@ -144,6 +149,12 @@ import * as api from '@/util/api';
 import { encodePathParameter } from '@/util/api';
 import FileUpload from '@/components/ui/FileUpload.vue';
 import { FileUploadResult } from '@/ui-types';
+import TokenLoggingButton from '@/views/KeysAndCertificates/SignAndAuthKeys/TokenLoggingButton.vue';
```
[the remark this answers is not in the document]
TJaakkola: Yeah that's true but this button is also about logout. Don't know if that makes anymore sense but that's why I came up with this name. TokenLoginLogoutButton sounds confusing to me.

src/proxy-ui-api/frontend/src/components/ui/Expandable.vue
TJaakkola: I removed the Prop but then again it would be more consistent if we just add the typescript typing to all props instead of just part of them.

Riippi (approved):
LGTM

PR #779, opened 2020-11-11
XRDDEV-1457: Fix broken certificate details link

TJaakkola:
https://jira.niis.org/browse/XRDDEV-1457

### DIFF
--- a/src/proxy-ui-api/frontend/src/views/CertificateDetails/CertificateDetails.vue
+++ b/src/proxy-ui-api/frontend/src/views/CertificateDetails/CertificateDetails.vue
@@ -134,7 +134,7 @@ export default Vue.extend({
     },
     usage: {
       type: String,
-      required: true,
+      required: false,
     },
   },
   data() {
--- a/src/proxy-ui-api/frontend/src/router.ts
+++ b/src/proxy-ui-api/frontend/src/router.ts
@@ -291,7 +291,7 @@ const router = new Router({
         },
         {
           name: RouteName.Certificate,
-          path: '/certificate/:hash/:usage',
+          path: '/certificate/:hash',
           components: {
             default: CertificateDetails,
           },
--- a/src/proxy-ui-api/frontend/src/views/KeysAndCertificates/SignAndAuthKeys/CertificateRow.vue
+++ b/src/proxy-ui-api/frontend/src/views/KeysAndCertificates/SignAndAuthKeys/CertificateRow.vue
@@ -29,7 +29,7 @@
       <div class="name-wrap">
         <i
           class="icon-xrd_certificate icon clickable"
-          @click="certificateClick(cert, key)"
+          @click="certificateClick()"
         ></i>
         <div class="clickable-link" @click="certificateClick()">
           {{ cert.certificate_details.issuer_common_name }}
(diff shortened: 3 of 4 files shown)


src/proxy-ui-api/frontend/src/views/CertificateDetails/CertificateDetails.vue
```diff
@@ -134,7 +134,7 @@ export default Vue.extend({
     },
     usage: {
       type: String,
-      required: true,
+      required: false,
```
jansu76, line 137: I wanted to review this PR, but I don't think I understand frontend well enough yet.

In what situations is usage missing? Do the checks such as
```
        if (this.usage === KeyUsageType.SIGNING) {
        } else {
        }

```
work correctly when that happens?

TJaakkola: Whenever a keys usage -field is undefined. I'm not expert on how that can happen but it is possible when a hardware tokens keys certificate is not imported. 

That kind of logic may cause bugs as if usage is undefined the logic goes to else block which could be though as the block for KeyUsageType.AUTHENTICATION.

jansu76: It might be safer to replace code such as 

```
    showActivate(): boolean {
      if (this.certificate === null) {
        return false;
      }

      if (this.possibleActions.includes(PossibleAction.ACTIVATE)) {
        if (this.usage === KeyUsageType.SIGNING) {
          return this.$store.getters.hasPermission(
            Permissions.ACTIVATE_DISABLE_SIGN_CERT,
          );
        } else {
          return this.$store.getters.hasPermission(
            Permissions.ACTIVATE_DISABLE_AUTH_CERT,
          );
        }
      }

      return false;
    },
```

with 

```
    showActivate(): boolean {
      if (this.certificate === null) {
        return false;
      }

      if (this.possibleActions.includes(PossibleAction.ACTIVATE)) {
        if (this.usage === KeyUsageType.SIGNING) {
          return this.$store.getters.hasPermission(
            Permissions.ACTIVATE_DISABLE_SIGN_CERT,
          );
        } else if (this.usage === KeyUsageType.AUTHENTICATION) {
          return this.$store.getters.hasPermission(
            Permissions.ACTIVATE_DISABLE_AUTH_CERT,
          );
        }
      }

      return false;
    },
```
PossibleActions might already "take care of this" but there shouldn't be any downside to doing this, and it is a bit safer.

TJaakkola: I agree that this would be an improvement, but I'm a bit hesitant whether such refactoring in this task?

jansu76: I thought about it and ninja-removed my comment. It is not as simple as that, indeed. Better not change it too fast, I'll slack more details.

jansu76: This comment is better handled in a separate ticket, I will create one. 
This PR seems good to go.

PR #786, opened 2020-11-19
XRDDEV-1476

TJaakkola:
Prevent user from editing token friendly name if it is not possible


jansu76 (approved):
LGTM. Even tested it.

PR #790, opened 2020-11-21
XRDDEV-1486

petkivim:
- Display key id in friendly name field when:
  - Name is an empty string.
  - Name property does not exist.
- Make `label` property required in the `KeyLabel` OpenAPI model.
  - Set the minimum lenght of the property to zero so that label can be an empty string.
  - Add one negative test case.

JIRA: https://jira.niis.org/browse/XRDDEV-1486

### DIFF
--- a/src/proxy-ui-api/frontend/src/views/KeyDetails/KeyDetails.vue
+++ b/src/proxy-ui-api/frontend/src/views/KeyDetails/KeyDetails.vue
@@ -295,7 +295,7 @@ export default Vue.extend({
         .finally(() => (this.deleting = false));
     },
     setKeyName(): void {
-      if (this.key.name === '') {
+      if (this.key.name === undefined || this.key.name === '') {
         this.key.name = this.key.id;
       }
     },
--- a/src/proxy-ui-api/frontend/src/views/KeyDetails/KeyDetails.vue
+++ b/src/proxy-ui-api/frontend/src/views/KeyDetails/KeyDetails.vue
@@ -145,6 +145,7 @@ import { encodePathParameter } from '@/util/api';
 import WarningDialog from '@/components/ui/WarningDialog.vue';
 import { mapGetters } from 'vuex';
 import { PossibleActions } from '@/openapi-types/models/PossibleActions';
+import { isEmpty } from '@/util/helpers';
 
 export default Vue.extend({
   components: {
@@ -295,7 +296,7 @@ export default Vue.extend({
         .finally(() => (this.deleting = false));
     },
     setKeyName(): void {
-      if (this.key.name === undefined || this.key.name === '') {
+      if (isEmpty(this.key.name)) {
         this.key.name = this.key.id;
       }
     },
--- a/src/proxy-ui-api/src/main/resources/openapi-definition.yaml
+++ b/src/proxy-ui-api/src/main/resources/openapi-definition.yaml
@@ -6020,13 +6020,16 @@ components:
         name: my-key-0
     KeyLabel: # ok
       type: object
+      required:
+        - label
       properties:
         label:
           type: string
           format: text
           description: label for the new key
-          minLength: 1
+          minLength: 0
           maxLength: 255
+          example: My new key
       example:
         label: My new key
     KeyLabelWithCsrGenerate: # ok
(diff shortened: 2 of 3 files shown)


src/proxy-ui-api/src/main/resources/openapi-definition.yaml
```diff
@@ -6020,13 +6020,16 @@ components:
...
       type: object
+      required:
+        - label
       properties:
         label:
           type: string
           format: text
           description: label for the new key
-          minLength: 1
+          minLength: 0
```
jansu76, line 6030: Such a small change, yet I spent a long time in a rabbit hole, thinking about it....

**Null vs empty string for "value does not exist"**

Looks like this detail has not had enough attention. Current develop is a bit inconsistent in this. For example adding a new key (only key, without CSR) - which UI does not use - with `POST /tokens/:id/keys` allows both `null` and `""` for key label. These go into database accordingly, as either null or empty string. 

ServiceDescription.disabled_notice works the same way. Using API you can set it to null or empty string. 

Or UI currently sets empty strings for disabled_notice, and key label (when a different endpoint is used).

Since there does not seem to be semantic difference between null and empty label, it is a bit bad that we offer API users possibility of setting either value. It would be better if null / empty string handling was consistent, and we picked one of the 2 options.

This PR opted for using empty strings, I would actually recommend **using nulls instead**. I think null more accurately represents the semantics of "no disabled notice", "no key label", "no value present for xyz". But regardless of which one we pick, consistency and predictability is the most important thing.

Besides key label and service description disabled notice, there may be other properties to go through as well. Based on a quick check, there are nullable varchar fields in tables `identifier`, `service`, `tsp` and `locale` too. And some may not be about database tables (like key.label)

One aspect is that old application used empty strings instead of nulls in DB, for disabled_notice at least. And we probably have both nulls and empty strings stored. But the API does not have to use the same mapping (null vs "") and it can (and probably should) also work so that it writes one style, and reads both styles. 

**API versioning**

If we change the API regarding this, we may be making breaking changes. This PR made null KeyLabel.label forbidden, it would break some API client that was using nulls for KeyLabel.label. Likewise if we refactor key label and service description disabled_notice and possibly some other properties too so that empty strings are forbidden and you should use nulls instead, we would break some API clients that were using empty strings.

In principle we should release a v2 of the API, or do something else to take care of versioning.  Not sure if this detail is so insignificant that we would instead break the principle.

**Wrap up**

Maybe could consider removing API change from this PR and creating a separate ticket for making a more system-wide solution to handling of nulls vs empty strings.

src/proxy-ui-api/frontend/src/views/KeyDetails/KeyDetails.vue
```diff
@@ -295,7 +295,7 @@ export default Vue.extend({
         .finally(() => (this.deleting = false));
     },
     setKeyName(): void {
-      if (this.key.name === '') {
+      if (this.key.name === undefined || this.key.name === '') {
```
jansu76, line 298: This does not actually check for `null`s but instead `undefined`s. Still it does work in practice. 

There is also helpers.ts#isEmpty that could be used for this kind of check.

jansu76 (changes requested):

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
petkivim: Thanks for raising some very good points. I agree that all the endpoints should follow the same approach consistently regardless of which alternative is eventually chosen. I reverted the API design related changes implemented in this task.

Let's create a separate ticket for making a system-wide solution to handling of nulls vs empty strings.

src/proxy-ui-api/frontend/src/views/KeyDetails/KeyDetails.vue
petkivim: Changed this to `helpers.ts#isEmpty`.

petkivim:
Successful build: https://jenkins.niis.org/job/pr-build-pipeline/1666/

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
jansu76: I created separate ticket for that: https://jira.niis.org/browse/XRDDEV-1500

jansu76 (approved):

PR #791, opened 2020-11-24
XRDDEV-1481

TJaakkola:
Improved error message for when user tries to register member while token is not logged in.

https://jira.niis.org/browse/XRDDEV-1481

### DIFF
--- a/src/signer/src/main/java/ee/ria/xroad/signer/protocol/handler/GetMemberSigningInfoRequestHandler.java
+++ b/src/signer/src/main/java/ee/ria/xroad/signer/protocol/handler/GetMemberSigningInfoRequestHandler.java
@@ -68,7 +68,8 @@ public class GetMemberSigningInfoRequestHandler extends AbstractRequestHandler<G
 
         if (memberKeys.isEmpty()) {
             throw CodedException.tr(X_UNKNOWN_MEMBER, "member_certs_not_found",
-                    "Could not find any certificates for member '%s'", message.getMemberId());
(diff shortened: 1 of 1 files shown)


src/signer/src/main/java/ee/ria/xroad/signer/protocol/handler/GetMemberSigningInfoRequestHandler.java
```diff
@@ -68,7 +68,8 @@ protected Object handle(GetMemberSigningInfo message) throws Exception {
 
         if (memberKeys.isEmpty()) {
             throw CodedException.tr(X_UNKNOWN_MEMBER, "member_certs_not_found",
-                    "Could not find any certificates for member '%s'", message.getMemberId());
+                    "Could not find any certificates for member '%s'. "
+                        + "Are you sure tokens containing the certifications are logged in?", message.getMemberId());
```
jansu76, line 72: Making modifications to texts shown in UI by changing signer code, which is not aware of UI, and has also other non-UI users, is a bit suspect. This exception is also thrown at least to `xroad-proxy`, `signer-cli`, and in theory could by unwanted in those. It is just a text change so the probability of it breaking things should be minimal, though.

Maybe we should add a way to define core error messages (or overrides to default messages) inside proxy-ui-api module. 

Let's work on this if we get more of these UI-rooted needs to change error messages. Now next release is quite soon, there's no sense in opening this issue more at this point.

jansu76 (approved):

PR #794, opened 2020-11-25
XRDDEV-1449

iluwatar:
- Add tests for different user roles checking that navigation between the main pages and tabs is successful


TJaakkola:
The changes in this PR seem to be ok. However I'd like to see the page objects for keys and certificates, settings and Diagnostics -pages to be split to their own files as has been done with endpoints page and service clients -pages. 

Current solutions work but pushing all the accessors and commands to one file will cause it to become huge and difficult to read.

TJaakkola (approved):

PR #795, opened 2020-11-26
XRDDEV-1503: Disable generate csr -button after clicking and only ena…

TJaakkola:
…ble it if creating csr fails

https://jira.niis.org/browse/XRDDEV-1503


Riippi (approved):
LGTM

PR #797, opened 2020-12-02
XRDDEV-1395 cache TTL

jansu76:
https://jira.niis.org/browse/XRDDEV-1395

### DIFF
--- a/src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/config/TtlConcurrentMapCacheConfiguration.java
+++ b/src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/config/TtlConcurrentMapCacheConfiguration.java
@@ -1,3 +1,28 @@
+/**
+ * The MIT License
+ * Copyright (c) 2019- Nordic Institute for Interoperability Solutions (NIIS)
+ * Copyright (c) 2018 Estonian Information System Authority (RIA),
+ * Nordic Institute for Interoperability Solutions (NIIS), Population Register Centre (VRK)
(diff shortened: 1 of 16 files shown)


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/config/TtlConcurrentMapCacheConfiguration.java
```diff
@@ -0,0 +1,75 @@
...
+ * ConcurrentMapCache autoconfiguration.
+ *
+ * storeByValue = true is not supported. If that is needed, either change this configuration or use
+ * e.g. Caffeine (perhaps better option)
+ */
+@Configuration
+@Slf4j
+public class TtlConcurrentMapCacheConfiguration {
+
+    public static final int TTL_SECONDS = 60;
```
Dos4dev, line 52: The value used in configuration files can easily be changed, and different profiles can use different values. Mey, I ask, why not get this from the properties file?

jansu76: Good point. I think I will change this.

jansu76: Changed in https://github.com/nordic-institute/X-Road/pull/797/commits/67e80cf72d39b41e7e55ffc56061888f5af28560

Dos4dev: Thank you for the fast response.

jhyoty (approved):

PR #801, opened 2020-12-08
XRDDEV-1494 v7 UI update for keys & certificates

Riippi:
https://jira.niis.org/browse/XRDDEV-1494


src/proxy-ui-api/frontend/src/views/KeysAndCertificates/SecurityServerTlsCertificate/SecurityServerTlsCertificate.vue
```diff
@@ -169,6 +183,7 @@ export default Vue.extend({
       this.$router.push({
         name: RouteName.GenerateInternalCSR,
       });
+      // this.showGenCsr = true;
```
TJaakkola, line 186: In general commented code should be removed but this line seems to be added in this file and I can't see that statement with a red background so why is this added in this PR? If it's not needed it should be removed.

TJaakkola (changes requested):

src/proxy-ui-api/frontend/src/views/KeysAndCertificates/SecurityServerTlsCertificate/SecurityServerTlsCertificate.vue
Riippi: Removed the line

TJaakkola (approved):

PR #806, opened 2020-12-12
XRDDEV-1465 

petkivim:
Update Keys and Certificates view search field label from "Service" to "Search". Other top level views (Clients + Backup and Restore) use the same label in the search field.

JIRA: https://jira.niis.org/browse/XRDDEV-1465


petkivim:
Build failed because of NPM audit.

PR #822, opened 2021-01-05
XRDDEV-1510 Null check for ocspBytes

carohauta:
Adds a null check for `ocspBytes` to avoid showing the ocsp status wrong and throwing an NPE

JIRA: https://jira.niis.org/browse/XRDDEV-1510


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/util/ClientUtils.java
```diff
@@ -52,9 +52,9 @@ public static boolean hasValidLocalSignCert(ClientId clientId, List<CertificateI
             String ocspResponseStatus = null;
             try {
                 ocspResponseStatus = OcspUtils.getOcspResponseStatus(certificateInfo.getOcspBytes());
-            } catch (OcspUtils.OcspStatusExtractionException | RuntimeException e) {
+            } catch (OcspUtils.OcspStatusExtractionException e) {
                 log.error(ERROR_OCSP_EXTRACT_MSG + " for client: " + clientId.toString(), e);
-                return false;
+                break;
             }
             if (clientId.memberEquals(certificateInfo.getMemberId())
```
jansu76, line 59: Come to think of it, I would recommend moving this memberEquals check to be done as first thing in `hasValidLocalSignCert` for each certificateInfo. If memberEquals == false we don't need to inspect the cert any further, and can proceed to next candidate.

Now for each call to `hasValidLocalSignCert(client x)` it potentially scans through all certificates, and goes to `OcspUtils.getOcspResponseStatus()` to parse OCSP responses from bytes, even if the certificate has nothing to do with client x. For main clients list this can result in magnitude of `n*m` unnecessary getOcspResponseStatus() calls (n = number of clients, m = number of certs).

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/util/ClientUtils.java
```diff
@@ -52,9 +52,9 @@ public static boolean hasValidLocalSignCert(ClientId clientId, List<CertificateI
             String ocspResponseStatus = null;
             try {
                 ocspResponseStatus = OcspUtils.getOcspResponseStatus(certificateInfo.getOcspBytes());
```
jansu76, line 54: A test might be justified, since there is a bit of complexity in this method (and the first implementation got it wrong) - maybe a test that would have failed before the fix, and passes after the fix?

jansu76 (changes requested):

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/util/ClientUtils.java
carohauta: Moved the comparison up in the method

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/util/ClientUtils.java
carohauta: Added a test

jansu76 (approved):

jansu76:
Changes are good, PR fixes the NPE problem.

Currently there is still some open discussion about how client wizard works. Not sure if this should already be merged, or if we should wait to for clarity on that.

Related: https://niis.slack.com/archives/GB2JH4LUE/p1611669639013000

PR #824, opened 2021-01-05
Switch the add wsdl and add rest permissions

Riippi:
Switches the permissions used for "add WSDL" and "add REST" buttons.
Doesn't affect UI functionality.


Riippi:
Merging this by myself as it's just two lines

PR #838, opened 2021-01-18
XRDDEV-1526

TJaakkola:
- E2E-tests for backup and restore functionality
- Fix taking screenshots configuration
- Set default size for the browser bigger so that tests work better

https://jira.niis.org/browse/XRDDEV-1526


src/proxy-ui-api/frontend/tests/e2e/pages/tabs/settingsTab.js
```diff
@@ -39,6 +41,61 @@ var settingsTabCommands = {
...
+  clickDeleteForBackup: function (backupFilename) {
+    this.click(
+      'xpath',
+      `//table[contains(@class, "xrd-table")]/tr/td[text() = "${backupFilename}"]/..//button[contains(@data-test, "backup-delete")]`,
+    );
+    return this;
+  },
+};
+
+var confirmationDialog = {
```
carohauta, line 88: Could maybe use `const`

src/proxy-ui-api/frontend/tests/e2e/specs/ss-backupandrestore.js
```diff
@@ -0,0 +1,270 @@
...
+
+    // Get the backend-generated name of the backup and close snackbar
+    const createdBackupFileNameTextObject = await browser.getText(
+      'xpath',
+      mainPage.elements.snackBarMessage,
+    );
+    const createdBackupFileName = createdBackupFileNameTextObject.value.split(
+      ' ',
+    )[1];
+    console.log('Created backup: ', createdBackupFileName);
```
carohauta, line 115: Is this `console.log` needed?

src/proxy-ui-api/frontend/tests/e2e/specs/ss-backupandrestore.js
```diff
@@ -0,0 +1,270 @@
...
+    mainPage.closeSnackbar();
+
```
carohauta, line 178: Could the backup creation be in a separate function since it is used in multiple tests? Also could the backup only be created once maybe in the first test case and then deleted in the last test case (where the backup is now created and deleted for a third time)?

src/proxy-ui-api/frontend/tests/e2e/specs/ss-backupandrestore.js
```diff
@@ -0,0 +1,270 @@
...
+        true,
+        `Backup file should exist in: ${testDataDir}${createdBackupFileName}`,
+      ),
+    );
+
+    // Import the created backup from local filesystem (first cancel the operation)
+    backupAndRestoreTab.addBackupToInput(testDataDir + createdBackupFileName);
+    browser.waitForElementVisible(backupFileAlreadyExistsDialog);
+    backupFileAlreadyExistsDialog.cancel();
+    browser.expect.element(backupFileAlreadyExistsDialog).to.not.be.visible;
```
carohauta, line 195: Not sure but wouldn't `waitForElementNotVisible(...)` actually wait (10000ms) for the element to not be visible where as `.to.not.be.visible` expects the element to not be visible immediately? Should `waitForElementNotVisible` be preferred in these cases?

src/proxy-ui-api/frontend/tests/e2e/pages/tabs/settingsTab.js
TJaakkola: Agree. This and all other var declarations changed to const.

src/proxy-ui-api/frontend/tests/e2e/specs/ss-backupandrestore.js
TJaakkola: Not necessarily but I left it as it could prove usefull in some situation when trying to figure out why this test is failing. If it's not logged the name of the created backup will be lost. Didn't remove it yet but I don't have strong opinions about it so if you feel it should be removed then we will.

src/proxy-ui-api/frontend/tests/e2e/specs/ss-backupandrestore.js
TJaakkola: Sounds reasonable. Changed.

src/proxy-ui-api/frontend/tests/e2e/specs/ss-backupandrestore.js
TJaakkola: Good point, not sure why I didn't separate that block to a function. 

I'm not sure what would be best practise, but I've preferred changing the state of test environment in each test separately to the wanted state so that adjacent tests won't become dependent on the previous being successfull. In this case it might have been a better approach but when implementing a larger test set it becomes really complicated to have that kind of expected state. I think it would make tests easier to be broken by accident. So I've decided to be consistent on this.

src/proxy-ui-api/frontend/tests/e2e/specs/ss-backupandrestore.js
carohauta: I guess there's no need to remove it since it is used elsewhere in the tests also and it's not like it clutters any production logs anyway

carohauta (approved):
🦄

PR #844, opened 2021-01-22
XRDDEV-1504: escape log messages

jhyoty:


jhyoty:
rescoped to auditlog only

PR #846, opened 2021-01-25
e2e test fixes v6.26

carohauta:
e2e environment was upgraded from X-Road 6.24 to 6.26. This upgrade introduced frontend changes that broke some of the e2e tests. This PR fixes those tests. 

This PR also changes Add Key and CSR wizard's `Done` buttons `disabled` status behavior: the button will be enabled after the API request returns success (instead of being enabled right after the request is sent)

### DIFF
--- a/src/proxy-ui-api/frontend/tests/e2e/specs/ss-backupandrestore.js
+++ b/src/proxy-ui-api/frontend/tests/e2e/specs/ss-backupandrestore.js
@@ -134,7 +134,6 @@ module.exports = {
     // Filtering backup list with the name of created backup there should be only one backup in the list
     backupAndRestoreTab.enterFilterInput(createdBackupFileName);
     browser.expect
-      // xrd-table has thead but no tbody so use '/tr' to get only the row elements that are direct children of table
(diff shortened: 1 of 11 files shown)


src/proxy-ui-api/frontend/tests/e2e/specs/ss-backupandrestore.js
```diff
@@ -135,7 +135,7 @@ module.exports = {
     backupAndRestoreTab.enterFilterInput(createdBackupFileName);
     browser.expect
       // xrd-table has thead but no tbody so use '/tr' to get only the row elements that are direct children of table
-      .elements('//table[contains(@class, "xrd-table")]/tr')
+      .elements('//table[contains(@class, "xrd-table")]/tbody/tr')
```
olliru, line 138: This code no longer matches the comment above

olliru (approved):

src/proxy-ui-api/frontend/tests/e2e/specs/ss-backupandrestore.js
carohauta: Removed the comment

PR #847, opened 2021-01-25
XRDDEV-1541 add subject dialog e2e

carohauta:
e2e tests for adding a new Service Client

JIRA: https://jira.niis.org/browse/XRDDEV-1541


olliru (approved):
Some of the locators have long absolute parts in the xpaths, like "/../../../../..//button". If an alternative and less fragile xpath is available, it would be good to use it.

carohauta:
> Some of the locators have long absolute parts in the xpaths, like "/../../../../..//button". If an alternative and less fragile xpath is available, it would be good to use it.

That's true. Found a better way for all of them. Now using max. of one `/../` like in the rest of the tests

PR #849, opened 2021-01-25
XRDDEV-1539 Alerts v7 UI update

Riippi:
https://jira.niis.org/browse/XRDDEV-1539

- Replaced notifications bar with global alerts and contextual alerts
- New style for snackbar
- Simple dialog component has a possibility to show error. Not used at the moment.


TJaakkola (changes requested):
Couple of fixes that are probably not part of this refactoring but we could as well fix those now.

TJaakkola (approved):

PR #857, opened 2021-01-27
XRDDEV-1499 lighter npm audit

jansu76:
https://jira.niis.org/browse/XRDDEV-1499

See also https://bitbucket.niis.org/projects/X-ROAD/repos/x-road-utils/pull-requests/145/overview, both should be merged at the same time

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -39,7 +39,7 @@ pipeline {
             steps {
                 sh 'cd src && ./update_ruby_dependencies.sh'
                 withCredentials([string(credentialsId: 'sonarqube-user-token-2', variable: 'SONAR_TOKEN')]) {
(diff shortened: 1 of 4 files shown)


jhyoty (commented):
Works, but looks unnecessarily complex. Consider using e.g. project property, it seems that testing that in the checkFrontAudit would have been sufficient; does not need playing with the conditional task dependencies, and it also avoids the extra logic in CI pipeline (just adding the property is enough). For example something like
```   
task checkFrontAudit(...)
...
    if ("strict" == project.properties['front-audit-mode']) {
      args=['run', 'check-audit']
    } else {
      args=['audit', '--production', '--audit-level=moderate']
    }
...
```
and then `gradle .... -Pfront-audit-mode=strict`

jhyoty (approved):

PR #860, opened 2021-01-29
XRDDEV-1537

iluwatar:


src/common-util/src/main/java/ee/ria/xroad/common/Version.java
```diff
@@ -78,6 +82,48 @@
...
+            // Java 8 or lower has format: 1.6.0_23, 1.7.0, 1.7.0_80, 1.8.0_211
+            versionString = versionString.substring(2, VERSION_STRING_SUFFIX_LENGTH);
+        } else {
+            // Java 9 or higher has format: 9.0.1, 11.0.4, 12, 12.0.1
+            int dot = versionString.indexOf(".");
+            if (dot != -1) {
+                versionString = versionString.substring(0, dot);
+            }
+        }
+        int result = Integer.parseInt(versionString);
```
jhyoty, line 113: This could fail with numberformatexception.

src/common-util/src/main/java/ee/ria/xroad/common/Version.java
```diff
@@ -78,6 +82,48 @@
...
+        String vendorString = System.getProperty(JAVA_VENDOR_PROPERTY);
+        String versionString = System.getProperty(JAVA_VERSION_PROPERTY);
+        String runtimeString = System.getProperty(JAVA_RUNTIME_PROPERTY);
+        StringBuilder vendorVersionBuilder = new StringBuilder();
+        if (vendorString != null) {
+            vendorVersionBuilder.append(vendorString + " ");
+        }
+        vendorVersionBuilder.append(runtimeString);
+        System.out.println(
+                String.format("%s %s (Java %s)", appName, XROAD_VERSION, vendorVersionBuilder.toString()));
```
jhyoty, line 101: Instead of static  "Java", the `java.runtime.name` system property would be more informative (defaulting to "Java" if it does not exist). The "java.runtime.version" could also default to "java.version", which is guaranteed to exist.

System.out.printf("...\n") can be used instead of ....println(String.format(..)). Or maybe this should use a logger instead; the version warnings are logged, so why not also the version info?

src/common-util/src/main/java/ee/ria/xroad/common/Version.java
```diff
@@ -78,6 +82,48 @@
...
+            // Java 9 or higher has format: 9.0.1, 11.0.4, 12, 12.0.1
+            int dot = versionString.indexOf(".");
+            if (dot != -1) {
+                versionString = versionString.substring(0, dot);
+            }
+        }
+        int result = Integer.parseInt(versionString);
+        if (result < minJavaVersion || result > maxJavaVersion) {
+            if (minJavaVersion == maxJavaVersion) {
+                log.warn(String.format(
```
jhyoty, line 116: Minor issue: the logger already supports simple message formatting (`log.warn("some value: {}", value)`, String.format is not needed here).

src/common-util/src/main/java/ee/ria/xroad/common/Version.java
```diff
@@ -78,6 +82,48 @@
...
+    /**
+     * Outputs version information and a warning if JVM version is not in the given range
+     * @param minJavaVersion the minimum supported version
+     * @param maxJavaVersion the maximum supported version
+     */
+    public static void outputVersionInfo(String appName, int minJavaVersion, int maxJavaVersion) {
+        // print app name + version and java vendor name + runtime version
+        String vendorString = System.getProperty(JAVA_VENDOR_PROPERTY);
+        String versionString = System.getProperty(JAVA_VERSION_PROPERTY);
+        String runtimeString = System.getProperty(JAVA_RUNTIME_PROPERTY);
```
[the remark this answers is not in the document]
iluwatar: Updated variable names

src/common-util/src/main/java/ee/ria/xroad/common/Version.java
iluwatar: Added exception handling

src/common-util/src/main/java/ee/ria/xroad/common/Version.java
iluwatar: Added fallbacks/defauls to system properties. Fixed system.out -> log.info

src/common-util/src/main/java/ee/ria/xroad/common/Version.java
iluwatar: Updated according to the suggestion

jhyoty (approved):

PR #866, opened 2021-02-01
XRDDEV-1559

TJaakkola:
… register new client if there is no active certificate for the member

https://jira.niis.org/browse/XRDDEV-1559

### DIFF
--- a/src/proxy-ui-api/frontend/src/views/AddClient/FinishPage.vue
+++ b/src/proxy-ui-api/frontend/src/views/AddClient/FinishPage.vue
@@ -107,12 +107,12 @@ export default Vue.extend({
       'currentSecurityServer',
       'tokens',
     ]),
-    showRegisterOption() {
+    showRegisterOption(): boolean {
       return (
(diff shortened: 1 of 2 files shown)


src/proxy-ui-api/frontend/src/views/AddClient/FinishPage.vue
```diff
@@ -37,7 +37,7 @@
       <p>{{ $t('wizard.finish.note') }}</p>
       <p></p>
 
-      <div v-if="showRegisterOption">
+      <div v-if="showRegisterOption & canRegisterClient">
```
Riippi, line 40: Should this be && ?

Riippi (changes requested):

src/proxy-ui-api/frontend/src/views/AddClient/FinishPage.vue
TJaakkola: yes, fixed

PR #868, opened 2021-02-03
XRDDEV-1563

TJaakkola:
Add v-if to all usages of v-dialog component to remove it from the DOM while it is not visible in the UI

https://jira.niis.org/browse/XRDDEV-1563

### DIFF
--- a/src/proxy-ui-api/frontend/src/components/client/SelectClientDialog.vue
+++ b/src/proxy-ui-api/frontend/src/components/client/SelectClientDialog.vue
@@ -24,7 +24,7 @@
    THE SOFTWARE.
  -->
 <template>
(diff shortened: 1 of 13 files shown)


src/proxy-ui-api/frontend/src/components/client/SelectClientDialog.vue
```diff
@@ -24,7 +24,7 @@
    THE SOFTWARE.
  -->
 <template>
-  <v-dialog :value="dialog" width="750" scrollable persistent>
+  <v-dialog v-if="dialog" :value="dialog" width="750" scrollable persistent>
```
Riippi, line 27: A bit theoretical note but is there any benefit keeping :value="dialog"  in these? I guess it works just similarly without them.

PR #873, opened 2021-02-08
XRDDEV-1477 usability issues

Riippi:
https://jira.niis.org/browse/XRDDEV-1477

Addressed usability issues in "add member" and "add subject" dialogs.
- selected items are cleared if the search is refreshed
- more vertical space for the list
- clear icons added also for text fields
- Search button has loading status and the list is cleared on search so the user knows what is going on.


olliru (approved):
Looks ok

PR #883, opened 2021-02-17
XRDDEV-1583: add xrd-prefix to common components

TJaakkola:
Add xrd-prefix to all shared-ui components so that is more clear what components are being used

https://jira.niis.org/browse/XRDDEV-1583


Riippi:
I think we could rename the `xrd-large-button` to just `xrd-button` because there is no longer a small version?

TJaakkola:
Yeah, sounds reasonable. I've made that change.

Riippi (approved):
LGTM

PR #884, opened 2021-02-18
XRDDEV-1571

amoog:
added more fields to version response, simple versionservice tests


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/VersionService.java
```diff
@@ -28,24 +28,37 @@
...
+
 
     /**
-     * Returns X-Road software version number
+     * Returns X-Road software version number and java version information
      * @return
      */
-    public String  getVersion() {
-        return Version.XROAD_VERSION;
+    public VersionInfo getVersionInfo() {
```
carohauta, line 50: Using an OpenAPI Model Object in the service layer – instead, a separate DTO should be used and converted in the controller layer.

To have a clear separation between the service and controller layers, it was decided to avoid using OpenAPI models in the service layer. More info about the architectural decisions here: https://confluence.niis.org/display/XRDDEV/API+backend+implementation

carohauta (changes requested):
Also, there is an npm audit issue that should be fixed by merging `develop-7.0` back into this branch

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/service/VersionService.java
amoog: I definitely and absolutely do not agree with this "architectural" decision but will make the change.

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/service/VersionServiceTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+@RunWith(SpringRunner.class)
+@SpringBootTest
+@WithMockUser
+public class VersionServiceTest {
```
carohauta, line 46: With this configuration a new Spring context might get created – and that adds up to the time it takes to run all tests. `VersionService` seems rather simple so maybe the test does not even need a Spring context? Or it could inherit the service test base class (`AbstractServiceTestContext`) for a common context?

carohauta (approved):
🦄

PR #886, opened 2021-02-22
XRDDEV-1576

jhyoty:


src/centralserver/admin-ui/src/App.vue
```diff
@@ -0,0 +1,34 @@
+<template>
```
carohauta, line 1: All `*.vue` and `*.ts` files are missing the license header

carohauta (commented):
`admin-ui` is still missing the license configuration file `license-check-and-add-config.json` and therefore the `license-add` and `license-check` tasks are not working. Also the `LICENSE` file is missing. The config file (`license-check-and-add-config.json`) will need some more ignore paths (e.g. gradle files) because the centralserver module structure differs from the proxy-ui-api.

Due to this some of the frontend files are missing license headers (specifically `*.ts` and `*.vue` files).

Otherwise I think this will work as a great skeleton project to begin with

src/centralserver/admin-ui/src/App.vue
jhyoty: I think that could be fixed in XRDDEV-1606 since it will anyway change the admin-ui -- probably even replacing many of the current placeholder files that currently miss the license.

We could also consider removing the separate license-add task from front-end projects and use the same gradle plugin as everything else. It should be capable of handling the ts/js/vue files as well with some additional configuration.

carohauta: Yeah that is probably the right ticket to figure out the license headers and how the gradle license plugin could be utilised for this.

carohauta (approved):
🦄

PR #888, opened 2021-02-22
XRDDEV-427

TJaakkola:
Add x-forwarded-for header value to proxy accesslog
https://jira.niis.org/browse/XRDDEV-427

### DIFF
--- a/src/proxy/src/main/java/ee/ria/xroad/proxy/clientproxy/ClientProxy.java
+++ b/src/proxy/src/main/java/ee/ria/xroad/proxy/clientproxy/ClientProxy.java
@@ -251,7 +251,9 @@ public class ClientProxy implements StartStop {
 
         final Slf4jRequestLogWriter writer = new Slf4jRequestLogWriter();
         writer.setLoggerName(getClass().getPackage().getName() + ".RequestLog");
-        final CustomRequestLog reqLog = new CustomRequestLog(writer, CustomRequestLog.EXTENDED_NCSA_FORMAT);
+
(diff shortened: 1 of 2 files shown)


src/proxy/src/main/java/ee/ria/xroad/proxy/clientproxy/ClientProxy.java
```diff
@@ -251,7 +251,9 @@ private void createHandlers() throws Exception {
 
         final Slf4jRequestLogWriter writer = new Slf4jRequestLogWriter();
         writer.setLoggerName(getClass().getPackage().getName() + ".RequestLog");
-        final CustomRequestLog reqLog = new CustomRequestLog(writer, CustomRequestLog.EXTENDED_NCSA_FORMAT);
+
+        final CustomRequestLog reqLog = new CustomRequestLog(writer, "%{X-Forwarded-For}i "
```
jhyoty, line 255: This is problematic due to "X-Forwarded-For" being an optional header that is not always present. Consider appending it instead, and (like other hearders) wrapping to quotes so that missing value is easier to detect. e.g.
```
       ... CustomRequestLog.EXTENDED_NCSA_FORMAT + " \"%{X-Forwarded-For}i\"";
```

jhyoty (changes requested):

src/proxy/src/main/java/ee/ria/xroad/proxy/clientproxy/ClientProxy.java
TJaakkola: Changed.

jhyoty (approved):

PR #908, opened 2021-03-12
Xrddev 1044 e2etest refresh wsdl

olliru:
Added refresh wsdl test


src/proxy-ui-api/frontend/tests/e2e/pages/ssMainPage.js
```diff
@@ -65,6 +65,21 @@ var navigateCommands = {
...
+    exec(
+      'ssh ' +
+        this.api.globals.testdata.split(':')[1].substring(2) + // remove protocol and port data
+        ' "cp ' +
+        this.api.globals.testfiles_path +
+        '/' +
+        newfile +
+        ' ' +
+        this.api.globals.testfiles_path +
+        '/testserviceX.wsdl"',
```
TJaakkola, line 79: We should use template literals instead of string concatenation. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

src/proxy-ui-api/frontend/tests/e2e/pages/ssMainPage.js
```diff
@@ -311,6 +326,26 @@ var clientServicesCommands = {
     );
     return this;
   },
+  verifyServiceSSL: function (service, status) {
+    this.api.waitForElementVisible(
+      '//tr[.//td[@data-test="service-link" and contains(text(), "' +
+        service +
+        '")]]//*[contains(@class, "mdi-lock") and contains(@style, "' +
+        status +
+        '")]',
```
TJaakkola, line 335: Template literals

src/proxy-ui-api/frontend/tests/e2e/pages/ssMainPage.js
```diff
@@ -311,6 +326,26 @@ var clientServicesCommands = {
...
+    );
+    return this;
+  },
+  verifyServiceURL: function (service, url) {
+    this.api.waitForElementVisible(
+      '//tr[.//td[@data-test="service-link" and contains(text(), "' +
+        service +
+        '")]]//*[contains(text(), "' +
+        url +
+        '")]',
```
TJaakkola, line 345: Template literals

src/proxy-ui-api/frontend/nightwatch.json
```diff
@@ -31,8 +31,10 @@
         "test_cert" : "cert.cer",
         "export_cert" : "certs.tar.gz",
         "e2etest_testdata" : "/../testdata",
+        "testfiles_path" : "/var/lib/tomcat7/webapps/ROOT/test_data",
```
TJaakkola, line 34: Using this static path makes tests rely on this certain path. I think a better solution for testing wsdl-refresh would be to provide a http interface for switching the served wsdl description. Current solution may work but it feels overly constrictive if we think about setting up ephemeral testing environments. 

However I think we have some pressure to get these tests running so we could do the refactoring in later versions.

TJaakkola (changes requested):

olliru:
Added in another PR (X-Road v7)

PR #909, opened 2021-03-15
XRDEV-1605 encrypting backups

amoog:
encrypting backups


iluwatar:
@amoog please raise another pull request against `develop`

PR #918, opened 2021-03-23
XRDDEV-541

TJaakkola:
https://jira.niis.org/browse/XRDDEV-541

- Old (ocsp responder has been changed and old removed) ocsp-responses are removed from the diagnostics 
- Diagnostics shows 'Unable to verify OCSP response' when the certificate isn't matching the certificate of the given ocsp-responder url


iluwatar:
@TJaakkola please raise another pull request against `develop`

PR #925, opened 2021-03-29
XRDDEV-541

TJaakkola:
https://jira.niis.org/browse/XRDDEV-541

Old (ocsp responder has been changed and old removed) ocsp-responses are removed from the diagnostics
Diagnostics shows 'Unable to verify OCSP response' when the certificate isn't matching the certificate of the given ocsp-responder url


src/signer/src/main/java/ee/ria/xroad/signer/certmanager/OcspClientWorker.java
```diff
@@ -454,37 +451,39 @@ boolean isCachedOcspResponse(String certHash) throws Exception {
...
-            log.error("Error getting certificate chain for certificate " + cert.getSubjectX500Principal(), e);
+            log.error("Error getting certificate chain for certificate {}", cert.getSubjectX500Principal(), e);
         }
 
         return emptyList();
     }
 
     private void initializeDiagnostics() {
-        for (X509Certificate caCertificate : GlobalConf.getAllCaCerts()) {
+
```
jansu76, line 461: It's sometimes a bit difficult (or slow) to study the code and understand, what happens and why. 

Maybe some inline comments, or short javadocs even for private methods, would help (e.g. `initializeDiagnostics` could have a short description that outlines it's purpose)

jansu76 (approved):
LGTM. 

Maybe some additional comments here and there would be helpful.

PR #926, opened 2021-03-29
XRDDEV-1624

TJaakkola:
https://jira.niis.org/browse/XRDDEV-1624


src/shared-ui/src/components/XrdTable.vue
```diff
@@ -0,0 +1,100 @@
...
+   all copies or substantial portions of the Software.
+
+   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+   AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+   THE SOFTWARE.
+ -->
```
carohauta, line 51:
```suggestion
```

src/centralserver/admin-ui/src/components/layout/MainContentArea.vue
```diff
@@ -0,0 +1,51 @@
...
+<style scoped lang="scss">
+.content {
+  margin-top: 40px;
+  padding-left: 10px;
+  padding-right: 10px;
+  width: 100%;
+  max-width: 1600px;
+
+  padding-left: 92px;
+  @media only screen and (max-width: 920px) {
```
carohauta, line 47: Should the breakpoints (e.g. `920px`) be separated to a theme file of sorts since they are used in multiple files. Also the colors and maybe even space widths could be gathered into one theme that would be easy to maintain.

Just an idea. What do you think, is it too big of a change? Because we don't use a theme in Security Server either

carohauta (commented):
Some nice low level, themable CSS toolkits to consider to get rid of redundant style declarations:
[BassCss](https://basscss.com/)
[Tailwind CSS](https://tailwindcss.com/)
[Tachyons](https://tachyons.io/)

src/centralserver/admin-ui/src/components/layout/MainContentArea.vue
TJaakkola: **X-Road global theme**: 
I think it would be a good idea to have common themefiles for centralserver adminview and proxy-api-ui but I'm afraid doing that refactoring would take quite some time. This should have been though before/during proxy-api-ui implementation, now it's laborous. I think we should create a separate ticket to see whether that refactoring would be doable in reasonable time. 

We already have a couple of files containing global styles in centralserver admin view. We could try to approach 'global theme' by splitting those files to a somekind of theme. But that could be done between mock-view tickets to avoid confusion. 

**CSS-toolkits**: 
Vuetify also contains some helper classes, grid-system etc. I think using those helper kits would introduce some duplication there. Also that kind of tools are not used in proxy-api-ui so I'd prefer using the same approach in centralserver adminui.

**Media styles to global styles**: 
I guess moving styles in this component to global stylefiles would make sense in many cases but I designed this maincontentarea -component so that it shouldn't be necessary to move media styles global (of course this could be refactored if this component is too restrictive). If this component can be applied to all the pages where 1600px max-width view is used then I don't see why these styles be made global. 

MainContentArea -component could be moved to shared-ui also.

carohauta: Yes, Vuetify already seems to have quite a vast selection of utility classes. Maybe they should just be utilised more instead of creating duplicates. But that is not related to this task what so ever.

Also about the themes, maybe the Vuetify theme could be customised so there would be no need for creating one from scratch https://vuetifyjs.com/en/features/theme/#customizing
But that probably also requires some effort.

carohauta (approved):
🦄

PR #928, opened 2021-03-30
XRDDEV-1604 Java 11 Runtime

carohauta:
Security Server now depends on jre 11. Central Server still on jre 8

JIRA: https://jira.niis.org/browse/XRDDEV-1604


src/packages/src/xroad/ubuntu/generic/control
```diff
@@ -30,15 +30,15 @@ Description: X-Road central server
...
+Depends: ${misc:Depends}, rlwrap, ca-certificates-java, crudini, adduser, net-tools, iproute2, sudo, openssl
 Replaces: xroad-proxy (<< 6.8.9), xroad-common (<= 6.17.0-0.20171018082348git2ecf68ed)
 Breaks: xroad-common (<= 6.17.0-0.20171018082348git2ecf68ed)
 Description: X-Road base components
  X-Road base components and utilities
 
 Package: xroad-confclient
 Architecture: amd64
-Depends: ${misc:Depends}, xroad-base (=${binary:Version})
+Depends: ${misc:Depends}, openjdk-11-jre-headless | java11-runtime-headless | java11-runtime, xroad-base (=${binary:Version})
```
jhyoty, line 41: forces java 11 also on central server, which should still use java 8

src/packages/src/xroad/ubuntu/generic/control
```diff
@@ -55,7 +55,7 @@ Description: X-Road nginx component
 
 Package: xroad-signer
 Architecture: amd64
-Depends: ${misc:Depends}, xroad-base (=${binary:Version})
+Depends: ${misc:Depends}, openjdk-11-jre-headless | java11-runtime-headless | java11-runtime, xroad-base (=${binary:Version})
```
jhyoty, line 58: forces installing Java 11 also on central server, which should still use java 8

jhyoty (changes requested):

src/packages/src/xroad/ubuntu/generic/control
```diff
@@ -30,7 +30,7 @@ Description: X-Road central server
 Package: xroad-base
 Architecture: amd64
 Pre-depends: dpkg (>= 1.15.7.2)
-Depends: ${misc:Depends}, openjdk-8-jre-headless | java8-runtime-headless | java8-runtime, rlwrap, ca-certificates-java, crudini, adduser, net-tools, iproute2, sudo, openssl
```
jhyoty, line 33: I think it would simplify things if you keep this declaration (`openjdk-8-jre-headless | java8-runtime-headless | java8-runtime` on xroad-base, but add the `openjdk-11-jre-headless | java11-runtime-headless | java11-runtime` just to xroad-proxy. That would force java 11 when installing the security server,  but use java 8 otherwise (java11 provides  also the "java8-runtime-headless" virtual package, so the dependency is satisfied).

carohauta: Alright I think I got it now. Only xroad-proxy depends on jre11 now so it gets installed only when security server is installed.

PR #930, opened 2021-03-31
XRDDEV-1581-follow-up

iluwatar:


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/auth/securityconfigurer/ApiWebSecurityConfigurerAdapter.java
```diff
@@ -64,6 +64,10 @@ protected void configure(HttpSecurity http) throws Exception {
         filter.setExceptionIfHeaderMissing(false); // exception at this point
         // would cause http 500, we want http 401
         http
+            .authorizeRequests()
+                .antMatchers("/api/**/openapi")
+                .permitAll()
+                .and()
```
jansu76, line 70: This now has two `authorizeRequests()` blocks/calls. Are you sure this is legal way to use of the API? I am not sure how the two blocks interact with the other parts of the configuration. It seems to work in quick tests, but the configuration seems a bit unorthodox, and I am not sure if it is robust when we or Spring make some modifications?

There's also some parts that are not relevant for the always-public endpoint, like `ApiKeyAuthenticationManager` and `RequestHeaderAuthenticationFilter`, and possibly csr and session related configuration if this API is not meant to be used via frontend.

I think a structure with less question marks could be to introduce a new `WebSecurityConfigurerAdapter` reserved only for unauthenticated API calls. Probably put it in front of `ApiWebSecurityConfigurerAdapter` (index 2, move others accordingly)

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/auth/securityconfigurer/ApiWebSecurityConfigurerAdapter.java
```diff
@@ -64,6 +64,10 @@ protected void configure(HttpSecurity http) throws Exception {
         filter.setExceptionIfHeaderMissing(false); // exception at this point
         // would cause http 500, we want http 401
         http
+            .authorizeRequests()
+                .antMatchers("/api/**/openapi")
```
jansu76, line 68: Probably better to use exact match `/api/v1/openapi` to be completely sure. Now it matches also `/api/v1/some-other-path/openapi`. Framework seems to take care of simple attempts to mess with this `/api/v1/foobar?trick=/openapi` but who knows, maybe there could be some kind of exploit - and I guess there's no real need for the use of the wildcard?

jansu76 (changes requested):

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/auth/securityconfigurer/ApiWebSecurityConfigurerAdapter.java
iluwatar: Unauthenticated endpoints are now configured in the separate method `ApiWebSecurityConfigurerAdapter#configure(WebSecurity web)`

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/auth/securityconfigurer/ApiWebSecurityConfigurerAdapter.java
iluwatar: Using exact match now

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/OpenapiApiControllerTest.java
```diff
@@ -0,0 +1,55 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.niis.xroad.restapi.openapi.OpenapiApiController.OPENAPI_DEFINITION_FILENAME;
+
+public class OpenapiApiControllerTest extends AbstractApiControllerTestContext {
+
+    @Autowired
+    OpenapiApiController openapiApiController;
+
+    @Test
+    @WithMockUser(authorities = { "DOWNLOAD_OPENAPI" })
```
jansu76, line 48: This line should probably be removed, since permission DOWNLOAD_OPENAPI was removed? 

MockUser is not needed, and probably better off, since having one in the test can even be a tad misleading.

jansu76 (changes requested):

src/proxy-ui-api/src/test/java/org/niis/xroad/restapi/openapi/OpenapiApiControllerTest.java
iluwatar: Removed the extra line in commit https://github.com/nordic-institute/X-Road/pull/930/commits/66609ecb6c6e8886234224793ebd454300dc5ee3

iluwatar:
The build failure due to the front-end audit check is not originating from this change, so I'll merge this ticket

PR #931, opened 2021-03-31
XRDDEV-1612 remove executable code from backups

amoog:
- initialisation of environment variables from .properties file
- do not backup/restore executable .conf files
- move SS database backups to "custom" format

### DIFF
--- a/src/packages/src/xroad/common/base/etc/xroad/services/global.conf
+++ b/src/packages/src/xroad/common/base/etc/xroad/services/global.conf
@@ -16,9 +16,6 @@ fi
 export LD_LIBRARY_PATH="/usr/share/xroad/lib:$LD_LIBRARY_PATH"
 
 load_from_properties () {
-  ALLOWED_KEYS=("XROAD_PARAMS" "JETTY_PARAMS" "CONFCLIENT_PARAMS" "CONFPROXY_PARAMS" "MONITOR_PARAMS" "OPMON_PARAMS" \
-    "PROXY_UI_API_PARAMS" "XROAD_LOG_LEVEL")
-
   if [ -f /etc/xroad/services/local.properties ]; then
     while IFS='=' read -r key value
     do
--- a/src/packages/src/xroad/common/base/etc/xroad/services/global.conf
+++ b/src/packages/src/xroad/common/base/etc/xroad/services/global.conf
@@ -1,5 +1,8 @@
+#!/bin/bash
+
 # Please do not change this file! It will be overwritten by updates.
-# System specific changes should be added to /etc/xroad/services/local.conf
+# System specific changes should be added to /etc/xroad/services/local.properties
+
 ADDON_PATH="/usr/share/xroad/jlib/addon"
 
 umask 0027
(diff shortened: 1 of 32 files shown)


src/packages/src/xroad/common/proxy/usr/share/xroad/scripts/restore_db.sh
```diff
@@ -41,13 +41,18 @@ psql_dbuser() {
...
-BEGIN;
 DROP SCHEMA IF EXISTS "$db_schema" CASCADE;
+CREATE SCHEMA "$db_schema";
 EOF
-  cat "$dump_file"
-  echo "COMMIT;"
-} | psql_adminuser || abort "Restoring database failed."
+} | psql_adminuser || abort "Dropping schema failed."
+
+pgrestore || abort "Restoring database failed."
```
jhyoty, line 55: The previous implementation did not change the database if restore failed, but now the database can be left in a broken state. One option would be to just create the schema if it does not exist (because pg_restore will not recreate it), and use the --clean option in restore (which runs in a transaction).

jhyoty (changes requested):
Changes to 
* ansible/hosts/lxd_hosts.txt
* src/proxy-ui-api/frontend/package-lock.json`
seem unrelated to the task.

src/packages/src/xroad/common/proxy/usr/share/xroad/scripts/restore_db.sh
amoog: --clean does not drop objects  that are not in the backup... but then again this is only an issue when backup is made from a different versions - a situation that must be handled separately. So yes, lets go wit IF NOT EXISTS and --clean

src/packages/src/xroad/common/base/etc/xroad/services/global.conf
```diff
@@ -11,4 +14,18 @@ if [ -f /var/run/xroad/xroad-akka-env.properties ]; then
 fi
 
 export LD_LIBRARY_PATH="/usr/share/xroad/lib:$LD_LIBRARY_PATH"
-XROAD_PARAMS=" -XX:+UseG1GC -Dfile.encoding=UTF-8 -Xshare:auto -Djdk.tls.ephemeralDHKeySize=2048 "
+
+load_from_properties () {
+  if [ -f /etc/xroad/services/local.properties ]; then
+    while IFS='=' read -r key value
+    do
+      if [ -n "$key" ] && [ -n "$value" ] && [[ $key != \#* ]]; then
```
jhyoty, line 22: As an additional precaution, it would be good to ignore "PATH"

src/packages/src/xroad/common/base/usr/share/xroad/scripts/_restore_xroad.sh
```diff
@@ -142,7 +142,7 @@ setup_tmp_restore_dir() {
 extract_to_tmp_restore_dir () {
```
jhyoty, line 142: for additional security, extract should ensure that restored files don't have the execute mode bit set (with the ability to change path, that would again create a RCE vulnerability).

src/packages/src/xroad/common/proxy/usr/share/xroad/scripts/restore_db.sh
```diff
@@ -41,13 +41,17 @@ psql_dbuser() {
...
+
 { cat <<EOF
-BEGIN;
-DROP SCHEMA IF EXISTS "$db_schema" CASCADE;
+CREATE SCHEMA IF NOT EXISTS "$db_schema";
 EOF
-  cat "$dump_file"
-  echo "COMMIT;"
-} | psql_adminuser || abort "Restoring database failed."
+} | psql_adminuser || abort "Dropping schema failed."
```
jhyoty, line 52: Error message is now misleading.

jhyoty (changes requested):

src/packages/src/xroad/common/base/etc/xroad/services/global.conf
amoog: As discussed on slack lets go one step further and fix the list of environment variables that can be changed as this is the safest option.

src/packages/src/xroad/common/base/usr/share/xroad/scripts/_restore_xroad.sh
```diff
@@ -160,6 +165,8 @@ extract_to_tmp_restore_dir () {
       cp /etc/xroad/db.properties ${RESTORE_DIR}/etc/xroad/db.properties
   fi
   chown -R xroad:xroad ${RESTORE_DIR}/*
+  # remove executable bit from all files (but not from directories)
+  find "$RESTORE_DIR" -type f -executable -execdir chmod a-x {} +
```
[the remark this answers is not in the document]
amoog: Indeed, especially setuid bit is problematic too. There is a light possibility that some admin wants to change permissions on some files but as I cant find any example right now lets go with fixed permissions.

jhyoty (approved):

PR #933, opened 2021-04-07
XRDDEV-1631 Mock views for central server members pages

TJaakkola:
https://jira.niis.org/browse/XRDDEV-1631


src/centralserver/admin-ui/src/views/Members/Member/Subsystems/MemberSubsystems.vue
```diff
@@ -0,0 +1,153 @@
...
+  text-align: right;
+
+  .xrd-clickable {
+    margin-left: 10px;
+  }
+}
+
+.sorting {
+  margin-bottom: 5px;
+}
```
carohauta, line 148: I don't necessarily find it very convenient that a css class is called `sorting` and it only adds 5px bottom margin. Vuetify's `mb-1` utility class already adds 4px bottom margin so that could possibly be used instead.

At some point it would probably be good to come up with some sort of guide about the styles. For example there are a lot of duplicate css classes (with minor differencies) in the security server UI also: search for example `.clickable-link` from the codebase.

TJaakkola: Agreed. Don't know why I've left it that way as I've used those vuetifys helper css classes.

TJaakkola: Fixed. I guess I used 'sorting' as it gives a better hint on what is that class used for (positioning sorting arrow) but I removed it as unnecessary.

src/centralserver/admin-ui/src/components/layout/PageNavigation.vue
```diff
@@ -0,0 +1,67 @@
...
+
+<style scoped lang="scss">
+@import '../../assets/colors';
+
+.page-navigation-row {
+  // Unset default purple background
+  .v-tab::before:hover {
+    opacity: 0 !important;
+  }
+}
```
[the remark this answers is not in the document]
TJaakkola: Can't remember what I have been fixing with that declarations but now it's removed.

carohauta (approved):
🦄

PR #935, opened 2021-04-07
Fix archive transport script curl url specifying

qrlt:
Curl URL specifying is currently not by the book meaning that using URL with https is not possible.

`--url <url>` is the correct way

Thanks @Tuudik !


jhyoty:
Could you elaborate on the issue you are experiencing? For me, https urls seem to work without "--url".

jhyoty:
There were some potential problems in the script which are fixed in #952 . Closing for now, please provide additional details if you still have issues.

PR #945, opened 2021-04-21
XRDDEV-1549

iluwatar:


src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/validator/AbstractIdentifierValidator.java
```diff
@@ -1,75 +0,0 @@
...
-import org.springframework.validation.Validator;
-
-import java.util.Collection;
-import java.util.EnumSet;
-
-/**
- * Base class for validators that validate objects with fields that are used to create encoded identifiers
- */
-@Slf4j
-abstract class AbstractIdentifierValidator implements Validator {
```
carohauta, line 44: Removing this hides some validation error data. 

For example when adding a new subsystem `CS:ORG:1234:Sub%Sys\\t;em` would return the following validation error:
```
{
    "status": 400,
    "error": {
        "code": "validation_failure",
        "validation_errors": {
            "clientAdd.client.subsystemCode": [
                "IdentifierChars"
            ]
        }
    }
}
```

where the old validation would return a more specific validation error:
```
{
    "status": 400,
    "error": {
        "code": "validation_failure",
        "validation_errors": {
            "clientAdd.client.subsystemCode": [
                "NoPercents",
                "NoSemicolons",
                "NoBackslashes"
            ]
        }
    }
}
```

src/proxy-ui-api/src/main/resources/META-INF/custom-constraints.xml
```diff
@@ -0,0 +1,111 @@
...
+            </constraint>
+        </field>
+    </bean>
+    <bean class="LocalGroupDescription" ignore-annotations="false">
+        <field name="description">
+            <constraint annotation="org.niis.xroad.restapi.openapi.validator.NoControlChars">
+            </constraint>
+        </field>
+    </bean>
+    <bean class="ServiceDescriptionAdd" ignore-annotations="false">
```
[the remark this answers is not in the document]
iluwatar: Well spotted. I added a validator for that one as well in commit https://github.com/nordic-institute/X-Road/pull/945/commits/c09fcee2b244484b153a735d521ff9c3f3144bc9

src/proxy-ui-api/src/main/java/org/niis/xroad/restapi/openapi/validator/AbstractIdentifierValidator.java
iluwatar: This is true, it was a conscious design decision. If we wanted to keep that level of detail in the outcoming validation errors, we would need to implement separate validators for each (`NoPercentsValidator`, `NoSemicolonsValidator` etc.) instead of relying on a single validator `IdentifierCharsValidator` as we do now. I think from the end-users perspective this wouldn't add much since we can formulate the error messages in a generic manner e.g. "The field contains unallowed characters [%;\...]"

carohauta: It is probably a good compromise between codebase maintainability and user experience

carohauta (approved):
🦄

PR #953, opened 2021-04-30
XRDDEV-1640 Add openjdk11 or generic java8 dependencies for xroad-jetty9

carohauta:
JIRA: https://jira.niis.org/browse/XRDDEV-1640

### DIFF
--- a/src/packages/src/xroad-jetty9/ubuntu/generic/control
+++ b/src/packages/src/xroad-jetty9/ubuntu/generic/control
@@ -7,6 +7,6 @@ Standards-Version: 3.8.3
 
 Package: xroad-jetty9
 Architecture: all
-Depends: ${misc:Depends}, openjdk-8-jre-headless | java8-runtime-headless | java8-runtime, xroad-base (=${binary:Version})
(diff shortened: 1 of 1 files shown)


src/packages/src/xroad-jetty9/ubuntu/generic/control
```diff
@@ -7,6 +7,6 @@ Standards-Version: 3.8.3
 
 Package: xroad-jetty9
 Architecture: all
-Depends: ${misc:Depends}, openjdk-8-jre-headless | java8-runtime-headless | java8-runtime, xroad-base (=${binary:Version})
+Depends: ${misc:Depends}, xroad-base (=${binary:Version}), openjdk-11-jre-headless | java8-runtime-headless | java8-runtime
```
jhyoty, line 10: In 7.0 jetty is only used by xroad-center, therefore it should depend on java 8, not 11 (and in 7.x, there is no stand-alone jetty any more) --> there seems to be nothing to fix.

carohauta: Yes, it seems that there are no reasons for this change

PR #955, opened 2021-05-03
XRDDEV-1653 Update error view

Riippi:
https://jira.niis.org/browse/XRDDEV-1653

Updated 404 error page


src/proxy-ui-api/frontend/src/views/AppError.vue
```diff
@@ -23,12 +23,115 @@
...
  -->
+
 <template>
-  <v-layout align-center justify-center column>oops, page not found</v-layout>
+  <div class="xrd-view-common">
+    <TabsBase />
+    <AlertsContainer />
+    <v-container>
+      <div class="xrd-view-title pt-6">{{ $t('404') }}</div>
+      <v-layout align-center justify-center column fill-height elevation-0>
```
carohauta, line 33: It's a pretty simple view. I don't think this `v-layout` is needed here(?)

src/proxy-ui-api/frontend/src/views/AppError.vue
```diff
@@ -23,12 +23,115 @@
...
+        <v-card flat class="xrd-card custom-card">
+          <v-card-title> </v-card-title>
+
+          <v-card-text>
+            <div class="content-wrap">
+              <v-img
```
carohauta, line 39: The image does not scale down for smaller screens.

carohauta (commented):
I noticed that the 404 page is not shown when a user with insufficient privileges tries to go to some forbidden route.

For example, if a user with `security-officer` role tries to go to `https://ss1/#/settings/backup`, s/he is just redirected to a default route which is `https://ss1/#/clients`. Would it make sense to show the 404 page instead?

src/proxy-ui-api/frontend/src/views/AppError.vue
Riippi: Yes, seems it's not needed so I removed it.

src/proxy-ui-api/frontend/src/views/AppError.vue
Riippi: Fixed

Riippi:
> I noticed that the 404 page is not shown when a user with insufficient privileges tries to go to some forbidden route.
> 
> For example, if a user with `security-officer` role tries to go to `https://ss1/#/settings/backup`, s/he is just redirected to a default route which is `https://ss1/#/clients`. Would it make sense to show the 404 page instead?

This was discussed in slack and the best solution would be to add a another view for "no permission" case. That would be separate task.

PR #956, opened 2021-05-04
Fixed a bug that doesn't create recovery.conf on RHEL7/8 standby nodes

gudvardur:
According to https://github.com/nordic-institute/X-Road/blob/develop/doc/Manuals/LoadBalancing/ig-xlb_x-road_external_load_balancer_installation_guide.md#44-configuring-the-replica-instance-for-replication

You should create on standby node  the file `recovery.conf` on RHEL7/8 and Ubuntu 18.04... But the ss_cluster playbook doesn't do that... Hence the postgresql replication fails on RHEL7/8.

### DIFF
--- a/ansible/ss_cluster/roles/slave/tasks/serverconf_slave_db.yml
+++ b/ansible/ss_cluster/roles/slave/tasks/serverconf_slave_db.yml
@@ -21,7 +21,9 @@
       dest: "{{ postgresql_data_dir }}/recovery.conf"
       owner: postgres
       group: postgres
-  when: ansible_distribution_release == "bionic"
+  when: >
+    ansible_distribution_release == "bionic" or 
+    (ansible_distribution == "RedHat" and (ansible_distribution_major_version == "7" or ansible_distribution_major_version == "8"))
 
 # recovery.conf is no longer supported in PostgreSQL12
 # see https://www.2ndquadrant.com/en/blog/replication-configuration-changes-in-postgresql-12/


ansible/ss_cluster/roles/slave/tasks/serverconf_slave_db.yml
```diff
@@ -21,7 +21,9 @@
       dest: "{{ postgresql_data_dir }}/recovery.conf"
       owner: postgres
       group: postgres
-  when: ansible_distribution_release == "bionic"
+  when: >
```
jansu76, line 24: Hi, thank you for creating this PR!

We discussed this in the team and thought that it would be better to compare postgresql version instead of Linux distribution version. This would be a bit more clear to the maintainer of this playbook, and more robust e.g. in case we e.g. run the playbook in new distribution versions in future.

Alternative implementation is at https://github.com/nordic-institute/X-Road/pull/969/files - have a look at it, you can backport the changes to your PR, or we could merge 969 directly. 

I tested PR 969 on Ubuntu 18/20 and Centos 7/8. On a related not, running on centos brought into light some bugs, they were fixed in https://github.com/nordic-institute/X-Road/pull/968

jansu76 (changes requested):

ansible/ss_cluster/roles/slave/tasks/serverconf_slave_db.yml
jansu76: @gudvardur I will proceed to merge the alternative implementation https://github.com/nordic-institute/X-Road/pull/969/files now

jansu76:
Closing, alternative implementation https://github.com/nordic-institute/X-Road/pull/969/files merged instead

PR #970, opened 2021-05-20
Xrddev 1573 e2etest diagnostics javaversion

turkkaholmqvist:
Continuing work that Olli started, added verification for JavaVersion field. 

Asserts verify that value converts to `int` or `float` when converted to numeric format. This should prevent any other text values from passing.

### DIFF
--- a/src/proxy-ui-api/frontend/tests/e2e/specs/ss-diagnostics.js
+++ b/src/proxy-ui-api/frontend/tests/e2e/specs/ss-diagnostics.js
@@ -24,9 +24,11 @@
  * THE SOFTWARE.
  */
 
+var assert = require('assert');
+
 module.exports = {
   tags: ['ss', 'diagnostics'],
-  'Dagnostics java version': (browser) => {
+  'Field JavaVersion contains numeric data': (browser) => {
     const frontPage = browser.page.ssFrontPage();
     const mainPage = browser.page.ssMainPage();
     const diagnosticsTab = browser.page.tabs.diagnosticsTab();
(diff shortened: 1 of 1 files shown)


src/proxy-ui-api/frontend/tests/e2e/specs/ss-diagnostics.js
```diff
@@ -0,0 +1,55 @@
...
+    const frontPage = browser.page.ssFrontPage();
+    const mainPage = browser.page.ssMainPage();
+    const diagnosticsTab = browser.page.tabs.diagnosticsTab();
+
+    // Open SUT and check that page is loaded
+    frontPage.navigate();
+    browser.waitForElementVisible('//*[@id="app"]');
+
+    // Enter valid credentials
+    frontPage.signinDefaultUser();
```
TJaakkola, line 41: After merging develop to this branch you can replace these three lines (and frontPage-variable declaration) by

`browser.LoginCommand();`

TJaakkola (changes requested):

src/proxy-ui-api/frontend/tests/e2e/specs/ss-diagnostics.js
TJaakkola: also feel free to rename that 'loginCommand' if that feels awkward

src/proxy-ui-api/frontend/tests/e2e/specs/ss-diagnostics.js
```diff
@@ -0,0 +1,55 @@
...
+
+    // Navigate to target page
+    mainPage.openDiagnosticsTab();
+    browser
+      .waitForElementVisible(diagnosticsTab.elements.javaVersion)
+      .getText(diagnosticsTab.elements.javaVersion, function (javaVersion) {
+        assert(typeof javaVersion.value == 'string', 'is not a string');
```
TJaakkola, line 48: Using this notation here doesn't make a difference but it's a common bad practise to use two equality signs instead of three. Generally you should always do equality comparison with '===' instead of '=='

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Strict_equality

PR #980, opened 2021-06-01
XRDDEV-1648

jhyoty:


src/common-messagelog/src/main/java/ee/ria/xroad/common/messagelog/archive/Grouping.java
```diff
@@ -0,0 +1,100 @@
...
+ * THE SOFTWARE.
+ */
+package ee.ria.xroad.common.messagelog.archive;
+
+import ee.ria.xroad.common.messagelog.MessageRecord;
+
+import java.util.Objects;
+
+interface Grouping {
+    boolean includes(MessageRecord record);
```
Dos4dev, line 34: Usually, a method that returns a boolean has a prefix is. However, I guess isIncludes would be more convenient. https://stackoverflow.com/a/10009912/9183844

jhyoty: That convention is used for getter methods returning boolean state. Here, `includes` is not a "getter"; it is conceptually like e.g. java.util.Set::contains.

iluwatar (approved):

PR #986, opened 2021-06-08
XRDDEV-1694 Storybook fix + Vue linter level update

Riippi:
https://jira.niis.org/browse/XRDDEV-1694
- Fix buttons in storybook examples
- Update npm dependencies
- Vue is is locked to 2.6.12 for now as the next version causes some typescript warning that has to be investigated.
- Vue linter rules level is now recommended instead of essential. This catches a few issues that were unnoticed. Also makes the code a bit more readable.


carohauta (approved):
🦄

PR #990, opened 2021-06-14
Xrddev 1682 prune repetition from tests

turkkaholmqvist:
I changed scheme of how page-specific functions are being imported, now they are imported once in `before`-block of test suite, removing repetition from actual tests.

Also I also removed utils -folder from source directory, it shouldn't be there since this src is used to distinguish different tests from utility-functions.


iluwatar (changes requested):
The changes look good, but the pull request build is failing due to a front-end audit check failure. I think this can be resolved by merging the `develop` back to this branch. After the build passes, I can do the merging.

carohauta (approved):
Audit issues have been resolved 🦄

PR #992, opened 2021-06-17
XRDDEV-1683, XRDDEV-1684, XRDDEV-1685 Global configuration tabs

Riippi:
https://jira.niis.org/browse/XRDDEV-1683
https://jira.niis.org/browse/XRDDEV-1684
https://jira.niis.org/browse/XRDDEV-1685

Internal configuration, external configuration and trusted anchors tabs with mocked data.


src/centralserver/admin-ui/src/views/GlobalConfiguration/shared/Configuration.vue
```diff
@@ -0,0 +1,251 @@
...
+   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+   IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+   FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+   AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+   THE SOFTWARE.
+ -->
+<template>
+  <div class="about">
```
carohauta, line 27: `about` class seems redundant – at least could not find it and seems oddly named. It is also in TrustedAnchors.vue, MockSubview.vue, MockView1.vue and MockView2.vue

 But then again this is only the mock so I guess it does not really matter at this time

carohauta (commented):
Navigating to the Global configuration tab does not reveal the contents of the first tab: Internal Configuration. Probably missing some default route or something?

src/centralserver/admin-ui/src/views/GlobalConfiguration/shared/Configuration.vue
Riippi: Fixed the routing. Reason is that tab component was using a name of the "parent route". The parent route shouldn't even have a name.
Removed the about class because it was a remnant from even earlier mock.

carohauta (approved):
🦄

PR #993, opened 2021-06-17
XRDDEV-1651 cs login backend

carohauta:
Adds the `UserApiController` to the centralserver admin-service backend. Would have been nice to move it to the `common-rest-api` module instead of duplicating it, but there were some problems with how the Spring OpenAPI Generator works in comparison to the openapi-typescript-codegen that it used in the frontend.

To summarize the problems:
The different frontends for Central Server and Security Server have no visibility to the api-definition in `common-rest-api`
-> therefore the `User` TypeScript interface will not be generated for either frontend
-> the client side of the `/user` endpoint will not be generated for either frontend

See the JIRA issue: https://jira.niis.org/browse/XRDDEV-1651


jhyoty:
Seems that #997 makes this PR outdated (or obsolete).

carohauta:
This PR is not needed anymore as @jhyoty mentioned. Closing

PR #997, opened 2021-06-30
XRDDEV-1715 Common api codegen

carohauta:
Adds common code generation for the frontends

JIRA: https://jira.niis.org/browse/XRDDEV-1715


src/centralserver/admin-ui/package.json
```diff
@@ -5,14 +5,15 @@
   "scripts": {
     "serve": "rm -rf src/openapi-types && npm run generate-types && vue-cli-service serve",
     "build": "rm -rf src/openapi-types && npm run generate-types && vue-cli-service build",
-    "generate-types": "openapi --input ../admin-service/src/main/resources/openapi-definition.yaml --output ./src/openapi-types --exportCore true --client fetch",
+    "generate-types": "npm run prepare-api-definition && openapi --input ./openapi-definition-frontend.yaml --output ./src/openapi-types/cs-types --exportCore true --client fetch && openapi --input ../../common-rest-api/src/main/resources/common-openapi-definition.yaml --output ./src/openapi-types/common-types --exportCore true --client fetch",
```
jansu76, line 8: This branch https://github.com/nordic-institute/X-Road/blob/XRDDEV-1535-openapi-for-cs/src/proxy-ui-api/frontend/package.json#L15 has a working example of openapi definition bundling with `prepare-api-definition` (from `apidevtools/swagger-cli`), would it maybe be an easier route to take?

src/common-rest-api/src/main/resources/common-openapi-definition.yaml
```diff
@@ -17,20 +17,32 @@ servers:
 tags:
   - name: dummy
```
jansu76, line 18: tags here, and in proxy-ui-api/openapi-definition, have some issues. 

common-openapi-definition endpoint uses tag "user", which is not introduced
proxy-ui-api/openapi-definition introduces tag "user", which is not used in any endpoint

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
```diff
@@ -110,12 +110,12 @@ paths:
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/ErrorInfo'
+                $ref: '../../../../common-rest-api/src/main/resources/common-openapi-definition.yaml#/components/schemas/ErrorInfo'
```
jansu76, line 113: It would probably be better to introduce local schema object `ErrorInfo` which refs to common-rest-api/ErrorInfo. This way the long relative path to common-openapi-definition does not need to be repeated in so many places, and we have easier time if folder structure or naming needs to be updated (and it's a bit more clear for the reader as well I guess)

Like here for `User`: https://github.com/nordic-institute/X-Road/blob/XRDDEV-1715-importMappings/src/proxy-ui-api/src/main/resources/openapi-definition.yaml#L6876-L6877

It seemed like the end result (= generated source code) is not affected by this(?)

jansu76 (changes requested):
So it was discovered that openapi generator has option `importMappings`, which can be used to get rid of the duplicate objects problem in Java: https://confluence.niis.org/display/XRDDEV/Additional+notes+about+code+generation+and+common-rest-api#Additionalnotesaboutcodegenerationandcommonrestapi-Update:finalform:importMappings

It should probably be taken into use here and used to prevent common model objects `ErrorInfo`, `CodeWithDetails` and `User` from being generated in `proxy-ui-api` and `centralserver/admin-service`

src/centralserver/admin-ui/package.json
carohauta: I tried to avoid adding new npm packages here. Don't know if it's easier with the `swagger-cli` package but in the long run it might come in handy 👍 -> changed to suggested

src/common-rest-api/src/main/resources/common-openapi-definition.yaml
carohauta: Fixed the tags

jhyoty (approved):

PR #999, opened 2021-07-02
XRDDEV-1721: Exclude /etc/xroad/gpghome from cluster rsync

jhyoty:
Fixes cluster rsync fails due to filesystem permissions.


carohauta (approved):
🦄

PR #1001, opened 2021-07-06
XRDDEV-1697 OpenAPI version exception

carohauta:
Adds an exception when adding a new servicedescription (OpenAPI) with a version different from `3.0.x`

JIRA: https://jira.niis.org/browse/XRDDEV-1697


src/proxy-ui-api/src/main/java/org/niis/xroad/securityserver/restapi/wsdl/UnsupportedOpenApiVersionException.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+package org.niis.xroad.securityserver.restapi.wsdl;
+
+import org.niis.xroad.restapi.exceptions.ErrorDeviation;
+import org.niis.xroad.restapi.service.ServiceException;
+
+/**
+ * Thrown if the provided OpenAPI document version differs from 3.0.x
+ */
+public class UnsupportedOpenApiVersionException extends ServiceException {
```
jhyoty, line 34: Perhaps this could be a subclass of the OpenApiParser.ParserException, that would avoid method signature and try-catch changes where handling that exception specially is not required; Or, maybe version mismatch could have been reported using the ParserException with specific ErrorDeviation without even introducing a new exception.

carohauta: True, but I think that the implementation is in line with the rest of the `proxy-ui-api` codebase. We have made an intentional decision to be quite verbose with the thrown exceptions. Of course, as a known downside, this might lead to some extra try-catch handling and also results in some of the method signatures being quite bloated with exceptions.

At the moment the parser is used only by `addServiceDescription`, `updateServiceDescription` and `refreshServiceDescription` and I think it's good to see the explicit exception in their signatures. Not sure how I feel about "hiding" the `UnsupportedOpenApiVersionException` under `ParsingException`.

jhyoty: One could argue that the UnsupportedVersionException is a (subclass of) ParsingException, and at the controller level it is handled exactly the same as the parsing error -- just wrapped to a more general exception. Therefore polluting the signatures and forcing changes to higher level code seems unnecessary here. Anyway, I accept that this is in line with other code, and maybe we revisit the general issue some other time.

jhyoty (approved):

PR #1015, opened 2021-08-06
XRDDEV-1676

iluwatar:


src/proxy-ui-api/src/main/java/org/niis/xroad/securityserver/restapi/service/InternalTlsCertificateService.java
```diff
@@ -198,25 +205,23 @@ public X509Certificate importInternalTlsCertificate(byte[] certificateBytes) thr
...
         X509Certificate internalCert = getInternalTlsCertificate();
         PublicKey internalPublicKey = internalCert.getPublicKey();
-        PublicKey newCertPublicKey = newCert.getPublicKey();
-        if (!internalPublicKey.equals(newCertPublicKey)) {
+
+        boolean found = newCertChain.stream().anyMatch(c -> c.getPublicKey().equals(internalPublicKey));
+        if (!found) {
             throw new KeyNotFoundException("The imported cert does not match the internal TLS key");
-        } else if (internalCert.equals(newCert)) {
-            throw new CertificateAlreadyExistsException("The imported cert already exists");
```
carohauta, line 219: Everything looks good but here you have removed the exception for when trying to import the currently existing cert again. This was in the acceptance criteria of an earlier ticket https://jira.niis.org/browse/XRDDEV-1659 so I'm just wondering if was removed on purpose

iluwatar: Good point. The reason I removed it was because the check didn't work with certificate chains. I didn't see it as that important. But since you pointed out that showing an error when an existing cert is uploaded again was an acceptance criterion in a previous ticket, I guess I need to restore it and refactor it to make it work.

iluwatar: Fixed now @carohauta

carohauta (approved):
🦄

PR #1022, opened 2021-08-18
XRDDEV-1734 Central server svg icons

Riippi:
https://jira.niis.org/browse/XRDDEV-1734
- Adds new svg icons
- Changed most of the old icons to the new format
- Fix header texts in the trust services view
- Fix some issues with style template tags


src/shared-ui/src/components/icons/IconBase.vue
```diff
@@ -0,0 +1,73 @@
...
+    <title :id="iconName" lang="en">{{ iconName }} icon</title>
+    <slot :fillColor="color" />
+  </svg>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+
+export default Vue.extend({
+  name: 'XrdIconBase',
```
carohauta, line 46: I get the idea of having the `XrdIconBase` component but IMO the usage is not too convenient at the moment:
```
<xrd-icon-base color="red">
  <XrdIconErrorNotification />
</xrd-icon-base>
```

What would it take to get it syntactically improved to something like this?
```
<XrdIconErrorNotification color="red" />
```
Perhaps the base component could be baked into the icon components

Riippi: I could be wrong, but it's not possible to inherit a part of the template. So I don't know how to share the wrapping part. It's possible to inherit methods etc. via mixins, but that would not help in this case. Also if there is some sharing a solution it would add more code and "moving parts" into every icon file. Currently they are super simple.

This solution is done following the example in Vue "official" cookbook. https://vuejs.org/v2/cookbook/editable-svg-icons.html

carohauta (approved):
🦄

PR #1036, opened 2021-09-06
XRDDEV-1670 conf param prefix

carohauta:
JIRA: https://jira.niis.org/browse/XRDDEV-1670

Check [proxy.conf](https://github.com/nordic-institute/X-Road/blob/124ba41f58f9e7a3f94a4db66df6bcd2e05da074/src/packages/src/xroad/common/proxy/etc/xroad/services/proxy.conf) for commented code


jhyoty:
It seems we could simplify this by lifting the logic to a helper function in global.conf and also simplifying the logic a bit:
https://github.com/nordic-institute/X-Road/compare/XRDDEV-1670-conf-param-prefix...XRDDEV-1670-refactor

And to make things a bit simpler, it is not necessary the separately consider local.properties and local.conf. Defaults + "local.properties" should be the input  for potential local.conf modifications.

carohauta:
I tested and applied @jhyoty 's refactoring – makes the code way simpler

carohauta:
Successful build: https://jenkins.niis.org/job/pr-build-pipeline/2303/consoleFull

jhyoty (approved):
LGTM. Seems I forgot to update signed-console.conf, fixed that.

PR #1037, opened 2021-09-06
XRDDEV-1763 dialog bugfix

Riippi:
https://jira.niis.org/browse/XRDDEV-1763
Fixes a browser console warning in subsystem > services view.


carohauta (approved):
🦄

PR #1040, opened 2021-09-09
Xrddev 1733 securityserver timestamping

turkkaholmqvist:
Tests for timestamping service selection, tests are based on Acceptance criteria given in https://jira.niis.org/browse/XRDDEV-1733. 

Things to consider:
- Some minor asserts are being omitted, for example it doesn't particulary test whether only one selection is possible.
- Test file could be further improved by making further nested objects for each dialog separately. Now element naming can be bit hard to follow and prone to simple mistakes since element purpose and location are in same string, eg. `timestampingAddDialogServiceSelection` is rather long and hard to read, it could be improved further by making `timestampingAddDialog.serviceSelection`.  
- There is no fallback on test setup, ie. test is based on assumption that in start of test, there is no any timestamping service set and same service is removed at last test.


src/proxy-ui-api/frontend/tests/e2e/pages/tabs/settingsTab.js
```diff
@@ -25,6 +25,7 @@
  */
 
 const path = require('path');
+var assert = require('assert');
```
carohauta, line 28: Should use `const` for immutable variables.

src/proxy-ui-api/frontend/tests/e2e/pages/tabs/settingsTab.js
```diff
@@ -41,7 +42,44 @@ const settingsTabCommands = {
...
+  assertTimestampingTableContents: function (expectedName, expectedUrl) {
+    // Service table row is visible
+    this.waitForElementVisible('@timestampingServiceTableRow');
+    // Delete button is visible in row
+    this.waitForElementVisible('@timestampingDeleteButton');
+    this.getText('@timestampingTableFirstCell', function (foundName) {
+      assert.equal(foundName.value, expectedName);
+    });
+    this.getText('@timestampingTableSecondCell', function (foundUrl) {
+      console.log(foundUrl.value);
```
carohauta, line 76: Is this for debugging purposes? Probably not needed

src/proxy-ui-api/frontend/tests/e2e/pages/tabs/settingsTab.js
```diff
@@ -118,9 +156,69 @@ const settingsTab = {
   sections: {
     systemParametersTab: {
       selector:
-        '//div[contains(@class, "v-tabs-bar__content")]//a[@data-test="system"]',
+        '//div[.//a[contains(@class, "v-tab--active") and @data-test="settings"]]//div[contains(@class, "base-full-width")]',
```
carohauta, line 159: Consider using a corresponding data-test attribute with value `system-parameters-tab` instead of looking up a div with `base-full-width` css class.

src/proxy-ui-api/frontend/tests/e2e/pages/tabs/settingsTab.js
turkkaholmqvist: Fixed, used `const` instead

src/proxy-ui-api/frontend/tests/e2e/pages/tabs/settingsTab.js
turkkaholmqvist: True, removed now.

src/proxy-ui-api/frontend/tests/e2e/pages/tabs/settingsTab.js
turkkaholmqvist: Added data-test attributes to actual code and used shorter and more descriptive locators where possible.

carohauta (approved):
🦄

PR #1043, opened 2021-09-15
XRDDEV-1626: Implemented new certificate profile Generic

rihohiiepuu:
Added GenericCertificateProfileInfoProvider and added tests to it.
Added few new DnFieldDescription labelKeys to reflect the new certificate profile
Added those keys to translation file.

Profile should be more compliant with the x.509 standard, so that more CAs could use this.

This code was built, tested and run on a Central Server and Security Server combination, with a CA set up to use the new profile.


src/common-util/src/main/java/ee/ria/xroad/common/certificateprofile/impl/GenericCertificateProfileInfoProvider.java
```diff
@@ -0,0 +1,144 @@
...
+
+    /**
+     * Auth cert
+     *
+     * CN = serverCode
+     * C = country
+     * serialNumber = memberCode
+     * O = memberName
+     */
+    private static class GenericAuthCertificateProfileInfo extends AbstractCertificateProfileInfo
```
petkivim, line 68: Since the authentication certificate is just a TLS clientAuth + serverAuth certificate, it should include the DNS name if some commonly used CA is used to issue the certs. Also, country code and all the fields in the ServerId should be present in a generic profile. In that way, it's possible to identify the Security Server that the certificate was issued to.

In addition, CN is actually deprecated as a server DNS name and subjectAlternateName should be used instead.

For example, using the Icelandic [auth cert profile](https://github.com/nordic-institute/X-Road/blob/develop/src/common-util/src/main/java/ee/ria/xroad/common/certificateprofile/impl/IsAuthCertificateProfileInfo.java) and setting `setReadOnly(false)` for the country code would include all the necessary information.

src/common-util/src/main/java/ee/ria/xroad/common/certificateprofile/impl/GenericCertificateProfileInfoProvider.java
```diff
@@ -0,0 +1,144 @@
...
+    /**
+     * Sign cert
+     *
+     * CN = memberName
+     * O = memberName
+     * businessCategory = memberClass
+     * C = country
+     * serialNumber = memberCode
+     */
+    private static class GenericSignCertificateProfileInfo extends AbstractCertificateProfileInfo
```
petkivim, line 95: The current implementation is missing instance identifier which makes it difficult to unambiguously map the signing certificate to a member. All fields in ClientID, country code and all fields in ServerId should be present in a generic profile. In that way, it's possible to identify the member who the certificate was issued to and the Security Server where it's used.

For example, using the Icelandic [sign cert profile](https://github.com/nordic-institute/X-Road/blob/develop/src/common-util/src/main/java/ee/ria/xroad/common/certificateprofile/impl/IsSignCertificateProfileInfo.java) and setting `setReadOnly(false)` for the country code would include all the necessary information.

petkivim (changes requested):
Thanks for the contribution. Some changes are required to the pull request.

src/common-util/src/main/java/ee/ria/xroad/common/certificateprofile/impl/GenericCertificateProfileInfoProvider.java
rihohiiepuu: The instance id was left out on purpose. CAs do not have a specific field to set X-Road instance id on, like in the case of SK profile. We can not presume, that CAs will make X-Road compliant certificate profiles.

Another important reason is that not all X-Road instances are country based, there are multiple X-Road instances around the world that are used by private businesses, in such case, the instance ID is even harder to put somewhere in the certificate.

rihohiiepuu: As of such,  X-Road instance should not be mentioned in the certificate as the certification in a very standard approach is about the subject. The most important purpose of this change is to be able to use existing standard sign certification services on X-Road in small starting  (also non-governmental) X-Road instances.

petkivim: I agree that we cannot presume that CAs will make X-Road compliant certificate profiles. However, the `serialNumber` field is already used in the profile and instead of containing only the member code, it could contain other information as well, e.g., `ServerId` which includes the instance identifier too.

In the current implementation it's possible to have multiple certificates with the same `Subject Distinguished Name` value. When a member has a sign certificate on multiple Security Servers, all the sign certificates have the same `Subject Distinguished Name`. This could be an issue with some CAs and it could be avoided by including the server code in the profile.

rihohiiepuu:
Renamed profile from Generic to Basic.
Changed auth certificate CN field to expect server DNS name instead of server code.
Added Country (C) field into both auth and sign certificate profile.

jhyoty:
Overall looks good to me. Couple of comments:
- Note: businessCategory is not one of the standard attribute types mentioned in RFC 5280, and the certificate details view displays it as an OID:  `Subject Distinguished Name: OID.2.5.4.15=memberClass, SERIALNUMBER=memberCode, ...`. 
- The country code is a required field, is that intentional?

jhyoty:
OK, just verifying that this is intentional.

jhyoty (approved):

PR #1045, opened 2021-09-17
XRDDEV-1749 Health check global conf

carohauta:
JIRA: https://jira.niis.org/browse/XRDDEV-1749

### DIFF
--- a/src/proxy/src/main/java/ee/ria/xroad/common/util/healthcheck/StoppableCombinationHealthCheckProvider.java
+++ b/src/proxy/src/main/java/ee/ria/xroad/common/util/healthcheck/StoppableCombinationHealthCheckProvider.java
@@ -79,14 +79,14 @@ public class StoppableCombinationHealthCheckProvider implements StoppableHealthC
         final int errorValidFor = 30;
 
         return Arrays.asList(
+                checkGlobalConfStatus()
+                        .map(withTimeout(timeout, TimeUnit.SECONDS, "Global conf validity"))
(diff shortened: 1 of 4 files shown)


src/proxy/src/main/java/ee/ria/xroad/common/util/healthcheck/StoppableCombinationHealthCheckProvider.java
```diff
@@ -83,6 +84,9 @@ public HealthCheckResult get() {
                         .map(cacheResultFor(resultValidFor, errorValidFor, TimeUnit.SECONDS)),
                 checkServerConfDatabaseStatus()
                         .map(withTimeout(timeout, TimeUnit.SECONDS, "Server conf database status"))
+                        .map(cacheResultFor(resultValidFor, errorValidFor, TimeUnit.SECONDS)),
+                checkGlobalConfStatus()
```
jhyoty, line 88: Global configuration check should be the first (cheap and fast unless the filesystem remote --> timeout and caching is necessary in edge cases).

Not changed in this issue, but the checkServerConfDatabaseStatus() does not work well -- due to caching in CachingServerConfImpl it does not actually check that database connection works once the security server identifier has been successfully loaded.

carohauta: Sounds reasonable. Moved it up in the list. 

Did not address the issue with `checkServerConfDatabaseStatus()` in this PR. Should we create a ticket about that?

jhyoty (approved):

Issue #1054, opened 2021-09-23
Compatibility issues with lxd 4 and Fedora when setting up the ansible-based dev environment


mbert:
On my Fedora system it took several steps to get the playbook `xroad_init.yml` running:
- The `lxd` installation used on my machine creates the Unix socket in a place where the `lxd_container` role searches for it.
- The host names in the `lxd_hosts.txt` inventory contain non-alphanumeric characters which is not accepted by `lxd` version 4.

As a remedy I needed to make the socket configurable and change the host names in the inventory.

PR #1055, opened 2021-09-23
Issue #1054: fix compatibility issues with LXD4 and Fedora.

mbert:
This patch fixes the above problem by changing the host names in the LXD example inventory and making the path to the socket for connecting to LXD configurable.

### DIFF
--- a/ansible/hosts/lxd_hosts.txt
+++ b/ansible/hosts/lxd_hosts.txt
@@ -1,27 +1,27 @@
 #central servers (ubuntu lxd containers)
 [cs_servers]
-xroadlxdcs ansible_connection=lxd
+xroad-lxd-cs ansible_connection=lxd
 
 #configuration proxies (ubuntu lxd containers, optional)
 [cp_servers]
-#xroadlxdcp ansible_connection=lxd
+#xroad-lxd-cp ansible_connection=lxd
 
 #certification authority, time stamping authority and ocsp service server for testing purposes (ubuntu)
 [ca_servers]
-xroadlxdcs ansible_connection=lxd
+xroad-lxd-cs ansible_connection=lxd
 
 #security servers (ubuntu lxd containers)
 [ss_servers]
-xroadlxdss1 ansible_connection=lxd
+xroad-lxd-ss1 ansible_connection=lxd
 #for a more realistic setup, add at least one additional server
-#xroadlxdss2 ansible_connection=lxd
+#xroad-lxd-ss2 ansible_connection=lxd
 
 [ss_servers:children]
 centos_ss
 
 #security servers (centos lxd containers, not supported in variant ee)
 [centos_ss]
-#xroadlxdrhss1 ansible_connection=lxd
+#xroad-lxd-rh-ss1 ansible_connection=lxd
 
 #container host
 [lxd_servers]
--- a/ansible/hosts/lxd_hosts.txt
+++ b/ansible/hosts/lxd_hosts.txt
@@ -1,27 +1,27 @@
 #central servers (ubuntu lxd containers)
 [cs_servers]
-xroad-lxd-cs ansible_connection=lxd
+xroadlxdcs ansible_connection=lxd
 
 #configuration proxies (ubuntu lxd containers, optional)
 [cp_servers]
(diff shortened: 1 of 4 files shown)


jhyoty:
Looks like initializing containers using LXD 4.19 (snap) and Ansible 2.9 on Fedora 35 works without these modifications perhaps the problem is elsewhere? Some firewalld configuration was necessary so that the containers could get IP address.

mbert:
> Looks like initializing containers using LXD 4.19 (installed using snap) and Ansible 2.9 on Fedora 35 works without these modifications, perhaps the problem is elsewhere? Some firewalld configuration was necessary so that the containers could get an IP address.

Indeed I did not install lxd using snap (like probably many others I don't like the idea of several different package managers on my system, hence I used this yum repo: https://copr.fedorainfracloud.org/coprs/ganto/lxc4/)

Since the path to the Unix domain socket can differ depending on the packager's choice, making it configurable with a sane default IMO makes sense here.

ansible/roles/restart-lxd/tasks/main.yml
```diff
@@ -2,4 +2,6 @@
   with_inventory_hostnames: ss_servers, cs_servers, cp_servers
   lxd_container:
     name: "{{item}}"
+    url: "{{ lxd_url | default('unix:/var/lib/lxd/unix.socket') }}"
+    snap_url: "{{ lxd_url | default('unix:/var/lib/lxd/unix.socket') }}"
```
jhyoty, line 6:
```suggestion
    url: "{{ lxd_url | default(omit) }}"
```
In case lxd_url is not defined, the lxd container module should be allowed to internally discover the URL.
And there should be no need to override the snap_url (the default here breaks the playbook on Ubuntu, or when using LXD from snap).

ansible/roles/init-lxd/tasks/main.yml
```diff
@@ -3,6 +3,8 @@
   with_inventory_hostnames: all:!lxd_servers:!centos_ss
   lxd_container:
     name: "{{item}}"
+    url: "{{ lxd_url | default('unix:/var/lib/lxd/unix.socket') }}"
+    snap_url: "{{ lxd_url | default('unix:/var/lib/lxd/unix.socket') }}"
```
jhyoty, line 7:
```suggestion
    url: "{{ lxd_url | default(omit) }}"
```

ansible/roles/init-lxd/tasks/main.yml
```diff
@@ -20,6 +22,8 @@
   with_inventory_hostnames: centos_ss
   lxd_container:
     name: "{{item}}"
+    url: "{{ lxd_url | default('unix:/var/lib/lxd/unix.socket') }}"
+    snap_url: "{{ lxd_url | default('unix:/var/lib/lxd/unix.socket') }}"
```
jhyoty, line 26:
```suggestion
    url: "{{ lxd_url | default(omit) }}"
```

ansible/hosts/lxd_hosts.txt
```diff
@@ -1,27 +1,27 @@
...
-#xroad-lxd-ss2 ansible_connection=lxd
+#xroadlxdss2 ansible_connection=lxd
 
 [ss_servers:children]
 centos_ss
 
 #security servers (centos lxd containers, not supported in variant ee)
 [centos_ss]
-#xroad-lxd-rh-ss1 ansible_connection=lxd
+#xroadlxdrhss1 ansible_connection=lxd
```
jhyoty, line 24: I did not encounter any issues with using LXD4 from copr-lxc4 or LXD 4.19 from snap on Fedora 35, or LXD 4.19 on Ubuntu 20.04 using the example inventory. And in any case, this is just an example (and xroadlxdrhss1 reads quite badly).

jhyoty (changes requested):

ansible/hosts/lxd_hosts.txt
jhyoty: Could you revert the changes to lxd_hosts.txt (or please provide more details of the problem). I was not able to reproduce the issue regarding host names.

jhyoty (approved):

PR #1059, opened 2021-09-28
XRDDEV-1771: Support encryption in asic container download

jhyoty:


petkivim:
This change should be mentioned in the Signed Document Download Service user manual:

https://github.com/nordic-institute/X-Road/blob/develop/doc/Manuals/ug-sigdoc_x-road_signed_document_download_and_verification_manual.md#2-signed-document-download-service

petkivim:
I'm sorry, I somehow missed that. I think that mentioning it like that is OK for now. But shouldn't the parameter be `archive-encryption-enabled` and not `messagelog-encryption-enabled`? The service returns the archive files encrypted if message log archive encryption is enabled. Instead, message log database encryption shouldn't affect the files returned by the service.

iluwatar:
> I'm sorry, I somehow missed that. I think that mentioning it like that is OK for now. But shouldn't the parameter be `archive-encryption-enabled` and not `messagelog-encryption-enabled`? The service returns the archive files encrypted if message log archive encryption is enabled. Instead, message log database encryption shouldn't affect the files returned by the service.

You are right, it should be `archive-encryption-enabled`. Seems to be a documentation mistake only and I think @jhyoty just pushed a fix to that in https://github.com/nordic-institute/X-Road/pull/1059/commits/4ec5d4db9e2628e462bc4b81e903db23cc9ae2b2

PR #1084, opened 2021-10-29
XRDDEV-1809: Enhance sidecar support

jhyoty:
* Refactor installer to better support building a sidecar image (make it
  possible to skip database installation).
* Take into account that xroad.properties in a docker container can be a
  mounted file (needs to be modifid in-place).
* Make it possible to locate xroad.propeties in /etc/xroad

### DIFF
--- a/src/packages/src/xroad/ubuntu/generic/xroad-proxy.config
+++ b/src/packages/src/xroad/ubuntu/generic/xroad-proxy.config
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/bin/bash
 . /usr/share/debconf/confmodule
 
 get_prop() {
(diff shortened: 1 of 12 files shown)


src/packages/src/xroad/ubuntu/generic/xroad-proxy.config
```diff
@@ -8,7 +8,9 @@ get_prop() {
 if [ "$1" = "configure" ] || [ "$1" = "reconfigure" ]; then
   while :; do
     RET=""
-    db_input high xroad-common/username || true
+    if db_input high xroad-common/username; [ $? -eq 30 ]; then
```
jansu76, line 11: Could maybe comment what is the significance of value "30"

jansu76 (changes requested):

jansu76 (approved):

PR #1085, opened 2021-10-31
XRDDEV-1830 Fix incorrect ordering

petkivim:
- Fix incorrect member, client, subsystem and service client ordering.
- Update tests so that incorrect ordering is detected.


src/proxy-ui-api/src/main/java/org/niis/xroad/securityserver/restapi/openapi/ClientsApiController.java
```diff
@@ -575,14 +583,16 @@ private ClientType getClientType(String encodedId) {
...
     public ResponseEntity<Set<ServiceClient>> getClientServiceClients(String id) {
         ClientId clientId = clientConverter.convertId(id);
-        Set<ServiceClient> serviceClients = null;
+        SortedSet<ServiceClient> sortedServiceClients = null;
         try {
-            serviceClients = serviceClientConverter.
+            Set<ServiceClient> serviceClients = serviceClientConverter.
                     convertServiceClientDtos(serviceClientService.getServiceClientsByClient(clientId));
+            sortedServiceClients = new TreeSet<>(serviceClientSortingComparator);
+            sortedServiceClients.addAll(serviceClients);
```
jhyoty, line 591: The converter creates a HashSet that is immediately converted to a TreeSet. Both data structures have some overhead, so it would be more efficient to make the converter directly return an ordered set (e.g. by providing the result set as a parameter, or ordering the input list and returning a LinkedHashSet).

src/proxy-ui-api/src/test/java/org/niis/xroad/securityserver/restapi/openapi/ClientsApiControllerIntegrationTest.java
```diff
@@ -165,6 +165,11 @@ public void getAllClients() {
                 clientsApiController.findClients(null, null, null, null, null, true, false, null, false);
         assertEquals(HttpStatus.OK, response.getStatusCode());
         assertEquals(11, response.getBody().size());
+        // Test sorting order
+        List<Client> list = new ArrayList<>(response.getBody());
+        assertEquals("EE:PRO:M1:SS1", list.get(0).getId());
+        assertEquals("DUMMY:PRO:M2:SS6", list.get(5).getId());
+        assertEquals("FI:GOV:M1", list.get(10).getId());
```
jhyoty, line 172: Hard-coding the expected order makes the tests brittle. Perhaps the test should verify that the ordering holds (that is, list(n) <= list(n+1)).

src/proxy-ui-api/src/main/java/org/niis/xroad/securityserver/restapi/openapi/ClientsApiController.java
petkivim: The implementation was moved to the Converter that orders the input list and returns a LinkedHashSet.

src/proxy-ui-api/src/test/java/org/niis/xroad/securityserver/restapi/openapi/ClientsApiControllerIntegrationTest.java
petkivim: The tests were updated so that they now validate the order using the related Comparator. The same Comparators are used for sorting too, but the Comparators have their own unit tests that verify that they're working correctly. In this way, the Comparators logic is tested elsewhere and the tests in this PR only validate that sorting is applied to API endpoints too.

src/proxy-ui-api/src/test/java/org/niis/xroad/securityserver/restapi/util/TestUtils.java
```diff
@@ -365,4 +368,27 @@ public static void mockServletRequestAttributes() {
...
         RequestContextHolder.setRequestAttributes(attributes);
     }
+
+    /**
+     * Checks if the sort order of the given Set is correct.
+     * @param set
+     * @param comparator
+     * @return
+     */
+    public static boolean isSortOrderCorrect(Set set, Comparator comparator) {
```
jhyoty, line 378: Raw types should not be used, consider
`public static <T> boolean isSortOrderCorrect(Set<T> set, Comparator<? super T> comparator)`
(requires also other changes)
And since we have Guava as a dependency, this function can be simplified to `Ordering.from(comparator).isOrdered(set)`

jhyoty (changes requested):

src/proxy-ui-api/src/test/java/org/niis/xroad/securityserver/restapi/util/TestUtils.java
petkivim: I removed the custom logic and used Guava instead. Also, the method signature was updated.

jhyoty (approved):

PR #1091, opened 2021-11-03
Xrddev 1782   Session status polling & logout to CS, too

teemukin65:
I moved the related backend machinery to common-rest-api, and removed from proxy-ui-api.


src/common-rest-api/src/main/java/org/niis/xroad/restapi/config/NotificationApiDisableSessionTimeoutFilter.java
```diff
@@ -55,7 +55,7 @@
 @Slf4j
 @Configuration
 @Order(Ordered.HIGHEST_PRECEDENCE)
-public class SessionTimeoutFilter extends GenericFilterBean {
+public class NotificationApiDisableSessionTimeoutFilter extends GenericFilterBean {
```
carohauta, line 58: Is there another `SessionTimeoutFilter` or why the complex class name?

src/common-rest-api/src/main/java/org/niis/xroad/restapi/controller/SessionStatusController.java
```diff
@@ -0,0 +1,70 @@
...
+
+@RestController
+@RequestMapping(SessionStatusController.NOTIFICATIONS_API_V1_PATH + "/session-status")
+@Slf4j
+@PreAuthorize("isAuthenticated()")
+@RequiredArgsConstructor
+public class SessionStatusController {
+
+
+    public static final String NOTIFICATIONS_API_V1_PATH = ControllerUtil.API_V1_PREFIX + "/notifications";
```
carohauta, line 52: Feels a bit sketchy that this part of the `/notifications` api is in this controller (in the common api), but some parts of the api remain only in `proxy-ui-api`. I wonder if this would be enough reason to move the session polling in to its own endpoint instead of `/notifications`. 

What do you think? Moving the session polling to another path is a breaking change but then again it is supposed to be used by the frontend only (and not by some arbitrary api-client).

src/centralserver/admin-ui/src/views/AppLogin.vue
```diff
@@ -159,7 +159,7 @@ export default (
   methods: {
     async submit() {
       // Clear error notifications when route is changed
-      this.$store.commit(StoreTypes.actions.RESET_NOTIFICATIONS_STATE);
+      await this.$store.dispatch(StoreTypes.actions.RESET_NOTIFICATIONS_STATE);
```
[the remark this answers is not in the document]
teemukin65: https://vuex.vuejs.org/api/#dispatch  states, that Vue `dispatch` returns a promise. 
And `await `-construct in general is intended to wait resolving of an ( asynchronous ) statement - and returns a resolved Promise. 

But besides of that,  what would be the more implicit way to show that?

src/common-rest-api/src/main/java/org/niis/xroad/restapi/config/NotificationApiDisableSessionTimeoutFilter.java
teemukin65: Yes, src/common-ui/src/main/java/ee/ria/xroad/commonui/SessionTimeoutFilter.java

src/common-rest-api/src/main/java/org/niis/xroad/restapi/controller/SessionStatusController.java
teemukin65: The NotificationApiDisableSessionTimeoutFilter enables making requests without pushing the session timeout further. That filter is based on certain API path - `/notifications`.  
That feature is needed for these periodic status polling requests. If endpoints are not concentrated to a single path, this filter has to be adjusted accordingly.  This kind of filter is a awkwardly hidden dependency. i.e. It's not apparent when you look in the controller that there is special session timeout handling applied for these requests. 

We could naturally make the filter select the polling requests according to some other rule. e.g. special HTTP header or some URL parameter values. But it poses new requirements then to the endpoint user -

teemukin65: What about changing the path of the endpoint  from `/notifications` to  perhaps more descriptive `/polling` ?

carohauta: Or maybe the `NotificationsApiController` class under proxy-ui-api could be renamed `NotificationsAlertsApiController` since it only handles `/notifications/alerts` now. And the `SessionStatusController` under common-api could be `NotificationsSessionStatusApiController` since it only handles `/notifications/session-status` path. Perhaps that would add more relation between these controllers (?)

src/centralserver/admin-ui/src/views/AppLogin.vue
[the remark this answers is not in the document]
carohauta: Yeah, seems that `async` is not required here since `dispatch` always returns a promise – even though the action itself is not returning a promise. What makes it confusing is that you can declare async actions as well, in Vuex
https://vuex.vuejs.org/guide/actions.html#composing-actions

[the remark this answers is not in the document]
teemukin65: That's one of the essences of the  wonderful world of promises.. Promises of promises of promises ....

src/common-rest-api/src/main/java/org/niis/xroad/restapi/controller/SessionStatusController.java
teemukin65: Done in : 7179dad  & 
4d6b2da

src/common-rest-api/src/main/java/org/niis/xroad/restapi/config/NotificationApiDisableSessionTimeoutFilter.java
```diff
@@ -55,14 +55,14 @@
...
 
     private static final String ORIGINAL_MAX_INACTIVE_ATTR =
             "originalMaxInactiveInterval";
 
     @Override
     public void doFilter(ServletRequest request, ServletResponse response,
-            FilterChain chain) throws IOException, ServletException {
+                         FilterChain chain) throws IOException, ServletException {
```
carohauta, line 65: Does this not cause checkstyle error now? Indentation should be 8 instead of 21 according to the checkstyle rules

teemukin65: fixed 
1887dc0

carohauta (approved):
🦄

PR #1093, opened 2021-11-03
XRDDEV-1801 Security Server node info

carohauta:
This PR consists of backend changes. 

Target version: 7.1

JIRA: https://jira.niis.org/browse/XRDDEV-1801


src/proxy-ui-api/src/main/resources/openapi-definition.yaml
```diff
@@ -6203,6 +6229,27 @@ components:
...
+        </ul>
+      example: PRIMARY
+      enum:
+        - PRIMARY
+        - REPLICA
+        - STANDALONE
+    NodeTypeResponse:
+      type: object
+      description: Response object containing the node type of the Security Server
+      properties:
```
teemukin65, line 6250: Is the node_type property really optional, or should we have the required - field list here?

src/proxy-ui-api/src/main/java/org/niis/xroad/securityserver/restapi/converter/NodeTypeMapping.java
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Mapping between NodeType in api (enum) and model (SystemProperties.NodeType)
+ */
+@Getter
+@RequiredArgsConstructor
+public enum NodeTypeMapping {
+    PRIMARY(SystemProperties.NodeType.MASTER, NodeType.PRIMARY),
+    REPLICA(SystemProperties.NodeType.SLAVE, NodeType.REPLICA),
+    STANDALONE(SystemProperties.NodeType.STANDALONE, NodeType.STANDALONE);
```
teemukin65, line 45: What is the reason/idea behind this term translation?  ie. MASTER -> PRIMARY  , SLAVE -> REPLICA

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
```diff
@@ -4489,13 +4489,39 @@ paths:
...
+      operationId: getNodeType
+      description: <h3>Administrator views the node type</h3>
+      responses:
+        '200':
+          description: node type information
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/NodeTypeResponse'
+        '400':
```
teemukin65, line 4506: How the user-provided parameters could be wrong now in the parameterless endpoint?   If not possible, I think this 400 should not be listed here neither.
Is there possibly some special auth or X-Road headers that could be in wrong format?

src/proxy-ui-api/src/test/java/org/niis/xroad/securityserver/restapi/openapi/SystemApiControllerTest.java
```diff
@@ -335,4 +339,27 @@ public void previewAnchor() throws IOException {
...
+    public void getNodeTypeStandalone() {
+        ResponseEntity<NodeTypeResponse> response = systemApiController.getNodeType();
+        assertEquals(response.getBody().getNodeType(), NodeType.STANDALONE); // default value is STANDALONE
+    }
+
+    @Test
+    @WithMockUser(authorities = { "VIEW_NODE_TYPE" })
+    public void getNodeTypePrimary() {
+        when(systemService.getServerNodeType()).thenReturn(SystemProperties.NodeType.MASTER);
+        ResponseEntity<NodeTypeResponse> response = systemApiController.getNodeType();
```
teemukin65, line 354: Should there be test for case where the getServerNodeType throws exception?

teemukin65 (changes requested):

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
carohauta: Yeah, it is required in the response. Fixed

src/proxy-ui-api/src/main/java/org/niis/xroad/securityserver/restapi/converter/NodeTypeMapping.java
carohauta: I think we are mostly referring to the master node as `primary` and the slave node as `replica` in the documentation. Thought that might be clearer to the end user.

src/proxy-ui-api/src/main/resources/openapi-definition.yaml
carohauta: I removed all error responses that are not possible. Left `401`, `403` and `500` since those are at least theoretically possible.

src/proxy-ui-api/src/test/java/org/niis/xroad/securityserver/restapi/openapi/SystemApiControllerTest.java
carohauta: The node type comes from a `.ini` file which is loaded into `SystemProperties` on startup. I think that it should be highly unlikely to get an exception from this method. It would probably mean that there were some errors loading the system properties in the first place, which would likely cause the system to not even start

PR #1103, opened 2021-11-15
XRDDEV-1826 Add extra padding bottom to clients data table

carohauta:
Add some padding to Security Server clients data table as seen in 7.x Central Server

JIRA: https://jira.niis.org/browse/XRDDEV-1826


Riippi (approved):
LGTM

PR #1105, opened 2021-11-17
XRDDEV-1844 Fix tabs slider bug

carohauta:
JIRA: https://jira.niis.org/browse/XRDDEV-1844

* upgrade Vuetify to 2.6.0
* change sub tab path matching to exact-path


Riippi (approved):
LGTM

PR #1123, opened 2021-12-02
XRDDEV-1861 Fix service tables title column

Riippi:
https://jira.niis.org/browse/XRDDEV-1861


Riippi:
Merged this by myself because the change was really small

PR #1128, opened 2021-12-09
Fix css classes in localgroups table

Riippi:
Fix class names for e2e tests


Riippi:
Merged this by myself as the changes were minimal

PR #1129, opened 2021-12-10
Add data-test for service clients table

Riippi:
Add data-test id for service clients table to fix e2e tests


Riippi:
Merging this by myself as the change is minimal

PR #1133, opened 2021-12-15
Xrddev 1807  CS address edit e2e test cases

teemukin65:


src/centralserver/admin-ui/tests/e2e/specs/040-cs-system_settings.js
```diff
@@ -0,0 +1,77 @@
...
+  'SystemSettings/System Parameters view has correct titles and ServerAddress edit opens correct dialog':
+    async (browser) => {
+      await settings
+        .systemSettingsViewIsVisible()
+        .verify.elementPresent('@systemSettingsParametersCard')
+        .verify.elementPresent('@systemSettingsParametersTitle')
+        .verify.elementPresent('@systemSettingsInstanceIdentifierFieldTitle')
+        .verify.containsText(
+          '@systemSettingsInstanceIdentifierFieldTitle',
+          'Instance Identifier',
```
Riippi, line 55: Seems that there are a couple of tests for titles that are just localised text. Is there any benefit on testing those? 
I think we should check only the essential "moving parts". Also maybe all the data-test tags are not needed if they are just for checking texts.

Riippi (changes requested):

src/centralserver/admin-ui/src/views/Settings/SystemSettings/SystemSettings.vue
```diff
@@ -182,7 +199,7 @@
                   </div>
                 </div>
               </td>
-              <td>
+              <td data-test="">
```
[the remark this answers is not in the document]
teemukin65: fixed: 1589a4d , also removed all data-test entries for fixed titles. Added similarly to other fields in the component

src/centralserver/admin-ui/tests/e2e/specs/040-cs-system_settings.js
teemukin65: Tests for fixed texts removed. acc4a39

teemukin65: And test case passes still locally.

Riippi (approved):

PR #1134, opened 2021-12-15
XRDDEV-1472 Update search methods

Riippi:
https://jira.niis.org/browse/XRDDEV-1472

Unifying the search methods in tables

### DIFF
--- a/src/proxy-ui-api/frontend/src/views/Clients/Services/Services.vue
+++ b/src/proxy-ui-api/frontend/src/views/Clients/Services/Services.vue
@@ -303,22 +303,24 @@ export default Vue.extend({
       }
 
       // Filter out service descriptions that don't include search term
-      const filtered = arr.filter((element) => {
-        return element.services.find((service) => {
-          return service.service_code
-            .toString()
-            .toLowerCase()
-            .includes(mysearch);
+      const filtered = arr.filter((element: ServiceDescription) => {
+        return element.services.find((service: Service) => {
+          return (
+            service.service_code.toString().toLowerCase().includes(mysearch) ||
+            service.url.toString().toLowerCase().includes(mysearch) ||
+            service.timeout.toString().toLowerCase().includes(mysearch)
+          );
         });
       });
 
       // Filter out services that don't include search term
-      filtered.forEach((element) => {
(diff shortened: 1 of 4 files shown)


src/proxy-ui-api/frontend/src/views/KeysAndCertificates/SignAndAuthKeys/SignAndAuthKeys.vue
```diff
@@ -120,25 +125,34 @@ export default Vue.extend({
...
           key.certificates = certs;
+
+          // Filter the CSR:s
+          const csrs = key.certificate_signing_requests.filter(
+            (csr: TokenCertificateSigningRequest) => {
+              if (csr.id) {
+                return csr.id.toLowerCase().includes(mysearch);
+              }
+              return false;
+            },
```
teemukin65, line 153: Should it be possible to filter with word "Request" , which is visible in entries for  CSR:s?  If you want to see only CSR:s...

teemukin65 (changes requested):

src/proxy-ui-api/frontend/src/views/Clients/Services/Services.vue
```diff
@@ -303,22 +303,24 @@ export default Vue.extend({
...
       });
 
       // Filter out services that don't include search term
-      filtered.forEach((element) => {
-        element.services = element.services.filter((service) => {
-          return service.service_code
-            .toString()
-            .toLowerCase()
-            .includes(mysearch);
+      filtered.forEach((element: ServiceDescription) => {
```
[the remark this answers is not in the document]
Riippi: forEach returns void so it can't return in this case. This would need a larger refactor.

src/proxy-ui-api/frontend/src/views/KeysAndCertificates/SignAndAuthKeys/SignAndAuthKeys.vue
Riippi: In practice it should filter with the localised word for "Request". Seems to me a bit too complicated system for a very rare use case. But I'll mention this in a sprint review.

PR #1140, opened 2021-12-20
XRDDEV-1856 Align alerts with the styleguide

carohauta:
* margin-top 16px instead of 8px

JIRA: https://jira.niis.org/browse/XRDDEV-1856


Riippi (approved):
LGTM

PR #1143, opened 2021-12-27
XRDDEV-1862 CS loading placeholders

Riippi:
https://jira.niis.org/browse/XRDDEV-1862

### DIFF
--- a/src/centralserver/admin-ui/src/views/Settings/SystemSettings/SystemSettings.vue
+++ b/src/centralserver/admin-ui/src/views/Settings/SystemSettings/SystemSettings.vue
@@ -132,7 +132,14 @@
           </div>
         </xrd-simple-dialog>
         <table class="xrd-table mt-0 pb-3">
-          <tbody>
+          <XrdEmptyPlaceholderRow
+            :colspan="2"
+            :loading="loadingServices"
+            :data="managementServices"
+            no-items-text="no items"
+          />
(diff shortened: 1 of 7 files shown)


src/centralserver/admin-ui/src/views/Settings/SystemSettings/SystemSettings.vue
```diff
@@ -132,7 +132,14 @@
           </div>
         </xrd-simple-dialog>
         <table class="xrd-table mt-0 pb-3">
-          <tbody>
+          <XrdEmptyPlaceholderRow
```
carohauta, line 135: This table is a bit different from the `Member classes` table below because this one is not a Vuetify component. Therefore the loading state also has a different look (no loading bar). 

Could this be somehow unified by at least using the same Vuetify loading bar in both tables?

This part here actually also results in a broken table structure because the `tr` element provided by `XrdEmptyPlaceholderRow` is not inside a `tbody`. Probably does not affect anything anyhow but just to point this out

Riippi: Different loading bar was due a wrong build of shared-ui.
In this case `tbody` is optional for the `table` because there are no separate header or footer elements.

carohauta (approved):
🦄

PR #1147, opened 2022-01-05
XRDDEV-1863 pinia

Riippi:
https://jira.niis.org/browse/XRDDEV-1863
Replace vuex with pinia


src/centralserver/admin-ui/src/components/ui/InitNotification.vue
```diff
@@ -42,7 +42,7 @@
         color="primary"
         outlined
         data-test="notification-button"
-        @click="continueInit"
+        @click="setContisdfnueInit(false)"
```
teemukin65, line 45: I think there is some typo. At least incorrect language, if not unknown function?
```suggestion
        @click="setContinueInit(false)"
```

src/centralserver/admin-ui/src/main.ts
```diff
@@ -54,8 +61,8 @@ axios.defaults.headers.get.Accepts = 'application/json';
 Vue.use(Router);
 
 new Vue({
+  pinia,
```
teemukin65, line 64: Why this is at the beginning of the list?   Example at: https://pinia.vuejs.org/getting-started.html#installation hints that the Pinia is at the end of this list. I just want to learn...

src/centralserver/admin-ui/src/views/Members/MemberList.vue
```diff
@@ -30,15 +30,15 @@
...
+      <xrd-button data-test="add-member-button">
         <xrd-icon-base class="xrd-large-button-icon"
           ><xrd-icon-add
         /></xrd-icon-base>
 
         {{ $t('members.addMember') }}</xrd-button
       >
     </div>
-
+    numero: {{ getCount }}
```
teemukin65, line 41: Would this be a leftover from some intermediate draftings?  To be removed or localized...

src/centralserver/admin-ui/src/views/Settings/SystemSettings/SystemSettings.vue
```diff
@@ -378,29 +380,32 @@ export default (
...
-        await this.$store.dispatch(StoreTypes.actions.FETCH_SYSTEM_STATUS);
-        await this.$store.dispatch(
-          StoreTypes.actions.SHOW_SUCCESS,
+        await this.updateCentralServerAddress({
+          central_server_address: serverAddress,
+        });
+
+        await this.fetchSystemStatus();
+
+        this.showSuccess(
```
teemukin65, line 405: I think `await` should be added for this , too. Otherwise isEditingServerAddress can get false too early.

teemukin65 (changes requested):
Pinia provides for so much more  clean & concise code!

src/centralserver/admin-ui/src/components/layout/AppIcon.vue
```diff
@@ -37,15 +37,18 @@
...
+  computed: {
+    ...mapState(userStore, ['getFirstAllowedTab']),
+  },
   methods: {
     home(): void {
       this.$router
         .replace({
-          name: this.$store.getters[StoreTypes.getters.FIRST_ALLOWED_TAB]?.to
-            ?.name,
+          name: this.getFirstAllowedTab.to.name,
```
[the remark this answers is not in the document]
Riippi: `to` property is mandatory in Tab interface. So at least in theory the IDE should complain if someone is creating `Tab` without `to`. Actually it's possible to have `to` ( = Location) without `name`.  Anyways if we would have `Tab` without `to` or `name` it replace would just fail. And I don't know what else would be possible in this situation.

src/centralserver/admin-ui/src/views/AppLogin.vue
```diff
@@ -157,9 +161,16 @@ export default (
     },
   },
   methods: {
+    ...mapActions(userStore, ['login', 'fetchUserData']),
```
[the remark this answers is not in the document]
Riippi: I'm not sure if I understand the question, but mapActions adds login and fetchUserData as methods to this component.

src/centralserver/admin-ui/src/main.ts
Riippi: No real reason. I moved it to first when I was desperately trying to troubleshoot some issue. It didn't help real reason was somewhere else. 
Moved it to last now just so it looks like that example.  In practice it should not make any difference.

src/centralserver/admin-ui/src/views/Members/MemberList.vue
Riippi: Yes, this was never meant to be pushed in git.

src/centralserver/admin-ui/src/views/Settings/SystemSettings/SystemSettings.vue
Riippi: `await` would need a promise to be returned. `showSuccess` doesn't return anything because it's meant to be kind of "fire and forget" action.

src/centralserver/admin-ui/src/views/InitialConfiguration/InitialConfiguration.vue
```diff
@@ -278,9 +292,8 @@ export default (
       if (!this.disabledFields.pin) {
         formData.software_token_pin = this.pin;
       }
-      await this.$store.dispatch(StoreTypes.actions.RESET_NOTIFICATIONS_STATE);
-      await this.$store
-        .dispatch(StoreTypes.actions.INITIALIZATION_REQUEST, formData)
+      await this.resetNotifications();
+      await this.initalizationRequest(formData)
         .then(
```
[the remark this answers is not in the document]
Riippi: I fixed this. Also removed two if checks because they didn't affect the end result any ways.

src/centralserver/admin-ui/src/components/layout/AppIcon.vue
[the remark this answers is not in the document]
teemukin65: So it is OK to leave as you have selected here. 
No reason to change back.

src/centralserver/admin-ui/src/views/AppLogin.vue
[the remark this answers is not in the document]
teemukin65: Just interested to know 
if IDE helps you remembering the action names..

src/centralserver/admin-ui/src/views/Settings/SystemSettings/SystemSettings.vue
teemukin65: Ajoitus on tässä kai ainoa mahdollisesti virhettä aiheuttava asia. 
await on mahdollinen myös tässä: 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await
"If the value of the expression following the await operator is not a Promise, it's converted to a resolved Promise"

src/centralserver/admin-ui/src/views/InitialConfiguration/InitialConfiguration.vue
```diff
@@ -292,50 +292,38 @@ export default (
...
-                );
-                this.$refs.initializationForm.setErrors({
-                  'init.identifier': identifierErrors,
-                  'init.address': addressErrors,
-                });
-                this.showError(error);
-              }
-              return;
-            }
-            if (invalidParamsErrors(errorInfo).length > 0) {
```
teemukin65, line 313: This simplification was a good finding!
Is there any use left for the functions `invalidParamsErrors` and `isWeakPinError`? If not the function definitions could be removed also...

teemukin65 (changes requested):

src/centralserver/admin-ui/src/views/AppLogin.vue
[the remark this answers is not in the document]
Riippi: Yes. There is autocomplete. Looks like intelliJ underlines a wrong name with red, vscode doesn't, but that might depend on vue plugin. Also build shows an error if it's wrong.

src/centralserver/admin-ui/src/views/Settings/SystemSettings/SystemSettings.vue
Riippi: I don't see any issue with timing success snackbar. It's fired to show, then it shows up and goes away on it's own and rest of the app can go on during that time. Even if the snackbar would fail for some reason it should not stop anything.

src/centralserver/admin-ui/src/views/InitialConfiguration/InitialConfiguration.vue
Riippi: Seems that they are not used. I removed those now.

PR #1149, opened 2022-01-07
Xrddev 1821 ssMainpage streamlining

teemukin65:


src/proxy-ui-api/frontend/tests/e2e/pages/ssMainPage.js
```diff
@@ -693,374 +135,75 @@ module.exports = {
...
-        saveButton: '//button[@data-test="save-service-parameters"]',
-        closeButton: '//button[@data-test="close"]',
-        urlHelp: '//div[@data-test="service-parameters-service-url-label"]//i',
-        timeoutHelp: '//div[@data-test="service-parameters-timeout-label"]//i',
-        verifyCertHelp: '//div[@data-test="service-parameters-verify-tls-label"]//i',
-        activeTooltip: '//div[contains(@class, "v-tooltip__content") and contains(@class,"menuable__content__active")]//span',
-        addSubjectsButton: '//button[@data-test="show-add-subjects"]',
-        removeAllButton: '//button[@data-test="remove-subjects"]',
-      },
+      commands: [operationDetailsCommands],
```
turkkaholmqvist, line 180: These arrays are not necessary as long as there is only one import inside of it - it's not an issue but i wonder whether we should establish some sort of common practice that array is always created regardless if it has only one member or other way around, never use array for single reference.

iluwatar:
It's currently conflicting with `develop` branch

src/proxy-ui-api/frontend/vue.config.js
```diff
@@ -3,7 +3,7 @@ const path = require('path');
 
 module.exports = {
   devServer: {
-    proxy: process.env.PROXY_ADDRESS || 'https://localhost:4100',
+    proxy: process.env.PROXY_ADDRESS || 'https://localhost:5005',
```
iluwatar, line 6: Is this change intentional?

iluwatar (approved):
Added one question there, but otherwise I think it's good to be merged once the conflicts have been addressed.

src/proxy-ui-api/frontend/vue.config.js
teemukin65: No.. Needs to be fixed. Leaked due to local port mapping setup.

src/proxy-ui-api/frontend/tests/e2e/pages/ssMainPage.js
teemukin65: I have seen these single-item array mandatory even thought not required by the documentation. If left out the test cases fail due to not finding those commands.

src/proxy-ui-api/frontend/vue.config.js
teemukin65: fixed in : 1159ab7

PR #1158, opened 2022-01-13
Xrddev 1875

teemukin65:


iluwatar (changes requested):
To me, it seems that Spring Boot is still at the same version 2.5.8.
See https://github.com/nordic-institute/X-Road/blob/92655b9eb09e151719f33b739e04eef881e9dbd7/src/gradle.properties#L23

PR #1163, opened 2022-01-21
XRDDEV-1879 refactor CS login

iluwatar:


src/centralserver/admin-ui/src/views/AppLogin.vue
```diff
@@ -187,63 +186,45 @@ export default (
...
-        )
-        .catch((error) => {
+      try {
+        await this.login(loginData);
+        await this.fetchServerVersion();
+        await this.fetchSystemStatus();
+        await this.routeToMembersPage();
+      } catch (error) {
+        // Display invalid username/password error in inputs
+        if (error?.response?.status === 401) {
```
Riippi, line 196: I think that this "if 401 then show invalid username/password" check should be done only for the  this.login(loginData) call. 
Other api calls could be grouped under more general try catch with normal showError. For example if fetchServerVersion throws 401 error it is because of something else than wrong username/password.

src/centralserver/admin-ui/src/views/AppLogin.vue
```diff
@@ -187,63 +186,45 @@ export default (
...
+        } else {
           this.showError(error);
-          this.loading = false;
-        });
+        }
+        // Clear loading state
+        this.loading = false;
+      }
     },
     async requestUserData() {
```
Riippi, line 218: requestUserData is not called at all, but it should be, right after this.login().
Now after the login there is no permissions data etc. and the tabs are not visible.

src/centralserver/admin-ui/src/views/AppLogin.vue
```diff
@@ -187,63 +186,45 @@ export default (
...
       this.$refs.form.reset();
       this.loading = true;
       return this.fetchUserData();
     },
+    async routeToMembersPage() {
+      this.$router.replace({ name: RouteName.Members, })
```
Riippi, line 224: Comma is not needed at the end.

Riippi (changes requested):

src/centralserver/admin-ui/src/views/AppLogin.vue
iluwatar: Fixed in https://github.com/nordic-institute/X-Road/pull/1163/commits/8a6d39152883774f111a1b8b2b149fb5c9e0d4ba

src/centralserver/admin-ui/src/views/AppLogin.vue
iluwatar: Fixed in https://github.com/nordic-institute/X-Road/pull/1163/commits/8a6d39152883774f111a1b8b2b149fb5c9e0d4ba

src/centralserver/admin-ui/src/views/AppLogin.vue
iluwatar: Fixed in https://github.com/nordic-institute/X-Road/pull/1163/commits/8a6d39152883774f111a1b8b2b149fb5c9e0d4ba

PR #1178, opened 2022-02-04
XRDDEV-1921: Fix message log not working after upgrade (RHEL)

jhyoty:
* Fix symlink to config file not correctly created during upgrade
* Minor packaging fixes

### DIFF
--- a/src/packages/src/xroad/redhat/SPECS/common.inc
+++ b/src/packages/src/xroad/redhat/SPECS/common.inc
@@ -25,3 +25,7 @@ if [ "$1" -gt 1 ] ; then \
     fi \
 fi \
 %{nil}
+
+%define systemd_try_restart() \
+systemctl --quiet try-restart %{?*} >/dev/null 2>&1 || : \
+%{nil}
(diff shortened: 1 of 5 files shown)


src/packages/src/xroad/redhat/SPECS/common.inc
```diff
@@ -25,3 +25,7 @@ if [ "$1" -gt 1 ] ; then \
     fi \
 fi \
 %{nil}
+
+%define systemd_try_restart() \
+systemctl --quiet try-restart %{?*} >/dev/null 2>&1 || : \
+%{nil}
```
teemukin65, line 31: Ha!  this ':' is so clever shortcut for writing 'true' that I'm really impressed!  But for sake of the clarity, it could be more descriptive to use 'true', though.

teemukin65 (changes requested):

src/packages/src/xroad/redhat/SPECS/common.inc
jhyoty: Using `|| :` to ignore error return is a common (ba)sh idiom.

PR #1184, opened 2022-02-14
XRDDEV-1907 asicverifier fix

amoog:
Added option to override configuration expiry time, this also works when there is no metadata file.

### DIFF
--- a/src/common-util/src/main/java/ee/ria/xroad/common/conf/globalconf/ConfigurationDirectoryV2.java
+++ b/src/common-util/src/main/java/ee/ria/xroad/common/conf/globalconf/ConfigurationDirectoryV2.java
@@ -62,6 +62,9 @@ import static ee.ria.xroad.common.conf.globalconf.ConfigurationUtils.escapeInsta
  */
 @Slf4j
 public class ConfigurationDirectoryV2 implements ConfigurationDirectory {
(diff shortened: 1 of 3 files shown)


src/common-util/src/main/java/ee/ria/xroad/common/conf/globalconf/ConfigurationDirectoryV2.java
```diff
@@ -62,6 +62,9 @@
  */
 @Slf4j
 public class ConfigurationDirectoryV2 implements ConfigurationDirectory {
+    // overrides expires field from metadata
+    // currently used by asicverifier
+    public static OffsetDateTime expiresOverride;
```
jhyoty, line 67: A global flag feels like a code smell here. Perhaps it should be an option to GlobalConf.reload() to ignore expiration. Or restore the old behavior that ignores expiration if the metadata does not exists.

jhyoty (changes requested):
Some other issues:
ConfigurationDirectoryV2: 
- on line 76, expirationDate seems unused
- 269: isInThePast() is unused
- path, instanceIdentifier, and the parameter maps could be final
Potential data race:
- load*Parameters modify the parameters' expiration time when reusing existing parameters.

PR #1186, opened 2022-02-15
Xrddev 1908 refactor service clients

turkkaholmqvist:
Modified two test suites to match structure we are using everywhere else in project. Tests are still same, syntax is to:
- some parts have been moved to setup-blocks for repeatability
- page objects are set in before block to remove duplicate code
- some single use functions are being moved as part of testcase to remove branching and unnecessary complexion


jansu76:
LGTM

jansu76:
build failure is unrelated to changes

PR #1200, opened 2022-03-16
Xrddev 1947 refactor service tests 

turkkaholmqvist:
Refactored service tests to support after and before -blocks, also improved tests slightly by making couple page object stored reusable functions.


src/proxy-ui-api/frontend/tests/e2e/specs/ss-clients-wsdlservices.js
```diff
@@ -24,55 +24,72 @@
  * THE SOFTWARE.
  */
 
+let frontPage, mainPage, clientsTab, clientInfo, clientServices, operationDetails, sslCheckFail;
+let addSubjectsPopup, removeAccessRightPopup, removeAllAccessRightsPopup, serviceChangePopup, serviceDetails,servicesPopup ;
+
 module.exports = {
-  tags: ['ss', 'clients', 'wsdlservices'],
+  tags: ['ss', 'clients', 'wsdlservices', 'WIP'],
```
jansu76, line 31: Is this tag here intentionally?

src/proxy-ui-api/frontend/tests/e2e/pages/tabs/clientsTabClientInfo.js
```diff
@@ -152,6 +152,18 @@ var clientLocalGroupsCommands = {
...
 var clientServicesCommands = {
+  groupMemberTableContains: function (expectedText) {
+    this.api.waitForElementVisible(`//table[contains(@class, "group-members-table")]//td[contains(text(), "${expectedText}")]`);
+    return this;
+  },
+  errorMessageIsShown: function (expectedMessage) {
+    // Verifies that error message element is shown, if argument is given, function polls that text from page.
+    this.api.waitForElementVisible('//div[contains(@class, "v-messages__message")]');
+    if (expectedMessage) {
+      this.api.waitForElementVisible(`//*[text()[contains(.,"${expectedMessage}")]]`)
```
jansu76, line 163: Does this search for expectedMessage anywhere on the page? If yes, there's a risk of incorrect result if someone will use using this to find a simple text that is found somewhere else on the page but not on the error message. Would be better to scoped this to search only inside the error message div, perhaps something similar to groupMemberTableContains (untested code)

```
this.api.waitForElementVisible(`//div[contains(@class, "v-messages__message")]//*[text()[contains(.,"${expectedMessage}")]]`);
```

jansu76 (changes requested):

src/proxy-ui-api/frontend/tests/e2e/pages/tabs/clientsTabClientInfo.js
turkkaholmqvist: Yes, that is true that it can find similair string anywhere from page if it's visible, but as long as string is long enough it's very unlikely that it will fail on that. 

It's also true that this string locator you provided would be better solution, I intended originally do it like that, but after intensive trial and error, i wasn't able to get it work - I was trying to implement some AND-operations that both of xpaths are being solved at same time, while it can probably work like as your example, eg. that second xpath is considered as  "any sibling of first xpath", which is much clearer syntax wise as well.

jansu76: "...as long as string is long enough it's very unlikely" is true but at the same time kind of a trap, that usually causes grief when someone else inevitably misuses this some months / years later...

```
'//div[contains(@class, "v-messages__message") and contains(text(), "URL")]'
```

seems to work

src/proxy-ui-api/frontend/tests/e2e/specs/ss-clients-wsdlservices.js
turkkaholmqvist: No, it's to run only that test i am working on, I will remove it.

src/proxy-ui-api/frontend/tests/e2e/pages/tabs/clientsTabClientInfo.js
turkkaholmqvist: fixed this according example, seems to work now.

src/proxy-ui-api/frontend/tests/e2e/specs/ss-clients-wsdlservices.js
jansu76: Seems to still be in the PR?

jansu76 (dismissed):

PR #1233, opened 2022-05-02
XRDDEV-1994 Configuration proxy shall always clean up the temporary f…

andresrosenthal:
…iles, even when a error occurs during the process


src/configuration-proxy/src/main/java/ee/ria/xroad/confproxy/util/OutputBuilder.java
```diff
@@ -150,14 +151,22 @@ public final void moveAndCleanup() throws Exception {
         log.debug("Moving '{}' to '{}'", tempConfPath, targetConf);
 
         Files.move(tempConfPath, targetConf, StandardCopyOption.ATOMIC_MOVE);
+    }
+
+    /**
+     * Cleans up any remaining temporary files.
+     * @throws IOException in case of unsuccessful file operations
+     */
+    public final void cleanup() throws IOException {
```
ricardas-buc, line 160: How about using AutoCloseable interface? try-with-resources pattern would suite this usecase well

ricardas-buc (changes requested):

ricardas-buc (approved):

PR #1264, opened 2022-06-06
XRDDEV-2025 Add members without configured encryption keys to Diagnos…

enelir:
…tic page Message log archive encryption table

### DIFF
--- a/src/proxy/src/main/java/ee/ria/xroad/proxy/ProxyMain.java
+++ b/src/proxy/src/main/java/ee/ria/xroad/proxy/ProxyMain.java
@@ -40,7 +40,6 @@ import ee.ria.xroad.common.Version;
 import ee.ria.xroad.common.conf.globalconf.GlobalConfUpdater;
 import ee.ria.xroad.common.conf.serverconf.CachingServerConfImpl;
 import ee.ria.xroad.common.conf.serverconf.ServerConf;
-import ee.ria.xroad.common.conf.serverconf.model.ClientType;
 import ee.ria.xroad.common.identifier.ClientId;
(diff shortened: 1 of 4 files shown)


andresrosenthal:
[rerun-ci]

ricardas-buc (approved):

petkivim:
This issue is scoped to version `7.1.0` so it should be merged into `beta-7.1.0` branch.

petkivim:
The icon that indicates that the default encryption key is used should be gray when the element is disabled (=message log add-on is disabled):

Screenshot 2022-06-08 at 7 05 25

Just like the encryption status icon is grey in the screenshot.

src/proxy/src/main/java/ee/ria/xroad/proxy/ProxyMain.java
```diff
@@ -238,7 +240,23 @@ private static void createServices() throws Exception {
...
+            return ServerConf.getMembers().stream()
+                    .filter(ProxyMain::isMemberRegistered)
+                    .collect(Collectors.toList());
+        } catch (Exception e) {
+            log.error("Failed to get members from server configuration", e);
+            throw e;
+        }
+    }
+
+    private static boolean isMemberRegistered(ClientId member) {
```
petkivim, line 257: This filter can be removed. The list should include all local clients regardless of their status. There's no need to restrict the list to registered clients only.

enelir:
JIRA: https://jira.niis.org/browse/XRDDEV-2025

petkivim (approved):
Looks good to me.

PR #1269, opened 2022-06-08
Remove "buildAll" gradle task (from develop)

andresrosenthal:


src/build.gradle
```diff
@@ -17,6 +17,7 @@ plugins {
     id 'idea'
     id "io.spring.dependency-management" version "1.0.11.RELEASE" apply false
     id 'org.springframework.boot' version "${springBootVersion}" apply false
+    id "com.dorongold.task-tree" version "2.1.0"
```
ricardas-buc, line 20: Is this for visualization or it has some other use?

andresrosenthal: Correct, just for visualization

ricardas-buc (approved):

monikalnortal (approved):

PR #1276, opened 2022-06-15
XRDDEV-2031: Fix test case, failing for mismatch in time assertion

monikalnortal:
Test fix for a story https://jira.niis.org/browse/XRDDEV-2031


enelir (commented):
okey

PR #1279, opened 2022-06-16
XRDDEV-2059 Add global group list API service

enelir:
JIRA: https://jira.niis.org/browse/XRDDEV-2059


src/centralserver/admin-service/src/main/java/org/niis/xroad/centralserver/restapi/openapi/GlobalGroupsApiController.java
```diff
@@ -0,0 +1,92 @@
...
+    }
+
+    @Override
+    public ResponseEntity<Void> deleteGlobalGroupMembers(String groupId, Members members) {
+        throw new NotImplementedException("deleteGlobalGroupMembers not implemented yet");
+    }
+
+    @Override
+    @PreAuthorize("hasAuthority('VIEW_GLOBAL_GROUPS')")
+    public ResponseEntity<Set<GlobalGroup>> findGlobalGroups(String containsMember) {
```
enelir, line 74: In openApi definition was optional String containsMember, but in figma UI i don't see that this is necessary. Should we this implement now (this jira ticket) or we should implement this when we really need this?

enelir: containsMember filter was implemented in this ticket

PR #1281, opened 2022-06-17
XRDDEV-1842 Update the External Load Balancer Installation Guide

petkivim:
Replace the word "replica" with "secondary" in the External Load Balancer Installation Guide.

JIRA: https://jira.niis.org/browse/XRDDEV-1842


raits (approved):
Looks good to me.

PR #1318, opened 2022-08-18
XRDDEV-2112 Retain endpoints and ACLs when removing service description

mloitm:
When service description is removed, retain endpoints and ACLs if those
are referenced by other service versions


andresrosenthal:
[rerun-ci]

andresrosenthal (approved):

PR #1339, opened 2022-09-06
XRDDEV-1935 Make "add member" case independent from a existing memberclass

andresrosenthal:


src/e2e-tests/src/intTest/resources/behavior/ui/csAddMember.feature
```diff
@@ -8,6 +8,9 @@ Feature: Central Server Add Member
...
 
   Scenario: Member is created and listed
-    Given Members tab is selected
-    When A new member is added
+    Given CentralServer Settings tab is selected
+    And System settings tab is selected
+    And A new memberclass XXX is added
+    And Members tab is selected
+    When A new member with XXX memberclass is added
     Then A new member is listed
```
ricardas-buc, line 16: "is listed" checks for count, this can get flaky if someone else adds new member during test execution

src/e2e-tests/src/intTest/resources/behavior/ui/csAddMember.feature
```diff
@@ -8,6 +8,9 @@ Feature: Central Server Add Member
     And User xrd logs in to CentralServer with password secret
 
   Scenario: Member is created and listed
-    Given Members tab is selected
-    When A new member is added
+    Given CentralServer Settings tab is selected
+    And System settings tab is selected
+    And A new memberclass XXX is added
```
ricardas-buc, line 13: XXX is bit confusing and will be hard to search. Lets make it clear that this came from test automation.

We've also discussed making it dynamic, but we can leave it as is for now.

src/e2e-tests/src/intTest/resources/behavior/ui/csAddMember.feature
andresrosenthal: Would adding with e2e-specific code, class & name values and later checking if that combination exists in the table be sufficient?

ricardas-buc: I would think so, we cannot add duplicates right?

andresrosenthal: class-code combination has to be unique yes

src/e2e-tests/src/intTest/java/org/niis/xroad/test/ui/glue/centralserver/CentralServerMemberStepDefs.java
```diff
@@ -52,25 +50,26 @@ public void userNavigatesToMembersTab() {
...
+        scenarioContext.putStepData("memberClass", memberClass);
 
         $(BTN_ADD_MEMBER).click();
-        $(INPUT_MEMBER_NAME).setValue("New test member");
+        $(INPUT_MEMBER_NAME).setValue("E2E Test Member");
         $(SELECT_MEMBER_CLASS).click();
-        getOption("ORG").click();
-        $(INPUT_MEMBER_CODE).setValue("1234567-8");
+        getOption(memberClass).click();
+        $(INPUT_MEMBER_CODE).setValue("e2e-test-member");
```
ricardas-buc, line 61: If we would add this one as step arg it would be highly reusable step.

ricardas-buc (approved):

PR #1343, opened 2022-09-11
XRDDEV-2069 Add Gloubal groups member search view

enelir:
JIRA: https://nordic-institute.atlassian.net/browse/XRDDEV-2069

### DIFF
--- a/src/centralserver/admin-service/src/main/java/org/niis/xroad/centralserver/restapi/repository/GlobalGroupMemberRepository.java
+++ b/src/centralserver/admin-service/src/main/java/org/niis/xroad/centralserver/restapi/repository/GlobalGroupMemberRepository.java
@@ -40,7 +40,6 @@ import org.niis.xroad.centralserver.restapi.entity.XRoadId_;
 import org.springframework.data.jpa.domain.Specification;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.jpa.repository.JpaSpecificationExecutor;
-import org.springframework.data.jpa.repository.Query;
 import org.springframework.stereotype.Repository;
 import org.springframework.util.CollectionUtils;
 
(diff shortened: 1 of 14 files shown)


src/centralserver/admin-service/src/main/java/org/niis/xroad/centralserver/restapi/repository/GlobalGroupMemberRepository.java
```diff
@@ -0,0 +1,117 @@
...
+        return builder.or(
+                caseInsensitiveLike(root, builder, criteria.getQuery(), member.get(XRoadId_.MEMBER_CODE)),
+                caseInsensitiveLike(root, builder, criteria.getQuery(), member.get(XRoadId_.MEMBER_CLASS)),
+                caseInsensitiveLike(root, builder, criteria.getQuery(), member.get(XRoadId_.SUBSYSTEM_CODE)),
+                caseInsensitiveLike(root, builder, criteria.getQuery(), member.get(XRoadId_.X_ROAD_INSTANCE)),
+                caseInsensitiveLike(root, builder, criteria.getQuery(), member.get(XRoadId_.OBJECT_TYPE))
+        );
+    }
+
+    @Query("SELECT m FROM #{#entityName} m JOIN m.globalGroup g "
```
mloitm, line 114: if you would rename this method to `findByGlobalGroupId`,  then it doesn't need `@Query` annotation.

Also this query could be simplified like this:
```sql 
"SELECT m FROM GlobalGroupMember m WHERE m.globalGroup.id = :groupId"
```
Use `#{#entityName}` only if you need dynamic entity name (e.g. in abstract base class). Otherwise you loose IDE support for autocomplete and syntax checking/highlighting.

mloitm (approved):

PR #1344, opened 2022-09-14
Xrddev 2130

turkkaholmqvist:
I am trying to recreate tests from Nightwatch to new stack and here are results regarding login and logout suites.

Most obvious issues:
-  I made steps to file `CommonUIStepDefs.java`, I see that naming scheme refers it as "commonUI", which in my understanding refers to pages before login. Now issue is that in logout case, functionalities refering to elements after login (namely session timeout popup and logout-button) are also inside of CommonUIsteps, which is bit incorrrect location-wise, although they seem to work.
- I have understanding that ```@Given("Something is done)```  in step definitions should match usecase in feature-files, eg. if we use it as "Given" then it's created with "Given" step definitions. However, this doesn't seem work, if I use "Then" tests fail on error. 
- Issue is that when you run undefined steps, error log is kind enough to output you basic code create snippet with exception and they really generated with "Then" and "When" respectively, it which doesn't work in my tests - hence i might have missed something semantics of how these steps are supposed to work.
- There is also artificial wait period created to account for inactivity in session timeout -case, it might be better to come up with dynamic wait so there wouldn't be such overhead and safety margin that is taken everytime when test is run. I am not completely sure how to make that syntax wise.
- Might also be worthwhile to explore possibility to make common browser-session for all tests, since now browser is closed and reopened for each case generating large overhead timewise where we just wait for browser to open. In permission tests, this same issue is even more prominent since there is login sequence before each test as well.

### DIFF
--- a/src/e2e-tests/src/intTest/java/org/niis/xroad/test/ui/glue/CommonUiStepDefs.java
+++ b/src/e2e-tests/src/intTest/java/org/niis/xroad/test/ui/glue/CommonUiStepDefs.java
@@ -30,11 +30,13 @@ import com.codeborne.selenide.Selenide;
 import io.cucumber.java.en.Given;
 import org.openqa.selenium.By;
 
+import java.time.Duration;
+
 import static com.codeborne.selenide.Selenide.$;
 
 public class CommonUiStepDefs extends BaseUiStepDefs {
     public static final By BTN_CLOSE_SNACKBAR = By.xpath("//button[@data-test=\"close-snackbar\"]");
-    public static final int MILLISECONDS_IN_SECOND = 1000;
+    public static final int MAX_WAIT_IN_SECONDS = 120;
 
     @Given("logout button is being clicked")
     public void logoutButtonIsClicked() {
@@ -44,18 +46,16 @@ public class CommonUiStepDefs extends BaseUiStepDefs {
                 .click();
     }
 
-    @Given("{int} seconds of inactivity is passed")
-    public void secondsOfInactivityIsPassed(Integer int1) {
-
-        Selenide.sleep(int1 * MILLISECONDS_IN_SECOND);
-
+    @Given("User becomes idle")
+    public void userBecomesIdle() {
+        Selenide.sleep(1);
     }
 
 
-    @Given("Error message about timeout appears")
+    @Given("after 120 seconds, session timeout popup appears")
     public void errorMessageAboutTimeoutAppears() {
         $(By.xpath("//button[@data-test='session-expired-ok-button']"))
-                .shouldBe(Condition.visible);
+                .shouldBe(Condition.visible, Duration.ofSeconds(MAX_WAIT_IN_SECONDS));
 
     }
 
--- a/src/e2e-tests/src/intTest/java/org/niis/xroad/test/ui/glue/CommonUiStepDefs.java
+++ b/src/e2e-tests/src/intTest/java/org/niis/xroad/test/ui/glue/CommonUiStepDefs.java
@@ -61,6 +61,46 @@ public class CommonUiStepDefs extends BaseUiStepDefs {
                 .click();
     }
 
+    @Given("Error message for incorrect credentials is shown")
+    public void errorMessageIsShown() {
+
+        $(By.xpath("//div[text()[contains(.,'Wrong username or password')]]"))
+                .shouldBe(Condition.visible, Duration.ofSeconds(20));
+    }
+
+     @Given("logout button is being clicked")
+        public void logoutButtonIsClicked() {
+         $(By.xpath("//button[@data-test='username-button']"))
+                        .shouldBe(Condition.visible)
+                        .shouldBe(Condition.enabled)
+                        .click();
+        }
+
+     @Given("{int} seconds of inactivity is passed")
+        public void seconds_of_inactivity_is_passed(Integer int1) {
+        try
+        {
+            Thread.sleep(int1 * 1000);
+        }
+        catch(InterruptedException e) {}
+        }
+
+
+    @Given("Error message about timeout appears")
+    public void error_message_about_timeout_appears() {
+        $(By.xpath("//button[@data-test='session-expired-ok-button']"))
+            .shouldBe(Condition.visible);
+
+    }
+
+     @Given("OK is clicked on timeout notification popup")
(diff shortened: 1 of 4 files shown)


src/e2e-tests/src/intTest/java/org/niis/xroad/test/ui/glue/CommonUiStepDefs.java
```diff
@@ -61,6 +61,46 @@ public void doLogin(final String username, final String password) {
...
+                        .shouldBe(Condition.visible)
+                        .shouldBe(Condition.enabled)
+                        .click();
+        }
+
+     @Given("{int} seconds of inactivity is passed")
+        public void seconds_of_inactivity_is_passed(Integer int1) {
+        try
+        {
+            Thread.sleep(int1 * 1000);
```
ricardas-buc, line 83: Use Selenide.sleep instead.
https://selenide.org/javadoc/current/com/codeborne/selenide/Selenide.html#sleep(long)

ricardas-buc: Or maybe wait for x seconds for login screen to appear?

ricardas-buc:
> I am trying to recreate tests from Nightwatch to new stack and here are results regarding login and logout suites.
> 
> Most obvious issues:
> 
> * I made steps to file `CommonUIStepDefs.java`, I see that naming scheme refers it as "commonUI", which in my understanding refers to pages before login. Now issue is that in logout case, functionalities refering to elements after login (namely session timeout popup and logout-button) are also inside of CommonUIsteps, which is bit incorrrect location-wise, although they seem to work.

CommonUIStepDefs should hold things which are common between cs,ss and different pages. A good example would be tooltips, error messages, common form elements and so on. In case of Login/logout I would create a separate definition file. 

> * I have understanding that `@Given("Something is done)`  in step definitions should match usecase in feature-files, eg. if we use it as "Given" then it's created with "Given" step definitions. However, this doesn't seem work, if I use "Then" tests fail on error.

Technically there is no difference which annotation in stepDef is used. It is only important that feature file has correct given/then/when usage.

> * Issue is that when you run undefined steps, error log is kind enough to output you basic code create snippet with exception and they really generated with "Then" and "When" respectively, it which doesn't work in my tests - hence i might have missed something semantics of how these steps are supposed to work.

Might be related to java imports. Which IDE are you using?

> * There is also artificial wait period created to account for inactivity in session timeout -case, it might be better to come up with dynamic wait so there wouldn't be such overhead and safety margin that is taken everytime when test is run. I am not completely sure how to make that syntax wise.

I've left a comment on that particular place with few suggestions. Ideally we should not be using sleeps at all, but waiting on some elements to show up.

> * Might also be worthwhile to explore possibility to make common browser-session for all tests, since now browser is closed and reopened for each case generating large overhead timewise where we just wait for browser to open. In permission tests, this same issue is even more prominent since there is login sequence before each test as well.

Yes, there are ways to improve execution time. We will improve this once we get the time to do it.

Additional note - checkstyle is failing, please run gradle build on e2e module locally, it will show you the errors.
` ./gradlew :e2e-tests:build   `

ricardas-buc (changes requested):

src/e2e-tests/src/intTest/java/org/niis/xroad/test/ui/glue/CommonUiStepDefs.java
turkkaholmqvist: I will try and go with this idea of "waiting until login screen appears", then it won't be sleeping anymore than it needs to. I need to figure out how it would make most sense Then/When -wise.

turkkaholmqvist:
> Might be related to java imports. Which IDE are you using?

I am using Webstorm, it's intelliJ -based product but to cust costs, i wasn't given full license since automation here was based on javascript.

>  In case of Login/logout I would create a separate definition file.

Sounds reasonable, will try to do that, will create something like `frontPageStepDef`

ricardas-buc:
> 



> > Might be related to java imports. Which IDE are you using?
> 
> I am using Webstorm, it's intelliJ -based product but to cust costs, i wasn't given full license since automation here was based on javascript.
> 
> > In case of Login/logout I would create a separate definition file.
> 
> Sounds reasonable, will try to do that, will create something like `frontPageStepDef`

Not sure how webstorm handles java language, but you can always go with intellij comunity edition for tests. It will cover all of your needs.

src/e2e-tests/src/intTest/java/org/niis/xroad/test/ui/glue/CommonUiStepDefs.java
turkkaholmqvist: Switched this to Selenide.sleep for now - waiting dynamically is bit cumbersome since i don't have way to integrate expected element to this particular step and have proper When/Then structure.

ricardas-buc: Could you elaborate more what way did not work for you? You can provide non-working test snippet, just to see what did not work

turkkaholmqvist: Issue is in this `When/Then` -structure, in case i would like to make reusable WHEN `XX seconds of inactivity is passed`, THEN `session timeout popup is appears`. 

Issue is that to separate waiting and what we are waiting, they are in different functions. It would be simple to make something like `when XX seconds of inactivity is passed, session timeout popup appears` but it seems bit bad practice to make such single use function that explicitly wait for some specific elemen  and i don't really know what i should put in that `Then` section if i would do such way.

ricardas-buc: It is completely fine and not a bad practice to make step like this:

When User becomes idle
Then After XX seconds of inactivity,  session timeout popup appears


"User becomes idle' - you can leave step definition empty, or just add Selenide.sleep(1). This is mostly for readability.

"After XX seconds of inactivity,  session timeout popup appears" - all the logic goes here.

Even though reusable step definitions are great, the intent is to have readable scenarios. Thats why making few specialised steps is ok.

PR #1349, opened 2022-09-21
XRDDEV-2084 delete subsystem api

justasnortal:
XRDDEV-2084 delete subsystem api


justasnortal:
https://nordic-institute.atlassian.net/browse/XRDDEV-2084

src/centralserver/admin-service/src/main/java/org/niis/xroad/centralserver/restapi/service/SubsystemService.java
```diff
@@ -61,4 +66,19 @@ private void saveSecurityServerClientName(Subsystem subsystem) {
...
+
+        if (isRegistered(subsystem)) {
+            throw new ValidationFailureException(ErrorMessage.SUBSYSTEM_REGISTERED_AND_CANNOT_BE_DELETED);
+        }
+
+        subsystemRepository.delete(subsystem);
+    }
+
+    private boolean isRegistered(Subsystem subsystem) {
+        return !subsystem.getServerClients().isEmpty();
```
andresrosenthal, line 81: Can `subsystem.getServerClients()` never be null?
If it can then would be good a idea to replace with: 
```suggestion
        return !CollectionUtils.isEmpty(subsystem.getServerClients())
```

andresrosenthal (approved):

PR #1393, opened 2022-10-17
XRDDEV-2145 API to get the certificate of a certification service

justasnortal:
API to get the certificate of a certification service.

https://nordic-institute.atlassian.net/browse/XRDDEV-2145


src/centralserver/admin-service/infra-api-rest/src/main/java/org/niis/xroad/centralserver/restapi/converter/CertificateDetailsDtoConverter.java
```diff
@@ -0,0 +1,39 @@
...
+ * THE SOFTWARE.
+ */
+
+package org.niis.xroad.centralserver.restapi.converter;
+
+import org.mapstruct.Mapper;
+import org.niis.xroad.centralserver.openapi.model.CertificateDetailsDto;
+import org.niis.xroad.centralserver.restapi.dto.CertificateDetails;
+
+@Mapper(componentModel = "spring")
```
ricardas-buc, line 34: we should consider upgrading to mapstruct 1.5.x, then we could do
@Mapper(componentModel = MappingConstants.ComponentModel.SPRING)

src/centralserver/admin-service/core/src/main/java/org/niis/xroad/centralserver/restapi/dto/converter/CaInfoConverter.java
```diff
@@ -0,0 +1,86 @@
...
+import static java.lang.String.valueOf;
+
+@Component
+@RequiredArgsConstructor
+public class CaInfoConverter {
+
+    private final KeyUsageConverter keyUsageConverter;
+
+    @SneakyThrows
+    public CertificateDetails convert(CaInfo caInfo) {
```
ricardas-buc, line 55: was use of mapstruct considered here?  I do see pros and cons of using it so curious regarding this decision

ricardas-buc (approved):

src/centralserver/admin-service/core/src/main/java/org/niis/xroad/centralserver/restapi/dto/converter/CaInfoConverter.java
justasnortal: Yes, it was. But when mapping from entity to dto (dto fields naming is based on openapi dto), many fields would be mapped using annotations. The amount of code lines would be similar, but it is more error prone.

src/centralserver/admin-service/infra-api-rest/src/main/java/org/niis/xroad/centralserver/restapi/converter/CertificateDetailsDtoConverter.java
justasnortal: Updated

justasnortal:
[rerun-ci]

PR #1416, opened 2022-11-07
XRDDEV-2164 complementations

andresrosenthal:


src/centralserver/admin-ui/src/locales/en.json
```diff
@@ -118,6 +118,9 @@
       "button": "Upload backup"
     }
   },
+  "customValidation": {
+    "invalidUrl": "URL is not valid, examples of valid URL-s: 'http://www.example.com', 'https://www.example.com'"
```
VPaliliunas, line 122: **_"URL-s"_** sounds incorrect, maybe use **_"examples of valid URL formats"_** or something similar
Also **_is not valid_** should just be **_is invalid_**

PR #1421, opened 2022-11-09
Jackson vulnerability fix in test-automation libraries

andresrosenthal:


src/centralserver/admin-service/application/build.gradle
```diff
@@ -23,6 +23,11 @@ dependencies {
     implementation('org.liquibase:liquibase-core') //TODO enabling runtime liquibase
 
     intTestImplementation("com.nortal.test:test-automation-core:${testAutomationFrameworkVersion}")
+    constraints {
```
ricardas-buc, line 26: can we consider adding this to root build.gradle (under src) and using version range instead of contant override? In theory this vulnerability can come from any dependency, no need to concentrate on test automation

PR #1443, opened 2022-11-28
Draft: xrddev-2126

ricardas-buc:


src/centralserver/admin-service/test-ui/src/intTest/java/org/niis/xroad/cs/test/ui/glue/CommonUiStepDefs.java
```diff
@@ -0,0 +1,81 @@
...
+import java.time.Duration;
+
+import static com.codeborne.selenide.Selenide.$;
+
+public class CommonUiStepDefs extends BaseUiStepDefs {
+    public static final int MAX_WAIT_IN_SECONDS = 120;
+
+    @Given("logout button is being clicked")
+    public void logoutButtonIsClicked() {
+        $(By.xpath("//button[@data-test='username-button']"))
```
mloitm, line 42: Maybe move selectors to pageObject?

ricardas-buc: Yep, seems like I've missed this class. will fix

justasnortal (approved):

PR #1456, opened 2022-12-15
XRDDEV-2181 Timestamping service E2E tests

enelir:
JIRA: https://nordic-institute.atlassian.net/browse/XRDDEV-2181

### DIFF
--- a/src/centralserver/admin-service/ui-system-test/src/intTest/java/org/niis/xroad/cs/test/ui/glue/CommonUiStepDefs.java
+++ b/src/centralserver/admin-service/ui-system-test/src/intTest/java/org/niis/xroad/cs/test/ui/glue/CommonUiStepDefs.java
@@ -27,6 +27,12 @@ package org.niis.xroad.cs.test.ui.glue;
 
 import com.codeborne.selenide.Selenide;
 import io.cucumber.java.en.Step;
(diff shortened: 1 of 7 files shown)


src/centralserver/admin-service/ui-system-test/src/intTest/java/org/niis/xroad/cs/test/ui/glue/CommonUiStepDefs.java
```diff
@@ -42,4 +48,20 @@ public void preparePage() {
                 + "      document.head.insertAdjacentHTML('beforeend', style);");
     }
 
+    @SuppressWarnings("checkstyle:MagicNumber")
+    @Step("Page is prepared to be loading tested")
+    public void loadingTesting() {
+
+        var chromedriver = (ChromeDriver) Selenide.webdriver().driver().getWebDriver();
```
ricardas-buc, line 55: I'd propose to add a "load-test flag here. And then create before and after hooks that will set 2g value and restore it to default after the test.

Also, I would phrase it "Browser is set to {string} network speeds"

ricardas-buc (changes requested):

ricardas-buc (approved):

PR #1460, opened 2022-12-19
XRDDEV-2195 Token login/logout integration tests

justasnortal:


src/centralserver/admin-service/core/src/main/java/org/niis/xroad/centralserver/restapi/facade/SignerProxyFacadeHttpImpl.java
```diff
@@ -0,0 +1,249 @@
...
+
+import java.util.Date;
+import java.util.List;
+import java.util.Map;
+
+@Profile("int-test")
+@ConditionalOnProperty(value = "signerProxyMockUri")
+@Component
+@Slf4j
+public class SignerProxyFacadeHttpImpl implements SignerProxyFacade {
```
ricardas-buc, line 65: Incosistent naming as it looks like non mockable impl, but in fact uses signerProxyMockUri

ricardas-buc (changes requested):

src/centralserver/admin-service/core/src/main/java/org/niis/xroad/centralserver/restapi/facade/SignerProxyFacadeHttpImpl.java
justasnortal: Renamed.

ricardas-buc (approved):

PR #1463, opened 2022-12-21
XRDDEV-2197 API endpoint for removing configuration signing keys

andresrosenthal:
https://nordic-institute.atlassian.net/browse/XRDDEV-2197


src/centralserver/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/entity/ConfigurationSigningKeyEntity.java
```diff
@@ -85,6 +85,10 @@ protected ConfigurationSigningKeyEntity() {
         //JPA
     }
 
+    public ConfigurationSigningKeyEntity(int id) {
```
justasnortal, line 88: Not sure if it is a good idea. We are using sequence for id. With this constructor the id must be provided to create a new entity.

src/centralserver/admin-service/core/src/test/java/org/niis/xroad/centralserver/restapi/service/ConfigurationSigningKeysServiceImplTest.java
```diff
@@ -35,49 +35,133 @@
...
+
     private final ConfigurationSigningKeyMapper configurationSigningKeyMapper = new ConfigurationSigningKeyMapperImpl();
 
     private ConfigurationSigningKeysServiceImpl configurationSigningKeysServiceImpl;
 
     @BeforeEach
     public void setup() {
-        configurationSigningKeysServiceImpl =
-                new ConfigurationSigningKeysServiceImpl(configurationSigningKeyRepository, configurationSigningKeyMapper);
+        configurationSigningKeysServiceImpl = new ConfigurationSigningKeysServiceImpl(configurationSigningKeyRepository,
```
justasnortal, line 81: Why recreate it before each test? configurationSigningKeysServiceImpl can be instantiated on class level.

andresrosenthal: Did not use class level initialization here, because it only works with `@InjectMocks`, which in turn does not work with `configurationSigningKeyMapper` regular object.
Come to think of it, a workaround would be to annotate `configurationSigningKeyMapper` with `@Spy`

src/centralserver/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/entity/ConfigurationSigningKeyEntity.java
andresrosenthal: Yeah, I needed this to verify wether the private key id is passed on when transforming from `ConfigurationSigningKeyEntity` to `ConfigurationSigningKey`. But it seems this field is not really needed as we use the String `keyIdentifier` instead, therefore removed the `id` from `ConfigurationSigningKey`

justasnortal (approved):

PR #1501, opened 2023-01-18
feat: XRDDEV-2211 internal conf generation

mloitm:


andresrosenthal (approved):

src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/entity/CentralServiceEntity.java
```diff
@@ -64,10 +64,9 @@ public class CentralServiceEntity extends AuditableEntity {
     @Setter
     private String serviceCode;
 
-    protected CentralServiceEntity() {
+    public CentralServiceEntity() {
```
andresrosenthal, line 67: Perhaps `@NoArgsConstructor` instead for consistency

andresrosenthal (approved):

PR #1503, opened 2023-01-18
feat: Git commit message validation

ricardas-buc:
This is initial version of git hook which will validate against https://www.conventionalcommits.org/en/v1.0.0/#summary

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -206,7 +206,7 @@ The latest development version is always available in the `develop` branch and t
 
 Pull requests made against the `X-Road/develop` branch MUST follow these conventions:
 
-- Pull request name format is `<TYPE>: <ISSUE_ID> <SHORT_DESCRIPTION>`, for example: `XRDDEV-1669 Allow overriding startup parameters`
+- Pull request name format is `<TYPE>: <ISSUE_ID> <SHORT_DESCRIPTION>`, for example: `feat: XRDDEV-1669 Allow overriding startup parameters`
   - `TYPE` = conventional commit type. One of: build|chore|ci|docs|feat|fix|perf|refactor|revert|style|test
   - `ISSUE_ID` = id of the feature's / bug's backlog item. If the pull request is not related to any backlog item,
     `ISSUE_ID` can be omitted. If the issue is a GitHub issue, it can be `ISSUE1669 Allow overriding startup parameters`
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -215,21 +215,34 @@ Pull requests made against the `X-Road/develop` branch MUST follow these convent
...
+<type>[optional scope]: <subject>
+
+[optional body]
+
+[optional footer(s)]
+```
+
+message should be formatted below:
+
+- Do not end the line with a period.
```
mloitm, line 233: Maybe `Do not end the subject line with a period.`?
It's ok to use period in body (like there's in the example below).

CONTRIBUTING.md
```diff
@@ -215,21 +215,34 @@ Pull requests made against the `X-Road/develop` branch MUST follow these convent
...
+- Capitalize the line and each paragraph.
 - Use the imperative mood in the subject line.
 - Wrap lines at 72 characters.
```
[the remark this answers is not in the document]
ricardas-buc: I guess it can stay as a suggestion? regex will not validate it, but usually if the message is long - use the body.

CONTRIBUTING.md
```diff
@@ -206,7 +206,8 @@ The latest development version is always available in the `develop` branch and t
 
 Pull requests made against the `X-Road/develop` branch MUST follow these conventions:
 
-- Pull request name format is `<ISSUE_ID> <SHORT_DESCRIPTION>`, for example: `XRDDEV-1669 Allow overriding startup parameters`
+- Pull request name format is `<TYPE>: <ISSUE_ID> <SHORT_DESCRIPTION>`, for example: `XRDDEV-1669 Allow overriding startup parameters`
```
petkivim, line 209: It would be good to include the type (`TYPE`) in the example too.

CONTRIBUTING.md
```diff
@@ -224,12 +237,14 @@ Git commit messages SHOULD follow the guidelines described below:
 - Wrap lines at 72 characters.
 - Use the body to explain what and why you have done something. In most cases, you can leave out details about how a change has been implemented.
 
-If a commit refers to an issue, add the issue ID before the subject. For example:
+If a commit refers to an issue, add the issue ID should be referenced. For example:
```
petkivim, line 240: Consider changing this:

> If a commit refers to an issue, add the issue ID should be referenced.

to:

> If a commit refers to an issue, the issue ID must be referenced.

mloitm (approved):

PR #1513, opened 2023-02-02
XRDDEV-2225 API for uploading and downloading configuration parts

justasnortal:
https://nordic-institute.atlassian.net/browse/XRDDEV-2225


src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/service/ConfigurationServiceImpl.java
```diff
@@ -251,6 +331,49 @@ public Set<DistributedFile> getAllConfigurationFiles(int version) {
                 .collect(toSet());
     }
 
+
+    @Override
+    public void uploadConfigurationPart(ConfigurationSourceType sourceType,
```
ricardas-buc, line 336: Maybe consider splitting into upload/download services? both service and unit test are reaching 500 lines

ricardas-buc (approved):

src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/service/ConfigurationServiceImpl.java
justasnortal: Extracted Anchor related code to separate method.

justasnortal:
[rerun ci]

justasnortal:
[rerun ci]

PR #1521, opened 2023-02-10
feat: XRDDEV-2242 API to view the details of a management request

VPaliliunas:


src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/entity/RequestEntity.java
```diff
@@ -66,6 +66,16 @@ public abstract class RequestEntity extends AuditableEntity {
     @Getter
     private int id;
 
+    @Column(name = "server_owner_name")
+    @Getter
+    @Setter
+    private String serverOwnerName;
```
ricardas-buc, line 72: How are these getting populated? Legacy implementation did duplicate data in database, we should avoid it whenever it is possible. For management requests we even have a view which is used in lists, it does not use this column.

ricardas-buc (changes requested):

src/central-server/admin-service/core-api/src/main/java/org/niis/xroad/cs/admin/api/service/ManagementRequestService.java
```diff
@@ -60,6 +61,7 @@ class Criteria {
      */
     Optional<Request> getRequest(int id);
 
+    Optional<ManagementRequestView> getRequestView(int id);
```
ricardas-buc, line 64: spacing

src/central-server/admin-service/core-api/src/main/java/org/niis/xroad/cs/admin/api/domain/ManagementRequestView.java
```diff
@@ -49,11 +50,14 @@ public class ManagementRequestView {
     private String securityServerOwnerName;
```
ricardas-buc, line 50: This DTO should fully replace ManagementRequestInfoDto, this one was created during refactoring, but never fully utilized

src/central-server/openapi-model/src/main/resources/openapi-definition.yaml
```diff
@@ -4826,6 +4826,68 @@ components:
...
+          readOnly: true
+        origin:
+          $ref: '#/components/schemas/ManagementRequestOrigin'
+        comments:
+          type: string
+          readOnly: true
+        type:
+          type: string
+          readOnly: true
+        security_server_identifier_id:
```
ricardas-buc, line 4845: How about reusing existing ManagementRequest object, but expanding it? Plus instead of using encodedIds use clientId/serverId objects and add encodedId field to those objects for places where it is needed.

src/central-server/admin-service/core-api/src/main/java/org/niis/xroad/cs/admin/api/domain/ManagementRequestView.java
```diff
@@ -49,11 +50,14 @@ public class ManagementRequestView {
...
     private String xroadInstance;
     private String memberCode;
-
     private String memberClass;
-
     private String serverCode;
-
+    private String clientOwnerName;
```
ricardas-buc, line 55: map to XRoadId types where it is possible. This will give us flexibility to quickly get encodedId or separate values

ricardas-buc (changes requested):

src/central-server/openapi-model/src/main/resources/openapi-definition.yaml
```diff
@@ -4826,6 +4826,68 @@ components:
         - member_class
         - member_code
         - server_code
+    ManagementRequestView:
```
ricardas-buc, line 4829: maybe ManagementRequestDetails would be more correct name? Then just extend ManagementRequest object and add missing data (probably certificates?)

ricardas-buc (approved):

PR #1533, opened 2023-02-23
XRDDEV-2259 Added API for getting security server authentication certificates

andresrosenthal:
https://nordic-institute.atlassian.net/browse/XRDDEV-2259


src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/service/SecurityServerServiceImpl.java
```diff
@@ -118,6 +123,17 @@ public List<FlattenedSecurityServerClientView> findClients(SecurityServerId serv
                 .getOrElseThrow(() -> new NotFoundException(SECURITY_SERVER_NOT_FOUND));
     }
 
+    @Override
+    public Set<SecurityServerAuthenticationCertificateDetails> findAuthCertificates(SecurityServerId id) {
+        return securityServerRepository.findBy(id)
+                .map(securityServerMapper::toTarget)
```
justasnortal, line 129: There is no need to map to dto.

andresrosenthal: No explicit necessity indeed

justasnortal (approved):

PR #1543, opened 2023-03-01
feat: Management request approve/decline API

VPaliliunas:
Management request approve/decline APi
Refs: XRDDEV-2240


src/central-server/admin-service/int-test/src/intTest/java/org/niis/xroad/cs/test/glue/ManagementRequestsApiStepDefs.java
```diff
@@ -58,45 +62,56 @@ public class ManagementRequestsApiStepDefs extends BaseStepDefs {
     private FeignManagementRequestsApi managementRequestsApi;
 
     private Integer managementRequestId;
-
     private byte[] authenticationCertificate;
 
     @SuppressWarnings("checkstyle:MagicNumber")
-    @Step("new security server {string} authentication certificate registered")
-    public void newSecurityServerRegistered(String securityServerId) throws Exception {
+    @Step("new security server {string} authentication certificate registered from {string}")
```
ricardas-buc, line 68: mention that this is origin, now it is unclear what "from" means if you're just reading report.

ricardas-buc (approved):

PR #1557, opened 2023-03-09
test: new integration tests

justasnortal:


src/central-server/admin-service/int-test/src/intTest/resources/behavior/api/securityServerApi.feature
```diff
@@ -37,6 +37,36 @@ Feature: Security Server API
...
+    And new member 'CS:TEST:member' is added with name 'name first'
+    And new member 'CS:TEST:another' is added with name 'name second'
+    And new security server 'CS:TEST:member:SS-1' authentication certificate registered with origin 'SECURITY_SERVER'
+    And management request is approved
+    And new security server 'CS:TEST:member:SS-3' authentication certificate registered with origin 'SECURITY_SERVER'
+    And management request is approved
+    And new security server 'CS:TEST:another:SS-2' authentication certificate registered with origin 'SECURITY_SERVER'
+    And management request is approved
+    Then security servers list, query '<$query>' paged by <$pageSize>, page <$requestedPage> contains <$itemsInPage> entries, <$itemsTotal> in total
+    Examples:
```
ricardas-buc, line 53: Examples does generate separate tests and @Modifying tag makes them reset database every time.
in this particular care it seems that we're browsing same endpoint, but with different values. Can't we just use same table, but process it in the stepDef and just execute multiple times under same step?

ricardas-buc (changes requested):

ricardas-buc (approved):

PR #1572, opened 2023-03-17
feat: Central server backup delete endpoint

VPaliliunas:
refs: XRDDEV-2275


src/central-server/admin-service/int-test/src/intTest/resources/behavior/api/backupsApi.feature
```diff
@@ -40,6 +40,17 @@ Feature: Backups API
     When Backup named test_backup.tar is downloaded
     Then Response is of status code 200
 
+  @Modifying
+  @ClearBackups
+  Scenario: Backup can be deleted
+    And Backup test_backup.tar is uploaded
```
mikkbachmann, line 46: Should this be 'Given' instead?

mikkbachmann (approved):

PR #1581, opened 2023-03-27
feat: API for deleting a trusted anchor

justasnortal:
https://nordic-institute.atlassian.net/browse/XRDDEV-2296


src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/service/TrustedAnchorServiceImpl.java
```diff
@@ -117,6 +120,18 @@ public TrustedAnchor findByHash(String hash) {
                 .orElseThrow(() -> new NotFoundException(TRUSTED_ANCHOR_NOT_FOUND));
     }
 
+    @Override
+    public void delete(String hash) {
+        final TrustedAnchorEntity trustedAnchor = trustedAnchorRepository.findFirstByTrustedAnchorHash(hash)
+                .orElseThrow(() -> new NotFoundException(TRUSTED_ANCHOR_NOT_FOUND));
```
ricardas-buc, line 126: you can also ass hash to exception, then it would be visible under metadata.

ricardas-buc (approved):

justasnortal:
[rerun-ci]

PR #1591, opened 2023-04-03
XRDDEV-2319: drop trigger insert_node_name and set db session variable xroad.current_ha_node_name

justasnortal:
https://nordic-institute.atlassian.net/browse/XRDDEV-2319


src/central-server/admin-service/infra-jpa/src/main/java/org/niis/xroad/cs/admin/jpa/config/CurrentHANodeNameSettingTransactionManager.java
```diff
@@ -0,0 +1,68 @@
...
+
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+
+/**
+ * Sets xroad.current_ha_node_name setting for each transaction.
+ * Required for history table stored procedures.
+ */
+@Component("currentHANodeNameSettingTransactionManager")
+public class CurrentHANodeNameSettingTransactionManager extends JpaTransactionManager {
```
mloitm, line 47: I would have used `spring.datasource.hikari.connection-init-sql` property (would have to create custom DataSource bean to parametrize it), but I guess this approach would work too.

Could you add a (int-)test as well to make sure it is applied?

mloitm (approved):

PR #1595, opened 2023-04-05
fix: global configuration hash calculation

justasnortal:
Hash should not include content type header.

### DIFF
--- a/src/central-server/admin-service/globalconf-generator/src/main/java/org/niis/xroad/cs/admin/globalconf/generator/MultipartMessage.java
+++ b/src/central-server/admin-service/globalconf-generator/src/main/java/org/niis/xroad/cs/admin/globalconf/generator/MultipartMessage.java
@@ -60,16 +60,28 @@ class MultipartMessage {
     }
 
     public String toString() {
-        var header = header("Content-Type",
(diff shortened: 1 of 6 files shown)


src/central-server/admin-service/globalconf-generator/src/main/java/org/niis/xroad/cs/admin/globalconf/generator/MultipartMessage.java
```diff
@@ -60,16 +60,28 @@ private static String secureRandomBoundary() {
     }
 
     public String toString() {
-        var header = header("Content-Type",
-                String.format("%s; charset=UTF-8; boundary=%s", contentType, boundary));
+        return toString(false);
+    }
+
+    public String toStringWithoutContentType() {
```
mloitm, line 66: The spec says:

>>> Signature of the directory, created using private key of the configuration source. The signature is calculated over the body of the first MIME part.

https://docs.x-road.global/Protocols/pr-gconf_x-road_protocol_for_downloading_configuration.html#23-format-of-signed-configuration

So theoretically, if there would be other headers, those would have to be skipped as well, so method name should reflect that too.  Maybe something like` toStringWithoutHeaders()` or  `bodyToString()` would be better?

mloitm (approved):

PR #1599, opened 2023-04-07
fix: configuration signing keys lookup by tokenId replaced to search …

justasnortal:
…by key identifiers.

Token identifier might be the same in cluster env.

Refs: XRDDEV-2324


justasnortal:
[rerun-ci]

ricardas-buc (approved):

PR #1610, opened 2023-04-13
XRDDEV-2308 UI to add and delete global group members

ovidijusnortal:


src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/entity/FlattenedSecurityServerClientViewEntity.java
```diff
@@ -87,6 +90,12 @@ public class FlattenedSecurityServerClientViewEntity extends AuditableEntity {
 
     @OneToMany(fetch = FetchType.LAZY, mappedBy = "flattenedSecurityServerClientView")
     private Set<FlattenedServerClientEntity> flattenedServerClients = new HashSet<>();
+
+    @ManyToOne(fetch = FetchType.EAGER, cascade = CascadeType.ALL)
```
justasnortal, line 94: Do we need cascade here?

ovidijusnortal: yea no reason for it

ovidijusnortal:
[rerun-ci]

ovidijusnortal:
[rerun-ci]

src/central-server/admin-service/int-test/src/intTest/java/org/niis/xroad/cs/test/api/FeignClientsApi.java
```diff
@@ -52,6 +52,7 @@ ResponseEntity<PagedClientsDto> findClients(
             @Size(min = 0, max = 255) @Valid @RequestParam(value = "member_code", required = false) String memberCode,
             @Size(min = 0, max = 255) @Valid @RequestParam(value = "subsystem_code", required = false) String subsystemCode,
             @Valid @RequestParam(value = "client_type", required = false) ClientTypeDto clientType,
-            @Valid @RequestParam(value = "security_server", required = false) String securityServer
+            @Valid @RequestParam(value = "security_server", required = false) String securityServer,
+            @Size(min = 0, max = 255)  @Valid @RequestParam(value = "excluding_group", required = false) Integer excludingGroup
```
justasnortal, line 56: @Size is not supported for Integer type, and @Valid probably not needed here.

justasnortal (approved):

PR #1630, opened 2023-04-25
fix: Subsystem registration approval

andresrosenthal:
Refs: XRDDEV-2301


andresrosenthal:
[rerun-ci]

src/central-server/admin-service/int-test/src/intTest/resources/behavior/api/managementRequestsApi.feature
```diff
@@ -66,6 +66,60 @@ Feature: Management requests API
...
+  @Modifying
+  Scenario: Register an existing subsystem as security server client
+    Given new security server 'CS:E2E:member-1:SS-X' authentication certificate registered with origin 'SECURITY_SERVER'
+    And management request is approved
+    And new member 'CS:E2E:member-2' is added
+    And new subsystem 'CS:E2E:member-2:subsystem-1' is added
+    And member 'CS:E2E:member-2' subsystems contains 'subsystem-1'
+    When client 'CS:E2E:member-2:subsystem-1' is registered as security server 'CS:E2E:member-1:SS-X' client from 'SECURITY_SERVER'
+    And management request is approved
+    And security server 'CS:E2E:member-1:SS-X' clients contains 'CS:E2E:member-2:subsystem-1'
```
mikkbachmann, line 92: This scenario is missing a 'Then'. Should one of these 'Ands' be 'Then'?

src/central-server/admin-service/int-test/src/intTest/resources/behavior/api/managementRequestsApi.feature
```diff
@@ -66,6 +66,60 @@ Feature: Management requests API
...
+  Scenario: Register & delete a new subsystem as security server client
+    Given new security server 'CS:E2E:member-1:SS-X' authentication certificate registered with origin 'SECURITY_SERVER'
+    And management request is approved
+    And new member 'CS:E2E:member-2' is added
+    And member 'CS:E2E:member-2' subsystems does not contain 'subsystem-1'
+    When client 'CS:E2E:member-2:subsystem-1' is registered as security server 'CS:E2E:member-1:SS-X' client from 'SECURITY_SERVER'
+    And management request is approved
+    And security server 'CS:E2E:member-1:SS-X' clients contains 'CS:E2E:member-2:subsystem-1'
+    And member 'CS:E2E:member-2' subsystems contains 'subsystem-1'
+    Then 'CS:E2E:member-2:subsystem-1' is deleted as security server 'CS:E2E:member-1:SS-X' client
```
mikkbachmann, line 79: with this wording (and for similar scenario above) it seems like the deletion is the outcome of registration. Is this correct?

src/central-server/admin-service/int-test/src/intTest/resources/behavior/api/managementRequestsApi.feature
andresrosenthal: **And** on line 92 should be replaced with a **Then** indeed

src/central-server/admin-service/int-test/src/intTest/resources/behavior/api/managementRequestsApi.feature
andresrosenthal: Pretty much copy-pasted from the member "register & delete" scenario. I guess it was done like this to include two actions into a single scenario. But I agree splitting register & delete into separate scenarios would perhaps be more correct from readability's point of view.

PR #1654, opened 2023-05-11
fix: Ensure management & registration services running with a valid API KEY conf

andresrosenthal:
Refs: XRDDEV-2382


andresrosenthal:
**NOTE:** As `xroad-center` package does not come with `road-center-registration-service` & `road-center-registration-service` -> Moved the management services' API KEY auto configuration into `xroad-centralserver` where the management & registration services are in fact included

VPaliliunas (approved):

PR #1674, opened 2023-05-23
docs: Backup Encryption Configuration

enelir:
Refs: XRDDEV-2376


petkivim (changes requested):
1. The list of installed packages in section 5.5 should be updated to match the current packages.

2. Section "2.10 Pre-configuration for Registration Web Service" needs to be updated a little bit since currently we only support installing the registration service on the same host with the admin service:

> The registration web service is installed by package xroad-center-registration-service. The registration web service can be installed on the same host with the central server or on a separate host. The package is included in the central server installation by default.

=>

> The registration web service is installed by package `xroad-center-registration-service`. The package is included in the central server installation by default.

PR #1677, opened 2023-05-23
feat: CS & SS configurability against host header injection

andresrosenthal:
Refs: XRDDEV-2387

### DIFF
--- a/src/common-rest-api/src/test/java/org/niis/xroad/restapi/controller/ApiKeysControllerTest.java
+++ b/src/common-rest-api/src/test/java/org/niis/xroad/restapi/controller/ApiKeysControllerTest.java
@@ -28,6 +28,7 @@ package org.niis.xroad.restapi.controller;
 import ee.ria.xroad.common.util.JsonUtils;
 
 import org.junit.jupiter.api.Test;
+import org.niis.xroad.restapi.config.AllowedHostnamesConfig;
 import org.niis.xroad.restapi.converter.PublicApiKeyDataConverter;
 import org.niis.xroad.restapi.domain.InvalidRoleNameException;
 import org.niis.xroad.restapi.domain.PersistentApiKeyType;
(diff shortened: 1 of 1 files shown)


src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/config/AllowedHostsConfig.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.ComponentScan;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.security.config.annotation.web.configuration.WebSecurityCustomizer;
+import org.springframework.security.web.firewall.StrictHttpFirewall;
+import org.springframework.util.CollectionUtils;
+
+@Configuration
+@ComponentScan(basePackages = {"org.niis.xroad.cs.admin"})
```
ricardas-buc, line 38: do we need extra component scan here?

src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/config/AllowedHostsConfig.java
```diff
@@ -0,0 +1,52 @@
...
+public class AllowedHostsConfig {
+
+    @Bean
+    public WebSecurityCustomizer webSecurityCustomizer(AdminServiceProperties properties) {
```
ricardas-buc, line 42: both cs and ss beans seem identical, why not move to common-rest-api? Props can be passed similarly to ip throttling or api client settings.

src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/config/AllowedHostsConfig.java
andresrosenthal: Probably not, due to IDE complaint didn't think twice

src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/config/AllowedHostsConfig.java
andresrosenthal: Indeed, good catch

PR #1686, opened 2023-05-29
feat: logging the original IP of the management request

justasnortal:
Refs: XRDDEV-2351

### DIFF
--- a/src/central-server/admin-service/api-client/src/main/java/org/niis/xroad/cs/admin/client/XForwardedForHeaderFeignInterceptor.java
+++ b/src/central-server/admin-service/api-client/src/main/java/org/niis/xroad/cs/admin/client/XForwardedForHeaderFeignInterceptor.java
@@ -1,20 +1,21 @@
-/**
+/*
  * The MIT License
+ * <p>
  * Copyright (c) 2019- Nordic Institute for Interoperability Solutions (NIIS)
  * Copyright (c) 2018 Estonian Information System Authority (RIA),
  * Nordic Institute for Interoperability Solutions (NIIS), Population Register Centre (VRK)
  * Copyright (c) 2015-2017 Estonian Information System Authority (RIA), Population Register Centre (VRK)
- *
+ * <p>
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
  * in the Software without restriction, including without limitation the rights
  * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
  * copies of the Software, and to permit persons to whom the Software is
  * furnished to do so, subject to the following conditions:
- *
+ * <p>
  * The above copyright notice and this permission notice shall be included in
  * all copies or substantial portions of the Software.
- *
+ * <p>
  * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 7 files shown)


src/central-server/admin-service/api-client/src/main/java/org/niis/xroad/cs/admin/client/XForwardedForHeaderFeignInterceptor.java
```diff
@@ -0,0 +1,58 @@
...
+        final RequestAttributes requestAttributes = RequestContextHolder.getRequestAttributes();
+        if (requestAttributes instanceof ServletRequestAttributes) {
+            HttpServletRequest request = ((ServletRequestAttributes) requestAttributes).getRequest();
+
+            String clientIp = request.getHeader(X_FORWARDED_FOR);
+            if (isBlank(clientIp)) {
+                clientIp = request.getRemoteAddr();
+            }
+
+            template.header(X_FORWARDED_FOR, clientIp);
```
andresrosenthal, line 54: When `forward-headers-strategy` is set to either `native` or `framework` (which is also the current case for registration & management service) then:
```java
template.header(X_FORWARDED_FOR, request.getRemoteAddr());
```
is sufficient, as it already is able to handle cases when request contains **X-Forwarded-for** header.
Furthermore it also handles the case when X-Forwarded-For is a [comma seperated string](https://udn.realityripple.com/docs/Web/HTTP/Headers/X-Forwarded-For#Syntax)

src/common-rest-api/src/main/java/org/niis/xroad/restapi/util/RequestHelper.java
```diff
@@ -96,7 +98,23 @@ public void runInRequestScope(RequestScopeVoidOperation operation) {
...
     }
+
+    /**
+     * Gets request senders IP address
+     */
+    public String getRequestSenderIPAddress() {
+        String ipAddress = null;
+        final HttpServletRequest currentHttpRequest = getCurrentHttpRequest();
+        if (currentHttpRequest != null) {
+            ipAddress = currentHttpRequest.getHeader(X_FORWARDED_FOR);
```
andresrosenthal, line 113: Same thing, just:
```java
ipAddress = currentHttpRequest.getRemoteAddr();
```
is sufficient when `server.forward-headers-strategy=native` is added to admin service's `application.yml`

andresrosenthal:
In [registration-service's](https://github.com/nordic-institute/X-Road/blob/develop/src/packages/src/xroad/common/center-registration-service/etc/xroad/nginx/registration-service.conf#L37) & [management service's](https://github.com/nordic-institute/X-Road/blob/develop/src/packages/src/xroad/common/center-management-service/etc/xroad/nginx/management-service.conf#L27) nginx conf `$remote_addr` should be replaced with `$proxy_add_x_forwarded_for`.
When there are additional proxies to nginx between client -> management/registration service communication then the client IP is lost and the last proxy before nginx is seen as the client

andresrosenthal (changes requested):

src/common-rest-api/src/main/java/org/niis/xroad/restapi/util/RequestHelper.java
justasnortal: yes, it is sufficient currently. But the mentioned classes might be used from other modules with different configuration. We do not have it right now, but I wanted to be on the safe side here - it should work when headers are still present.

andresrosenthal (approved):

PR #1695, opened 2023-06-02
chore: Central Server HA installation manual updates

justasnortal:
Refs: XRDDEV-2356

Note. The Monitoring HA State endpoint part will be updated with implementation story (XRDDEV-2401)


petkivim (approved):
Looks good to me.

PR #1698, opened 2023-06-02
chore: Relocate default properties values

justasnortal:


src/central-server/management-service/core/src/main/java/org/niis/xroad/cs/management/core/configuration/ManagementServiceProperties.java
```diff
@@ -1,21 +1,21 @@
 /**
  * The MIT License
- *
```
ricardas-buc, line 3: A "hack" for future: u  can switch loging to a simple comment and then intellij would not format it. ( /** -> /* )

ricardas-buc (approved):

PR #1703, opened 2023-06-06
Relocate commons

justasnortal:
Renaming common-rest-api to common-admin-api.
Moving common-* modules to common/ folder.


ricardas-buc (approved):
Since build was successful -approving

PR #1709, opened 2023-06-07
chore: remove European Regional Development Fund logo from documents

raits:
Remove the European Regional Development Fund logo from documents as the required period has passed (2022).

Additionally removed a few excess .gitkeep and .gitignore files from documentation folders.


petkivim (approved):
Look good to me.

PR #1712, opened 2023-06-08
feat: unified identifier checks

mloitm:
Refs: XRDDEV-1960


src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/openapi/validator/IdentifierCharsValidator.java
justasnortal, line 1: The identical class is in the central server. Maybe we can move it to commons and do not duplicate?
The same for IdentifierValidationErrorInfo, IdentifierChars classes (maybe more, didn't check)

mloitm: I moved IdentiferChars and NoControlChars validator related classes to commons.

justasnortal (approved):

PR #1741, opened 2023-07-07
feat: Introduce "minimum supported Client version" configuration parameter for Server Proxy

andresrosenthal:
Refs: XRDDEV-1999


enelir (approved):

src/proxy/src/main/java/ee/ria/xroad/proxy/serverproxy/ClientProxyVersionCheck.java
```diff
@@ -0,0 +1,74 @@
...
+import org.semver4j.Semver;
+
+import javax.servlet.http.HttpServletRequest;
+
+import java.util.Optional;
+
+import static ee.ria.xroad.common.ErrorCodes.X_CLIENT_PROXY_VERSION_NOT_SUPPORTED;
+
+@Slf4j
+public class ClientProxyVersionCheck {
```
andresrosenthal, line 45:
```suggestion
public class ClientProxyVersionVerifier {
```

src/proxy/src/main/java/ee/ria/xroad/proxy/serverproxy/ClientProxyVersionCheck.java
```diff
@@ -0,0 +1,74 @@
...
+    private static final String UNKNOWN_VERSION = "unknown";
+    private static final Semver MIN_SUPPORTED_CLIENT_VERSION;
+
+    static {
+        MIN_SUPPORTED_CLIENT_VERSION = Optional.ofNullable(SystemProperties.getServerProxyMinSupportedClientVersion())
+                .map(Semver::new)
+                .orElse(null);
+    }
+
+    void check(HttpServletRequest request) {
```
andresrosenthal, line 55: As this class does seem like an utility class I think this method should should be made static.
Note that then there's no need to inject an instance of this class to `ServerProxyHandler`

src/proxy/src/main/java/ee/ria/xroad/proxy/serverproxy/ClientProxyVersionCheck.java
```diff
@@ -0,0 +1,74 @@
...
+        String clientVersion = getVersion(request.getHeader(MimeUtils.HEADER_PROXY_VERSION));
+        String thisVersion = getVersion(ProxyMain.readProxyVersion());
```
andresrosenthal, line 57: This line could be moved to the line just before the ```if (!clientVersion.equals(thisVersion)) {``` check, where `thisVersion` variable is used for the first time.

src/proxy/src/test/java/ee/ria/xroad/proxy/serverproxy/ClientProxyVersionCheckTest.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Test
+    public void shouldRaiseError() {
+        System.setProperty(MIN_SUPPORTED_CLIENT_VERSION, VERSION_7_1_3);
+        final HttpServletRequest request = mock(HttpServletRequest.class);
+        when(request.getHeader(MimeUtils.HEADER_PROXY_VERSION)).thenReturn(CLIENT_VERSION_6_26_3);
+        clientProxyVersionCheck = new ClientProxyVersionCheck();
+
+        assertThrows(CodedException.class, () -> clientProxyVersionCheck.check(request));
```
andresrosenthal, line 72: Could you please also assert exception's message?

PR #1755, opened 2023-07-20
Fix: Rename configuration parameter wal_keep_segments to wal_keep_size

qrlt:
Starting from Postgresql 13, there has been a parameter rename and value change. 
https://www.postgresql.org/docs/13/release-13.html

wal_keep_size = wal_keep_segments * wal_segment_size (typically 16MB)

Latest X-Road is using Postgresql 14 and also needs to rename this parameter.


raits:
Hello,

Thank you for your contribution. I have added a ticket to the teams' backlog and we will review these changes as soon as possible.

The related ticket for the team can be found here: https://nordic-institute.atlassian.net/browse/XRDDEV-2452

mloitm (approved):

PR #1775, opened 2023-08-17
Add GitHub actions to run automated tests

raits:
Added a workflow that does the following:

- Reacts to pull request events (when opened and also when updated or reopened)
- Reacts to push events on the develop branch (this is required for the Gradle actions caching to work: https://github.com/gradle/gradle-build-action#using-the-caches-read-only)
- Builds the source code and runs unit tests along with integration tests
- Creates Debian packages for Ubuntu Jammy
- Uses packages created in the first job to run two additional jobs
  - One for the Central Server system tests
  - One for the Security Server system tests

To achieve this, I introduced a new parameter to the system tests (docker-root), which makes it possible to configure where the `Dockerfile` for the test is located. This defaults to the old hard-coded value.

Additionally introduced separate `Dockerfile`'s for the Central Server and Security Server to be used with GitHub actions, since trying to allow both this and the old use-case would have become too messy. The new files are located under `.github/docker/`.

When merging this PR, it would probably be best to squash the commits since there was quite a lot of experimenting to get it working, resulting in a lot of extra commits.


raits:
The last build failure on Jenkins looks like it is due to disk space issues there, but since the same test was run on the actions as well and I didn't change anything related to functionality, it should be fine.

ricardas-buc (approved):
Looks good. Ideally I would like to see a single dockerfile for both github and jenkins, but with our current setup it is needed.

raits:
> Looks good. Ideally I would like to see a single dockerfile for both github and jenkins, but with our current setup it is needed.

I agree, lets try to remember to come back to this in the future and see if we can improve things.

PR #1777, opened 2023-08-18
fix: add unique constraint to system_parameters to avoid duplicate rows…

mikkbachmann:
…and handle its violation with status code '409: Conflict'

refs: XRDDEV-2444

### DIFF
--- a/src/common/common-admin-api/src/main/java/org/niis/xroad/restapi/auth/securityconfigurer/CookieAndSessionCsrfTokenRepository.java
+++ b/src/common/common-admin-api/src/main/java/org/niis/xroad/restapi/auth/securityconfigurer/CookieAndSessionCsrfTokenRepository.java
@@ -52,7 +52,7 @@ public class CookieAndSessionCsrfTokenRepository implements CsrfTokenRepository 
     private HttpSessionCsrfTokenRepository httpSessionCsrfTokenRepository;
 
     /**
-     * Creates an instance of CookieAndSessionCsrfTokenRepository which holds a instances of
+     * Creates an instance of CookieAndSessionCsrfTokenRepository which holds instances of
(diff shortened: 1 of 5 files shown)


raits:
@mikkbachmann how safe is it to delete those values, is it possible that we might cause problems for existing instances when doing this?

Since we also have other spots in the database that require unique constraints to be added, I think we should consider if there is a better way to handle fixing the data. For instance, could we check this as an installation step, give the administrator an error stating what the issue is and fail the package installation until they have manually resolved it? This would allow the administrators to handle it based on their case.

Lets discuss it a bit at the daily on Monday.

src/common/common-admin-api/src/main/java/org/niis/xroad/restapi/auth/securityconfigurer/CookieAndSessionCsrfTokenRepository.java
```diff
@@ -52,7 +52,7 @@ public class CookieAndSessionCsrfTokenRepository implements CsrfTokenRepository
     private HttpSessionCsrfTokenRepository httpSessionCsrfTokenRepository;
 
     /**
-     * Creates an instance of CookieAndSessionCsrfTokenRepository which holds a instances of
+     * Creates an instance of CookieAndSessionCsrfTokenRepository which holds an instances of
```
juan-morales, line 55: :+1: 

success is in the details!

mikkbachmann:
> @mikkbachmann how safe is it to delete those values, is it possible that we might cause problems for existing instances when doing this?
> 
> Since we also have other spots in the database that require unique constraints to be added, I think we should consider if there is a better way to handle fixing the data. For instance, could we check this as an installation step, give the administrator an error stating what the issue is and fail the package installation until they have manually resolved it? This would allow the administrators to handle it based on their case.
> 
> Lets discuss it a bit at the daily on Monday.

Removed the delete statement. The installation step was done under ticket XRDDEV-2472.

andresrosenthal (approved):

PR #1788, opened 2023-09-04
Fix documentation for finding xroad id for cmd conf restore

qrlt:
This change will replace command in documentation regarding finding the X-Road Securityserver ID for command line restore section.

It's necessary since ID cannot be extracted from backup(gpg) files anymore.


raits (approved):
Thank you for your contribution! Looks good to me.

PR #1807, opened 2023-09-29
fix: allow recovering from timestamping service failure

mloitm:
Security server never recovered from timestamping service outage in case there were no messages timestamp in database.

Refs: XRDDEV-2398


mloitm:
[rerun-ci]

justasnortal (approved):

PR #1808, opened 2023-10-03
docs: Update System Parameters guide

petkivim:
- Remove an outdated section about about changing the global configuration interval on the Central Server.


raits (approved):
Looks good to me.

PR #1811, opened 2023-10-05
feat: java 17

justasnortal:
Refs: XRDDEV-2490

### DIFF
--- a/doc/Manuals/ig-ss_x-road_v6_security_server_installation_guide_for_rhel.md
+++ b/doc/Manuals/ig-ss_x-road_v6_security_server_installation_guide_for_rhel.md
@@ -205,22 +205,29 @@ Requirements to software and settings:
 
 *This is an optional step only for RHEL 7.*
 
-Java 17 is a prerequisite for running the Security Server. However, it is not included in the default repository and should be installed manually. Sample installation steps for Azul JDK are provided below, but other JDK distributions can also be used. 
+Java 17 is a prerequisite for running the Security Server. However, it is not included in the default repository and should be installed manually. Sample installation steps for Eclipse Temurin JDK are provided below, but other JDK distributions can also be used. 
 
-Follow these steps to install Azul Java 17 on RHEL 7:
+Follow these steps to install Temurin JDK 17 on RHEL 7:
 * Check the current java version:
 
         java -version
 
 If the current Java version is 17, following steps should be skipped.
 
-* Add the Azul repository to your system:
+* Add the Adoptium repository to your system:
 
-        sudo yum install -y https://cdn.azul.com/zulu/bin/zulu-repo-1.0.0-1.noarch.rpm
+        sudo cat <<EOF > /etc/yum.repos.d/adoptium.repo
+        [Adoptium]
+        name=Adoptium
+        baseurl=https://packages.adoptium.net/artifactory/rpm/${DISTRIBUTION_NAME:-$(. /etc/os-release; echo $ID)}/\$releasever/\$basearch
+        enabled=1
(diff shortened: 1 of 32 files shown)


ovidijusnortal:
yarn gradle plugin also has different versions depending on java version
` id 'org.siouan.frontend-jdk11' version "$frontendVersion"`
there is for java 17 variant also

justasnortal:
[rerun-ci]

doc/Manuals/ig-ss_x-road_v6_security_server_installation_guide_for_rhel.md
raits, line 1: Overall, the instructions are good. However, I wonder if the Azul version is the best one for us to pick here. When we were looking at options for replacing OpenJDK 8 when Ubuntu was set to drop support for it, we had a study to look into different providers: https://nordic-institute.atlassian.net/wiki/spaces/XRDDEV/pages/13176448/Open-source+Java+8+distributions+for+Ubuntu.

As noted on the page, we had some concerns about the support for the community edition. 

Back then, we preferred AdoptOpenJDK, which has now moved to an Eclipse foundation project as `temurin`. Since we also use it in our GitHub workflows, it might be better to go with that. It does appear to have RHEL 7 packages available: https://adoptium.net/installation/linux/.

src/central-server/admin-service/int-test/src/intTest/resources/application-override.yml
```diff
@@ -25,6 +25,7 @@ test-automation:
     testable-container:
       reuse-between-runs: ${reuse-between-runs}
       spring-boot:
+        base-image: azul/zulu-openjdk:17
```
raits, line 28: Could we use the default JAVA package available under Ubuntu 20/22 or RHEL 8 instead of Azul? Or as an alternative use `temurin` if we decide to suggest it for RHEL7 as well.

src/central-server/management-service/int-test/src/intTest/resources/application-override.yml
```diff
@@ -21,6 +21,7 @@ test-automation:
     testable-container:
       reuse-between-runs: true
       spring-boot:
+        base-image: azul/zulu-openjdk:17
```
raits, line 24: Same comment here regarding using the Azul package.

src/security-server/admin-service/int-test/src/intTest/resources/application-override.yml
```diff
@@ -25,6 +25,7 @@ test-automation:
     testable-container:
       reuse-between-runs: ${reuse-between-runs}
       spring-boot:
+        base-image: azul/zulu-openjdk:17
```
raits, line 28: Same comment here regarding the Azul package.

raits (changes requested):
Overall, it looks good, but I added some thoughts regarding the JAVA package choice.

doc/Manuals/ig-ss_x-road_v6_security_server_installation_guide_for_rhel.md
justasnortal: Added instructions how to install Temurin JDK on RHEL 7.

src/central-server/admin-service/int-test/src/intTest/resources/application-override.yml
justasnortal: Using eclipse-temurin:17-jammy as base image in int tests.

raits (approved):
Looks good to me now.

justasnortal:
[rerun-ci]

mloitm (approved):

PR #1831, opened 2023-10-23
feat: share global conf over HTTPS in configuration proxy

mloitm:


mloitm:
[rerun-ci]

src/packages/src/xroad/ubuntu/generic/xroad-confproxy.postinst
```diff
@@ -1,43 +1,114 @@
...
+
+  db_get xroad-common/cert-generation-retry
+  if [ "$RET" = false ]; then
+    exit 1
+  fi
+
+  echo "Retrying certificate generation"
+}
+
+create_conf_proxy_certificate() {
```
ricardas-buc, line 24: A minor thing, but having cert generation in a separate script file and just calling it during pacakge installation might allow quick re-generation whenever it is needed.

ricardas-buc (approved):

src/packages/src/xroad/ubuntu/generic/xroad-confproxy.postinst
mloitm: Re-generation can be invoked with dpkg-reconfigure. Also, the bulk of certificate generation is already in generate_certificate.sh script that can be invoked manually. 
I think I'm going to leave it as is it is for now.

PR #1836, opened 2023-10-27
fix: configuration client fixes regarding V3 introduction

andresrosenthal:
Refs: XRDDEV-1689


src/common/common-util/src/main/java/ee/ria/xroad/common/conf/globalconf/VersionedConfigurationDirectory.java
```diff
@@ -276,6 +320,16 @@ public static ConfigurationPartMetadata getMetadata(Path fileName) throws IOExce
         }
     }
 
+    public static boolean isVersion3(Path filePath) {
```
mloitm, line 323: This method name is tied to `CURRENT_GLOBAL_CONFIGURATION_VERSION`  value and should be renamed when value changes. Maybe rename it `isCurrentVersion`?

mloitm (approved):

PR #1839, opened 2023-10-30
Xrddev 2463 Security Server Vue 3 migration

mikkbachmann:


ovidijusnortal (approved):
Looks good from quick look, way to many files and changes for detailed run through and we probably all hidden(if they are) will be find and fixes in hands on testing or just basic usage of UI

PR #1850, opened 2023-11-08
fix: Avoid throwing exception when target shared parameters non existent during global group check

andresrosenthal:
Refs: XRDDEV-1689


andresrosenthal:
[rerun-ci]

src/common/common-verifier/src/main/java/ee/ria/xroad/common/conf/globalconf/GlobalConfImpl.java
```diff
@@ -548,7 +548,11 @@ public boolean isSubjectInGlobalGroup(ClientId subjectId, GlobalGroupId groupId)
     }
 
     SharedParameters.GlobalGroup findGlobalGroup(GlobalGroupId groupId) {
-        return getSharedParameters(groupId.getXRoadInstance()).getGlobalGroups().stream()
+        SharedParameters sharedParameters = confDir.getShared(groupId.getXRoadInstance());
+        if (sharedParameters == null) {
+            return null;
```
justasnortal, line 553: Would it require many changes to return Optional.empty() instead of null?

justasnortal (approved):

PR #1856, opened 2023-11-14
test: securityserver proxy healtcheck test cases 

ricardas-buc:
Refs: XRDDEV-1865


src/security-server/system-test/src/intTest/java/org/niis/xroad/ss/test/addons/SsAddonTestConfiguration.java
```diff
@@ -69,6 +74,21 @@ public FeignXRoadSoapRequestsApi feignManagementRequestsApi(
                 .target(FeignXRoadSoapRequestsApi.class, "http://localhost");
     }
 
+    @Bean
+    @SuppressWarnings("checkstyle:MagicNumber")
```
mloitm, line 78: Looks like unnecessary suppression, I don't see magic numbers here.

mloitm (approved):

PR #1863, opened 2023-11-20
 feat: Make confclient take into account verificiation certificates from shared-params when verifying configuration directory's signature

andresrosenthal:
Refs: XRDDEV-2519


src/configuration-client/src/main/java/ee/ria/xroad/common/conf/globalconf/ConfigurationDownloader.java
```diff
@@ -137,6 +149,51 @@ DownloadResult download(ConfigurationSource source, String... contentIdentifiers
...
+    private void addSupplementaryVerificationCerts(ConfigurationLocation location, List<byte[]> potentialCertsToBeAdded) {
+        List<byte[]> certsToBeAdded = new ArrayList<>();
+        potentialCertsToBeAdded.forEach(potentialCert -> {
+            if (location.getVerificationCerts().stream().noneMatch(cert -> Arrays.equals(cert, potentialCert))) {
+                certsToBeAdded.add(potentialCert);
+            }
+        });
+
+        if (!certsToBeAdded.isEmpty()) {
+            log.info("Adding supplementary verification certificates from shared parameters configuration file");
```
ricardas-buc, line 192: can we move this to debug or trace?

src/security-server/system-test/src/intTest/java/org/niis/xroad/ss/test/globalconf/glue/GlobalConfStepDefs.java
```diff
@@ -0,0 +1,70 @@
...
+    TestContainerService testContainerService;
+
+    @SuppressWarnings("checkstyle:MagicNumber")
+    @Step("Central Server's global conf is updated by a new active signing key")
+    public void centralServersGlobalConfSignKeysAreRotated() {
+        var newGlobalConfFiles = MountableFile.forClasspathResource("files/global_conf_signed_with_rotated_keys");
+        reverseProxyAuxiliaryContainer.getTestContainer().copyFileToContainer(newGlobalConfFiles, "/var/lib/xroad/public");
+
+        // As Security server polls for global conf every 3 secs, ensure that SS has the new global conf loaded
+        Selenide.sleep(4000);
```
ricardas-buc, line 58: is 1sec gap enough for slow instances like CI?

ricardas-buc (approved):

src/security-server/system-test/src/intTest/java/org/niis/xroad/ss/test/globalconf/glue/GlobalConfStepDefs.java
andresrosenthal: Initial (also slowest) configuration download took ~0.25 secs on my local virtualized docker environment.
CI instances have weaker specs true, but they're also linux meaning they run on native docker.
1 sec buffer should be enough in my estimation, but we could go with 2 secs just to be sure.

src/configuration-client/src/main/java/ee/ria/xroad/common/conf/globalconf/ConfigurationDownloader.java
andresrosenthal: This line is logged when a new configuration signing key is added or when the location cache (containing the verification certificates) is invalidated, which in practice happens very rarely. Leaving it on info means that this line is logged in production by default, which might come in handy.

PR #1865, opened 2023-11-21
feat: added couple of ACME properties to global conf

mikkbachmann:
refs: XRDDEV-2496

These are ACME related global configuration changes, that go with version 7.4. Other ACME related code changes will be added to version 7.5.

### DIFF
--- a/src/central-server/admin-service/globalconf-generator/src/main/java/org/niis/xroad/cs/admin/globalconf/generator/SharedParameters.java
+++ b/src/central-server/admin-service/globalconf-generator/src/main/java/org/niis/xroad/cs/admin/globalconf/generator/SharedParameters.java
@@ -57,6 +57,8 @@ class SharedParameters {
         private CaInfo topCA;
         private List<CaInfo> intermediateCAs;
         private String certificateProfileInfo;
+        private String acmeServerDirectoryURL;
+        private String acmeServerIPAddress;
     }
 
     @Data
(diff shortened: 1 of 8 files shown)


src/central-server/admin-service/globalconf-generator/src/main/java/org/niis/xroad/cs/admin/globalconf/generator/SharedParameters.java
```diff
@@ -68,6 +67,13 @@ public static class CaInfo {
         private byte[] cert;
     }
 
+    @Data
+    @AllArgsConstructor
+    public static class AcmeServer {
+        private String acmeServerDirectoryURL;
```
andresrosenthal, line 73:
```suggestion
        private String directoryURL;
```

src/central-server/admin-service/globalconf-generator/src/main/java/org/niis/xroad/cs/admin/globalconf/generator/SharedParameters.java
```diff
@@ -68,6 +67,13 @@ public static class CaInfo {
...
+        private String acmeServerIPAddress;
```
andresrosenthal, line 74:
```suggestion
        private String ipAddress;
```

src/central-server/admin-service/globalconf-generator/src/main/java/org/niis/xroad/cs/admin/globalconf/generator/SharedParameters.java
andresrosenthal: Enclosing class already defines the "acme server" context, no need for `acmeServer.acmeServerDirectory` redundancy.

andresrosenthal (changes requested):

andresrosenthal (approved):

PR #1866, opened 2023-11-22
feat: Update the Ubuntu packaging so that new installations use ports…

mikkbachmann:
… 8080 and 8443 by default similar to RHEL so that we open up port 80 and 443 to be used for ACME

refs: XRDDEV-2521


src/packages/src/xroad/ubuntu/generic/xroad-proxy.preinst
```diff
@@ -37,6 +37,44 @@ if [ "$1" = "upgrade" ];
...
+            present_in_local_ini=$(crudini --get ${local_ini} proxy connector-host 2>/dev/null)
+            if [[ -n "$present_in_local_ini" ]];
+              then
+                echo "client-https-port already present in local.ini, do not update local.ini"
+              else
+                echo "client-https-port not present in local.ini, update local.ini"
+                crudini --set ${local_ini} proxy client-https-port 443
+            fi
+        fi
+fi
```
petkivim, line 76: In my opinion, these checks should be implemented for `/etc/xroad/conf.d/local.ini` and not for `/etc/xroad/conf.d/override-securityserver-ee.ini`. This PR updates the `client-http-port` and `client-https-port` default property values defined in `proxy.ini`. That changes the values for all installations (new + upgrade) unless the default values have been overridden in a country-specific meta package (e.g., `override-securityserver-ee.ini`) or `local.ini`. Instead, the aim is to update the values for new installations only, not for upgrades. In addition, the Estonian meta package should keep the current defaults.

To avoid changing the values for existing installations, the old default values (`client-http-port=80` / `client-https-port=443`) must be written to `proxy.ini`. However, before writing the values, it must be checked if a local override exists already in `proxy.ini`. In that case, the existing local override must not be modified. Therefore, the logic for updating `proxy.ini` should be like this:

1. Does `client-http-port` / `client-https-port` already exist in `/etc/xroad/conf.d/local.ini`?
2. If yes, do nothing.
3. If no, add `client-http-port=80` / `client-https-port=443` to `/etc/xroad/conf.d/local.ini`.

There's no need to implement additional (pre)processing steps for the Estonian meta package. It's enough that the Estonian meta package overrides the default values (`client-http-port=80` / `client-https-port=443`) since the values in the Estonian meta package are applied to all new and existing installations (install + upgrade) if there's no local override in `local.ini`.

petkivim: Also, the migration should consider the version number of the Security Server that's upgraded. The migration should be executed only when upgrading from a Security Server version prior to 7.4.0 (< 7.4.0) to version 7.4.0 or higher. We don't want that the migration is executed when upgrading from 7.4.0 to a new version, e.g., 7.4.1, 7.5.0, etc. Executing the migration when upgrading from 7.4.0 to a new version would mean that the ports would change from `8080` / `8443` to `80` / `443` after the upgrade.

mikkbachmann: I updated the ee meta-package check to only check for the existence of the file not whats inside it and added the version check. Also refactored the echoed messages to better describe what's going on

petkivim (approved):

mloitm (approved):

PR #1878, opened 2023-12-01
chore: Update configuration signing key addition success message

andresrosenthal:


src/central-server/admin-service/ui/src/locales/en.json
```diff
@@ -447,7 +447,7 @@
         "title": "Add key",
         "labelField": "Key friendly name",
         "confirmButton": "Add",
-        "success": "Sign Key '{label}' added successfully"
+        "success": "Sign Key '{label}' added successfully. Remember not to activate it immediately, but rather after it has been distributed to all of the security servers."
```
petkivim, line 450: I'd change the message a little bit:

> Sign Key '{label}' added successfully. Please do not activate the key immediately. Please wait at least two global configuration refresh cycles before activating it.

petkivim: That's a very good point. Maybe we should mention the older Security Servers too since failing to remember them could cause a big service break. So maybe further update the notification like this:

> Sign Key '{label}' added successfully. Please do not activate the key immediately. Security Servers before version 7.4.0 require manually uploading a new configuration anchor before the activation. Starting from version 7.4.0, Security Servers pick up the new key automatically, but please wait at least two global configuration refresh cycles before activating it.

andresrosenthal: Yep, I also think the previous text would've been too long

petkivim (approved):

PR #1897, opened 2023-12-11
docs: Security Server LDAP setup over PAM

ricardas-buc:
https://nordic-institute.atlassian.net/browse/XRDDEV-2473


petkivim (approved):
Looks good to me. Even if this task was about the Security Server only, the same instructions apply for the Central Server too since it supports the LDAP mapping as well. To avoid having exactly the same instructions in both Security Server and Central Server user guides, we could add to the Central Server user guide a reference to the instruction available in the Security Server user guide. In that way, it would be enough to maintain the instructions in one document only.

PR #1903, opened 2023-12-13
XRDDEV-2546: Packaging fixes for RHEL

justasnortal:
Since java17 becomes available in the system only after posttrans phase, xroad services must be restarted after it.
Restarting xroad services in the posttrans phase.


justasnortal:
[rerun-ci]

mloitm (approved):

PR #1904, opened 2023-12-14
fix(serverconf): Correct the `hibernate-tools` dependency

sschuberth:
As discovered by an analysis with ORT [1], the `hibernate-tools` dependency does not exist in the currently set version "6.2.9.Final", making the `:serverconf:schemaExport` Gradle task fail. It was only available until version "5.6.15.Final", see [2]. The successor providing `org.hibernate.tool.ant.HibernateToolTask` seems to be `hibernate-tools-ant` [3], so use that instead.

However, this does not fully fix running `:serverconf:schemaExport` as it still fails with

    * What went wrong:
    Execution failed for task ':serverconf:schemaExport'.
    > java.lang.NoClassDefFoundError: ee/ria/xroad/common/identifier/ClientId$Conf

which is something that needs to be investigated separately.

[1]: https://github.com/oss-review-toolkit/ort
[2]: https://mvnrepository.com/artifact/org.hibernate/hibernate-tools
[3]: https://mvnrepository.com/artifact/org.hibernate.tool/hibernate-tools-ant

### DIFF
--- a/src/central-server/admin-service/ui/package.json
+++ b/src/central-server/admin-service/ui/package.json
@@ -22,7 +22,7 @@
   "dependencies": {
     "@fontsource/open-sans": "^5.0.12",
     "@mdi/font": "^7.2.96",
-    "@niis/shared-ui": "file:../../../shared-ui/niis-shared-ui-3.0.0.tgz",
+    "@niis/shared-ui": "../../../shared-ui",
     "@vee-validate/i18n": "~4.11.6",
     "@vee-validate/rules": "~4.11.6",
     "axios": "~1.4.0",
--- a/src/central-server/admin-service/ui/yarn.lock
+++ b/src/central-server/admin-service/ui/yarn.lock
@@ -829,9 +829,9 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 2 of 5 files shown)


sschuberth:
> which is something that needs to be investigated separately.

I've fixed this in a follow-up commit in the same PR.

sschuberth:
I've added another commit to fix `yarn install` issues for the `admin-service` UI projects.

sschuberth:
FYI, and I've also signed the CLA by now.

sschuberth:
Please let me know if the failing tests are really related to my problem, and how to reproduce them in my local environment eventually.

andresrosenthal:
The issue seems to be in Sonarcloud: `SONAR_TOKEN` secret is not passed to the runner as the workflow is triggered from a forked repository.

andresrosenthal:
@sschuberth Could you merge upstream's latest `develop` branch? It should fix the Jenkins build.

sschuberth:
> @sschuberth Could you merge upstream's latest `develop` branch? It should fix the Jenkins build.

I've rebased my PR and force-pushed.

andresrosenthal:
Congratulations on becoming a contributor to X-Road!

PR #1905, opened 2023-12-15
docs: Recreating Configuration Anchor file to publish over HTTPS

enelir:
Refs: XRDDEV-2493

### DIFF
--- a/doc/Manuals/ug-cs_x-road_6_central_server_user_guide.md
+++ b/doc/Manuals/ug-cs_x-road_6_central_server_user_guide.md
@@ -617,27 +617,13 @@ To delete an anchor file, follow these steps.
 
 ## 5.9 Publishing global configuration over HTTPS
 
-To publish global configuration over HTTPS it is necessary that configuration anchor XML file contains download url over HTTPS as well. When X-road is installed before 7.4 then internal/external configuration anchor should re-create manually. After recreating configuration anchor:
-- internal configuration anchor file must be distributed to the Security Server administrators and uploaded to the Security Servers;
-- external configuration anchor file must be distributed to the federation partner’s Central Server (or configuration proxy) administrators and uploaded to the Central Server (or configuration proxy).
+Starting from version 7.4.0, the Central Server supports publishing global configuration over HTTP and HTTPS. Instead, before version 7.4.0, only HTTP was supported.
 
-For example, recreated internal configuration anchor XML file looks like:
+Starting from version 7.4.0, a new private key and a self-signed TLS certificate are created automatically when installing a new Central Server or upgrading an existing installation from an older version. After the installation or upgrade, the Central Server Administrator must manually apply for a TLS certificate from a trusted Certificate Authority (CA) and then configure the certificate. The CA must be trusted by the Security Server's Java installation. More information about configuring the TLS certificate on the Central Server is available in the Central Server Installation Guide [CSI](#13-references).
 
-```xml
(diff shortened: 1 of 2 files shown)


doc/Manuals/ug-cs_x-road_6_central_server_user_guide.md
```diff
@@ -613,6 +615,30 @@ To delete an anchor file, follow these steps.
...
+        <downloadURL>http://test/internalconf</downloadURL>
+        <verificationCert>cert</verificationCert>
+    </source>
+    <source>
+        <downloadURL>https://test/internalconf</downloadURL>
+        <verificationCert>cert</verificationCert>
+    </source>
+</ns3:configurationAnchor>
+```
+
```
petkivim: Thanks to the changes in global configuration in X-Road version 7.4.0 and global configuration version `V3`, the `shared-params.xml` part of the global configuration now includes the global configuration download endpoints. It means that starting from version 7.4.0, the Security Server picks up the new HTTPS endpoint automatically. Instead, older Security Servers that still use global configuration `V2` don't need to know about them since they're not able to utilise the HTTPS endpoints. Therefore, I would change the text to this:

> Starting from version 7.4.0, the Central Server supports publishing global configuration over HTTP and HTTPS. Instead, before version 7.4.0, only HTTP was supported. 
>
> Starting from version 7.4.0, a new private key and a self-signed TLS certificate are created automatically when installing a new Central Server or upgrading an existing installation from an older version. After the installation or upgrade, the Central Server Administrator must manually apply for a TLS certificate from a trusted Certificate Authority (CA) and then configure the certificate. The CA must be trusted by the Security Server's Java installation. More information about configuring the TLS certificate on the Central Server is available in the Central Server Installation Guide [[CSI]](#13-references).
>
> Applying for a TLS certificate issued by a trusted CA is required, because the Security Server does not trust the new automatically generated self-signed certificate by default. The Security Server supports disabling certificate verification, but disabling it in production environments is not recommended. More information is available in the `[configuration-client]` section of the System Parameters User Guide [[UG-SYSPAR]](#13-references).

doc/Manuals/ug-sec_x_road_security_hardening.md
```diff
@@ -186,7 +186,7 @@ For example, setting the value `server-min-supported-client-version = 7.3.1` mea
 
 ## 5. Publish global configuration over HTTPS
 
-Starting from X-Road version 7.4, it is possible to publish global configuration over HTTPS using a TLS certificate issued by a trusted CA. The CA must be trusted by the Security Server's Java installation.
+Starting from X-Road version 7.4, it is possible to publish global configuration over HTTPS using a TLS certificate issued by a trusted CA. The CA must be trusted by the Security Server's Java installation. Also, the configuration anchor XML file should contain download URL over HTTPS, in detail see "Central Server user Guide" [UG-CS](#Ref_UG-CS) section "Publishing global configuration over HTTPS".
```
petkivim, line 189: Consider changing this:

> Also, the configuration anchor XML file should contain download URL over HTTPS, in detail see "Central Server user Guide" [UG-CS](#Ref_UG-CS) section "Publishing global configuration over HTTPS".

=>

> See the Central Server user Guide [[UG-CS]](#Ref_UG-CS) for details.

Also, the document version number and version history should be updated.

petkivim (changes requested):

petkivim (approved):

PR #1919, opened 2024-01-02
XRDDEV-2292: loopback ports added to documentation

justasnortal:
loopback ports added to documentation
unused properties deleted.

Refs: XRDDEV-2292


petkivim:
I think it would be better to add the loopback interfaces in a separate table instead of having them in the same table with the external ports. Also, a short explanation of what the loopback interfaces are in general and what they are used for would be good since it may not be clear to everyone. In addition, it should be emphasized that the loopback ports must be accessible locally from the Security Server host and not from external hosts.

PR #1922, opened 2024-01-03
feat: add pnpm workspaces for simplified UI development

ricardas-buc:
- Yarn replaced with pnpm which allows workspaces while still keeping vanilla node_modules per project structure.
- Simplified gradle front end build. shared-ui gradle module will be responsible of building whole workspace
- Consolidated dependency versioning into root package.json and shared-ui package.
- shared-ui now acts as parts of main project, VITE hmr functionality now works as expected


ovidijusnortal (approved):
looks good

PR #1930, opened 2024-01-11
feat: better UX for dialogs

ovidijusnortal:


ricardas-buc (approved):
On a high level - looks ok. PR is too big to do a proper review :)

PR #1934, opened 2024-01-17
build: avoid the use of deprecated Gradle constructs

mloitm:


src/build.gradle
```diff
@@ -32,7 +32,7 @@ sonarqube {
         property "sonar.dependencyCheck.reportPath", "build/reports/owasp-dependency-check/dependency-check-report.xml"
         property "sonar.dependencyCheck.htmlReportPath", "build/reports/owasp-dependency-check/dependency-check-report.html"
         property "sonar.exclusions", "**/build/generated-sources/**"
-        property "sonar.coverage.jacoco.xmlReportPaths", "${rootProject.layout.buildDirectory}/reports/jacoco/jacocoAggregatedReport/jacocoAggregatedReport.xml"
+        property "sonar.coverage.jacoco.xmlReportPaths", "${rootProject.layout.buildDirectory.get().asFile}/reports/jacoco/jacocoAggregatedReport/jacocoAggregatedReport.xml"
```
sschuberth, line 35: Shouldn't this then also rather be

    rootProject.layout.buildDirectory.dir('reports/jacoco/jacocoAggregatedReport/jacocoAggregatedReport.xml').get().asFile

to avoid string interpolation where not needed? Similar in more places below.

src/addons/messagelog/messagelog-addon/build.gradle
```diff
@@ -21,7 +21,7 @@ jar {
 
 tasks.register('copyGpg', Copy) {
     from 'src/test/gpg'
-    into "${layout.buildDirectory}/gpg"
+    into layout.buildDirectory.dir("gpg")
```
sschuberth, line 24: Nit: Now that no string interpolation is needed anymore, the string should use single-quotes instead of double-quotes. Similar in more places below.

src/build.gradle
mloitm: I agree, that it could be more consistent and we should set code style rules for gradle scripts. But that would be bigger change and I wouldn't wan't to incorporate it into this PR.

PR #1968, opened 2024-02-19
chore: extract identifiers, globalconf generated code out of common-util

ricardas-buc:


ovidijusnortal (approved):
Looks OK, vavr changes also seem to match Optional API

PR #2013, opened 2024-04-02
feat: migrate to gradle version catalog

ovidijusnortal:


src/common/common-admin-api/build.gradle
```diff
@@ -32,18 +27,18 @@ dependencies {
...
     testImplementation('org.springframework.boot:spring-boot-starter-test')
     testImplementation('org.springframework.security:spring-security-test')
-    testImplementation("org.hsqldb:hsqldb:$hsqldbVersion")
+    testImplementation(libs.hsqldb)
 
     implementation('com.fasterxml.jackson.datatype:jackson-datatype-jsr310')
-    implementation('jakarta.validation:jakarta.validation-api')
-    implementation("org.apache.tika:tika-core:$tikaVersion")
+    implementation(libs.jakarta.validationApi)
+    implementation("org.apache.tika:tika-core:2.9.1")
```
ricardas-buc, line 41: why added version?

src/gradle/libs.versions.toml
```diff
@@ -0,0 +1,112 @@
+[versions]
+# WARNING:
+# Dependabot does not properly handle version references in build files
+# for example 'implementation "io.dropwizard.metrics:metrics-core:${libs.versions.metrics.get()}"' will be ignored/won't get PR with version update if needed
+# In some cases referincing library directly like 'libs.protobuf.protoc' may not work, possible workaround could be:
+# protoc { artifact = libs.protobuf.protoc.get().toString() } where just using 'artifact = libs.protobuf.protoc' causes error
```
melbeltagy, line 6: I'm sorry but I'm confused by this comment. 
Is it a `warning`, or the reason why we are migrating to version catalog?

ovidijusnortal: no, just to notify some another dev that if he decides to have in build.gradle 
```
dependencies {
  implementation org.mapstruct:mapstruct:{libs.versions.mapstruc.get()t}
}
```
instead of something like this: 
```
dependencies {
  implementation libs.mapstruct
}
```
then dependabot won't notify or create PR for new library version.
for dependencies there is no need to reference just version but in some cases it's not possible to just use `libs.mapstruct` so you may need to do some additional steps to get it working

mikkbachmann (approved):

src/central-server/admin-service/int-test/build.gradle
```diff
@@ -13,14 +13,14 @@ dependencies {
...
-    intTestImplementation("com.nortal.test:test-automation-feign:$testAutomationFrameworkVersion")
-    intTestImplementation("com.nortal.test:test-automation-allure:$testAutomationFrameworkVersion")
-    intTestImplementation("org.liquibase:liquibase-core:$liquibaseVersion")
-    intTestImplementation("org.postgresql:postgresql:$postgresqlVersion")
-    intTestImplementation("org.projectlombok:lombok:${lombokVersion}")
+    intTestImplementation(libs.testAutomation.core)
+    intTestImplementation(libs.testAutomation.assert)
+    intTestImplementation(libs.testAutomation.containers)
+    intTestImplementation(libs.testAutomation.feign)
+    intTestImplementation(libs.testAutomation.allure)
```
ricardas-buc, line 20: maybe create test-automation bundle?

src/settings.gradle
```diff
@@ -1,7 +1,15 @@
 pluginManagement {
-  plugins {
-    id 'com.github.johnrengelman.shadow' version "$shadowJarVersion"
-  }
+    plugins {
+        id 'com.github.johnrengelman.shadow' version "8.1.1"
```
ricardas-buc, line 3: why not in version catalog?

ovidijusnortal: because here we get dilemma "which comes first versions or versions preprocessing" :D

src/build.gradle
```diff
@@ -3,22 +3,21 @@ buildscript {
...
     id 'jacoco-report-aggregation'
     id 'java'
     id 'idea'
-    id "io.spring.dependency-management" version "${springDependenciesVersion}" apply false
-    id 'org.springframework.boot' version "${springBootVersion}" apply false
-    id 'com.google.protobuf' version "$protobufGradleVersion" apply false
-    id 'org.siouan.frontend-jdk17' version "$frontendPluginVersion" apply false
+    alias(libs.plugins.springBoot) apply false
+    alias(libs.plugins.protobuf) apply false
+    alias(libs.plugins.frontendJDK17) apply false
```
ricardas-buc, line 20: `apply false` pattern was used to version control dependencies. With version catalog it should not be necessary .

ricardas-buc (approved):

PR #2172, opened 2024-05-28
build(npmrc): Explicitly configure the Node version to use

sschuberth:
While `package.json` files have `engines.node` set, that specifies a range of compatible runtime Node versions, but not the exact version to build the project.

Choose the current version from the 20 branch.

### DIFF
--- a/src/.npmrc
+++ b/src/.npmrc
@@ -1,4 +1,4 @@
 # pnpm-related options
 shamefully-hoist=true
 strict-peer-dependencies=false
-engine-strict = true
+engine-strict=true
--- a/src/.npmrc
+++ b/src/.npmrc
@@ -2,3 +2,4 @@
 shamefully-hoist=true
 strict-peer-dependencies=false
 engine-strict=true
+use-node-version=20.13.1


src/.npmrc
```diff
@@ -1,4 +1,5 @@
 # pnpm-related options
 shamefully-hoist=true
 strict-peer-dependencies=false
-engine-strict = true
+engine-strict=true
+use-node-version=20.13.1
```
sschuberth, line 5: I'm open to using a different version here according to your preferences @petkivim!

mloitm: Hi @sschuberth!
Node version is already managed by gradle plugin [here](https://github.com/nordic-institute/X-Road/blob/develop/src/shared-ui/build.gradle#L7), version number is defined in [gradle.properties](https://github.com/nordic-institute/X-Road/blob/develop/src/gradle.properties#L12).

Setting `use-node-version` here would cause it to download again even when using the same version number for gradle plugin.

sschuberth: I see. Maintaining the version in two places would not be nice anyway. I'll try to think of another approach to make the [ORT analyzer](https://oss-review-toolkit.org/ort/docs/tools/analyzer) use the correct Node version.

PR #2182, opened 2024-06-12
docs: Add ACME-related information to the documentation

petkivim:
- Update CS network diagram in the installation guide.
  - These changes are not related to ACME support, but they're improvements to the network diagram that was previously missing some information.
- Update SS network diagram in the installation guide and add a reference to the SS user guide.
- Add additional information about the ACME support to the SS user guide.
- Add information about the ACME support to the Security Architecture document.


raits (approved):
Looks good to me.

PR #2219, opened 2024-07-11
fix: disabled batch signature leads to message verification failures

ricardas-buc:


src/common/common-verifier/src/test/signatures/non-batch-sig/globalconf_v4/DEV/shared-params.xml.metadata
```diff
@@ -0,0 +1 @@
+{"configurationVersion":"3"}
```
justasnortal, line 1: Shouldn't it be version 4?

ricardas-buc: fixed

PR #2264, opened 2024-08-16
docs: Updates to the SS LB installation guide

petkivim:
- Add assumption that the load balancer must support TLS passthrough.
- Editorial updates.


raits (approved):
Looks good to me.

PR #2325, opened 2024-09-17
fix: apache cli default parser need also option's short name

enelir:
Refs: XRDDEV-2706

### DIFF
--- a/src/configuration-client/application/src/main/java/org/niis/xroad/confclient/ConfClientCLIMain.java
+++ b/src/configuration-client/application/src/main/java/org/niis/xroad/confclient/ConfClientCLIMain.java
@@ -55,10 +55,10 @@ public class ConfClientCLIMain {
     }
 
     /**
-     * Main entry point of configuration client. Based on the arguments, the client will either:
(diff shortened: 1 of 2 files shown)


src/configuration-client/application/src/main/java/org/niis/xroad/confclient/ConfClientCLIMain.java
```diff
@@ -87,9 +87,9 @@ private static CommandLine getCommandLine(String[] args) throws Exception {
         CommandLineParser parser = new DefaultParser();
         Options options = new Options();
 
-        options.addOption(OPTION_VERIFY_PRIVATE_PARAMS_EXISTS, false,
+        options.addOption("p", OPTION_VERIFY_PRIVATE_PARAMS_EXISTS, false,
```
justasnortal, line 90: I think the verify_internal_configuration.sh script should be updated, because it contains 
"... org.niis.xroad.confclient.ConfClientCLIMain -verifyPrivateParamsExists "$@"" and the "-verifyPrivateParamsExists" will not be matched with long name.

justasnortal (approved):

PR #2347, opened 2024-10-01
fix: use fixed time for proxy integration tests

mloitm:


src/common/common-core/src/main/java/ee/ria/xroad/common/util/TimeUtils.java
```diff
@@ -39,6 +42,8 @@
  * This class contains various time related utility methods.
  */
 public final class TimeUtils {
+    private static Clock clock = Clock.systemUTC();
```
ricardas-buc, line 45: Is utc correct here? I think by default it was using Clock.systemDefaultZone()

ricardas-buc (approved):

src/common/common-core/src/main/java/ee/ria/xroad/common/util/TimeUtils.java
mloitm: fixed

PR #2393, opened 2024-10-22
feat: add central server allure reports

raits:


.github/workflows/build.yaml
```diff
@@ -299,3 +369,37 @@ jobs:
...
+      - name: Add link to the report as a comment 
+        uses: actions/github-script@v7
+        if: always()
+        with:
+          script: |
+            github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: 'E2E Allure report: https://nordic-institute.github.io/X-Road/e2e/${{ github.run_number }}'
```
ricardas-buc, line 404: perhaps opening in a new tab would be preferred?

ricardas-buc:
There are few other places with allure reports, were those intentionally skipped or just missed? 
image

Would it be possible to consolidate all links into a single comment with a table? Maybe persist links into action runtime and create comment as a final step?

Also just a note: by generating report instead of using report generated by gradle, we loose execution logs tab which is a customization to see logs during whole execution.  Maybe zipping those logs and adding DL link to the same report table would suffice.

.github/workflows/build.yaml
raits: Based on what I found searching the topic it seems that this is not possible in GitHub comments.

PR #2415, opened 2024-11-05
fix: with removing client identifier should be remove related data as well

enelir:
Refs: XRDDEV-2729


src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/repository/LocalGroupRepository.java
```diff
@@ -48,12 +50,19 @@
...
 
     public LocalGroupType getLocalGroup(Long entityId) {
         LocalGroupDAOImpl localGroupDAO = new LocalGroupDAOImpl();
         return localGroupDAO.getLocalGroup(persistenceUtils.getCurrentSession(), entityId);
     }
 
+    public void deleteGroupMembersByMemberId(ClientId.Conf memberId) {
+        var session = persistenceUtils.getCurrentSession();
+        groupMemberDAO.findAllByGroupMemberId(session, memberId)
+                .forEach(session::remove);
```
mloitm, line 63: This could be done with single delete statement instead of fetching and deleting in a loop

mloitm (approved):

PR #2472, opened 2024-12-10
chore: update diagnostic group title to add help text for clarity

raits:
- Moved HelpButton to `components/ui/HelpButton.vue` from the original Keys and Certificates subview since it is a common component
- Made the help image optional since it is not required for all cases of help text
- Added help text to the mail notifications block to make it clearer which cases it will be used for


petkivim (approved):
Looks good to me.

PR #2476, opened 2024-12-12
feat: support non-batch signatures in message log and ASIC-E containers

mloitm:
Refs: XRDDEV-2687


src/asicverifier/src/test/resources/asic-containers/valid-signed-message/message.xml
```diff
@@ -1,25 +1,27 @@
...
+            <id:serviceVersion>v1</id:serviceVersion>
         </xroad:service>
-        <xroad:userId>EE:PIN:abc4567</xroad:userId>
-        <xroad:id>6289448e512292d17d54ef60ed2318e7</xroad:id>
-        <xroad:issue/>
+        <xroad:id>dd19fe68-7d37-41bd-b79c-8948ccdb36da</xroad:id>
+        <xroad:protocolVersion>4.0</xroad:protocolVersion>
     </SOAP-ENV:Header>
     <SOAP-ENV:Body>
-        <xroad:getState>aäõ</xroad:getState>
```
ricardas-buc, line 23: I assume idea here was to check for UTF8 support, might not be a big issue that we loose it, but worth noting

ricardas-buc (approved):

src/asicverifier/src/test/resources/asic-containers/valid-signed-message/message.xml
mloitm: Message is never parsed while generating or vaildating asic containers, so we are probably ok. This looks like a standard query from other tests, so it should be covered where it matters.

PR #2482, opened 2024-12-17
chore: Integrate a sample REST API (using OpenAPI) into the local environment

andresrosenthal:


ricardas-buc (changes requested):
As discussed, lets move this to a separate repo

ricardas-buc (approved):

PR #2516, opened 2025-01-09
chore: fix issues in X-Road documentation

raits:
Resolves the issue with heading levels being incorrect in various documents.

Also changed how the image is included in OpMon daemons architecture document so that it is properly included.


petkivim (approved):
Looks good to me.

PR #2547, opened 2025-01-17
chore: Add support for the publiccode.yml standard

petkivim:
More information about the standard is available [here](https://github.com/publiccodeyml/publiccode.yml).

### DIFF
--- a/publiccode.yml
+++ b/publiccode.yml
@@ -119,7 +119,7 @@ legal:
 localisation:
   availableLanguages:
     - en
-    - ee
+    - et
     - es
   localisationReady: true
 maintenance:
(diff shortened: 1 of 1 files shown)


publiccode.yml
```diff
@@ -0,0 +1,145 @@
...
+      - 'https://youtu.be/EvhssW_Uv8s'
+developmentStatus: stable
+legal:
+  license: MIT
+  mainCopyrightOwner: Nordic Institute for Interoperability Solutions (NIIS)
+  repoOwner: Nordic Institute for Interoperability Solutions (NIIS)
+localisation:
+  availableLanguages:
+    - en
+    - ee
```
raits, line 122: The ISO 639-2 language code for Estonian should be `et`: https://www.loc.gov/standards/iso639-2/php/langcodes_name.php?code_ID=133

raits (approved):
Just a minor correction about the Estonian language code, otherwise looks good to me.

publiccode.yml
petkivim: Fixed. It was a mix up with the country code.

PR #2641, opened 2025-02-14
fix: increase cs system-test timeout

enelir:


src/central-server/admin-service/ui-system-test/src/intTest/resources/application-override.yml
```diff
@@ -29,8 +29,8 @@ test-automation:
     driver-manager-enabled: true
     headless: true
     chrome-options-args: "--guest,--disable-features=OptimizationHints"
-    timeout: 10000
-    page-load-timeout: 10000
+    timeout: 15000
```
enelir, line 32: Latest cs system-test were wailing mostly with step:
```
User Approves Management Request for adding another client(member) to security server
	Element not found {By.xpath: //button[@data-test='add-member-button']}
```
Behind show button "Add member" is only priv check, that's why i thought the reason is probl loading problem

ricardas-buc (approved):

PR #2650, opened 2025-02-19
chore: Provision OpenBao for RHEL

andresrosenthal:


src/packages/src/xroad/redhat/SPECS/xroad-secret-store-local.spec
```diff
@@ -0,0 +1,148 @@
...
+        if [ -d "/opt/openbao/tls" ]; then
+            rm -f /opt/openbao/tls/tls.{key,crt} 2>/dev/null || true
+        fi
+        rm -f /etc/pki/ca-trust/source/anchors/openbao.crt 2>/dev/null || true
+    fi
+}
+
+trap cleanup EXIT
+
+if [ $1 -eq 1 ] || [ $1 -eq 2 ]; then  # 1 = fresh install, 2 = upgrade
```
ricardas-buc, line 61: Can we extract duplicate initialization code and share it between deb/rpm? This would also go inline with the idea of adding init scripts later

PR #2651, opened 2025-02-20
chore: use quarkus in op-monitor

justasnortal:


src/service/proxy/proxy-core/src/testFixtures/java/org/niis/xroad/proxy/core/test/util/ConfigUtils.java
```diff
@@ -0,0 +1,54 @@
...
+package org.niis.xroad.proxy.core.test.util;
+
+import io.smallrye.config.PropertiesConfigSource;
+import io.smallrye.config.SmallRyeConfigBuilder;
+import lombok.experimental.UtilityClass;
+
+import java.util.Map;
+
+@UtilityClass
+public class ConfigUtils {
```
ricardas-buc, line 37: Perhaps move to more generic place? perhaps fixtures under common properties?

ricardas-buc (approved):

PR #2664, opened 2025-02-28
chore: remove gitbook related files as we can't use it for documentation

raits:
We can't use GitBook anyway, so lets remove files we don't otherwise need.


petkivim (approved):
Looks good to me.

PR #2676, opened 2025-03-06
chore: Use Utilize OpenBao provided PKI for remaining "internal" TLS connections

andresrosenthal:


src/service/op-monitor/op-monitor-core/src/main/java/org/niis/xroad/opmonitor/core/QueryRequestProcessor.java
```diff
@@ -81,6 +81,9 @@ class QueryRequestProcessor {
         this.healthDataHandler = new HealthDataRequestHandler(
                 healthMetricRegistry);
 
+        log.info("++++++++++");
```
ricardas-buc, line 84: Perhaps move to debug?

ricardas-buc (approved):
Looks good, but additional logging should not be on info

PR #2685, opened 2025-03-11
feat: add serverconf and SS identifiers entity classes

enelir:
Refs: XRDDEV-2805


src/lib/serverconf-core/src/main/java/org/niis/xroad/serverconf/converter/GenericUniDirectionalMapper.java
```diff
@@ -1,20 +1,21 @@
 /*
  * The MIT License
+ * <p>
```
mloitm, line 3: License header should not have html tags, check your editor configuration.

src/lib/serverconf-core/src/main/java/org/niis/xroad/serverconf/entity/AccessRightTypeEntity.java
mloitm, line 1: Class names are a bit confusing, entities are *TypeEntity, sometimes *Entity, plain model objects are *Type. Maybe 
the "Type" part could be dropped like in CS?

mloitm (approved):

PR #2691, opened 2025-03-19
feat: the rest path never logged on the consumer side and on the producer side only the rest path of the saved endpoints is logged

enelir:
Refs: XRDDEV-2840


src/security-server/system-test/src/intTest/java/org/niis/xroad/ss/test/addons/glue/OpMonitorStepDefs.java
```diff
@@ -81,22 +80,41 @@ public class OpMonitorStepDefs extends BaseStepDefs {
     private static final String OPERATIONAL_DATA_JSON = "operational-monitoring-data.json.gz";
     private static final String OP_MONITORING_XSD = "http://x-road.eu/xsd/op-monitoring.xsd";
 
-    @SuppressWarnings("checkstyle:OperatorWrap")
+    @SuppressWarnings({"checkstyle:OperatorWrap", "checkstyle:MagicNumber"})
     @Step("Security Server Operational Data request was sent")
     public void executeGetSecurityServerOperationalData() throws Exception {
+        Thread.sleep(2000);
```
raits, line 86: Why is this sleep needed, is there a way we could get rid of it? My concern is that this might make the test unreliable if the CI environment for some reason is slower than usual.

raits (changes requested):
Overall looks good, but lets consider the use of Thread.sleep(). I would like to avoid it if possible.

src/security-server/system-test/src/intTest/java/org/niis/xroad/ss/test/addons/glue/OpMonitorStepDefs.java
enelir: In default setup operational monitoring statistics are available after 1 min when REST or SOAP request is made. Here, i changed system test env properties from 60 second to 1 secund:
```
[op-monitor]
records-available-timestamp-offset-seconds=1
[op-monitor-buffer]
socket-timeout-seconds=1
[op-monitor-service]
socket-timeout-seconds=1
```
But still 2 seconds of sleep is needed here to get statistics.

raits (approved):
Looks good to me.

PR #2701, opened 2025-03-25
feat: Local Kubernetes cluster deployment for containerized Security Server

andresrosenthal:
Refs: XRDDEV-2850


ricardas-buc (approved):
Looks like a good starting point for k8s. There is no point of nitpicking right now, we will iteratively improve this once we start trying various deployment scenarios.

PR #2728, opened 2025-04-06
feat: As a Developer I want to look into supporting ARM builds on Git…

mikkbachmann:
…Hub so that we can have ARM64 development images available

refs: XRDDEV-2835


.github/workflows/build.yaml
```diff
@@ -95,6 +114,50 @@ jobs:
...
+        env:
+          DEBEMAIL: 'info@niis.org'
+          DEBFULLNAME: 'NIIS'
+        run: ./src/packages/build-deb.sh jammy -release
+      - name: Build 24.04 (Noble) packages
+        env:
+          DEBEMAIL: 'info@niis.org'
+          DEBFULLNAME: 'NIIS'
+        run: ./src/packages/build-deb.sh noble -release
+      - name: Store deb files for system tests
```
andresrosenthal, line 155: system tests just use AMD packages, ARM packages are only used to publish images

.github/workflows/build.yaml
```diff
@@ -95,6 +114,50 @@ jobs:
...
+          dependency-graph: generate-and-submit
+          add-job-summary-as-pr-comment: always
+      - name: Download workspace
+        uses: pyTooling/download-artifact@v4 # https://github.com/actions/upload-artifact/issues/38
+        with:
+          name: workspace-for-arm
+      - name: Build native dependencies
+        working-directory: ./src
+        run: ./gradlew :service:signer:signer-api:make
+      - name: Build 22.04 (Jammy) packages
```
andresrosenthal, line 145: Should we consider building RHEL packages on ARM as well?
They're not needed elsewhere in the workflow, but it could serve as an additional check to verify that RHEL builds work on ARM.

andresrosenthal (approved):
Aside from a few small notes, looks good to me

PR #2732, opened 2025-04-07
chore: use DB in configuration-client to store the configuration-anchor

justasnortal:
Refs: XRDDEV-2867


src/service/configuration-client/configuration-client-core/src/main/java/org/niis/xroad/confclient/core/globalconf/ConfigurationAnchorProvider.java
```diff
@@ -0,0 +1,42 @@
...
+
+public interface ConfigurationAnchorProvider {
+
+    boolean isAnchorPresent();
+
+    Optional<byte[]> get() throws Exception;
+
+    void save(byte[] content) throws Exception;
+
+    String source();
```
ricardas-buc, line 40: Enum maybe?

ricardas-buc (approved):

src/service/configuration-client/configuration-client-core/src/main/java/org/niis/xroad/confclient/core/globalconf/ConfigurationAnchorProvider.java
justasnortal: It is used only on log messages, would help to identify problems if occurs. In case of file is used, it will contain the file path.

PR #2737, opened 2025-04-10
feat: implement reloadable vault key manager with retry logic and con…

ricardas-buc:
…figuration properties

Refs: XRDDEV-2879


src/common/common-properties/src/main/resources/xroad-common.yaml
```diff
@@ -53,3 +55,6 @@ xroad:
           - 127.0.0.1
         ttl-minutes: 720 # 12 hours
         refresh-interval-minutes: 300 # 5 hours
+        retry-base-delay: 5s
+        retry-exponential-backoff-multiplier: 1.5
+        retry-max-attempts: 10
```
justasnortal, line 60: Is it needed? There is already @WithDefault(..) in CommonRpcProperties

ricardas-buc: Spring part does not have any defaults for any of rpc props. But fair enough, gonna clean up a bit

justasnortal (approved):

justasnortal (approved):

PR #2778, opened 2025-04-30
Enhance the backup feature to support local OpenBao in native deployment setup

andresrosenthal:


src/packages/src/xroad/common/secret-store-local/usr/share/xroad/scripts/backup_openbao_db.sh
```diff
@@ -0,0 +1,37 @@
...
+  fi
+  pg_options="-c client-min-messages=warning -c search_path=$db_schema"
+else
+  echo "Unable to determine OpenBao PostgreSQL connection URL in /etc/openbao/openbao.hcl"
+  exit 1
+fi
+
+# Reading custom libpq ENV variables
+if [ -f /etc/xroad/db_libpq.env ]; then
+  source /etc/xroad/db_libpq.env
```
ricardas-buc, line 29: db_libpq is serverconf specific. We should probably create a new env file for openbao db.

src/packages/src/xroad/common/secret-store-local/usr/share/xroad/scripts/restore_openbao_db.sh
```diff
@@ -0,0 +1,104 @@
...
+  exit 1
+fi
+
+# Reading custom libpq ENV variables
+if [ -f /etc/xroad/db_libpq.env ]; then
+  source /etc/xroad/db_libpq.env
+fi
+
+if [[ ! -z $PGOPTIONS_EXTRA ]]; then
+  PGOPTIONS_EXTRA=" ${PGOPTIONS_EXTRA}"
```
ricardas-buc, line 49: PG env variables seem to be serverconf specific (for example messagelog does not use em). We might encounter some issues if we try reusing same vars.

src/packages/src/xroad/common/secret-store-local/usr/share/xroad/scripts/restore_openbao_db.sh
```diff
@@ -0,0 +1,104 @@
...
+}
+
+if [[ $FORCE_RESTORE == true ]] ; then
+  { cat <<EOF
+     DROP SCHEMA IF EXISTS "$db_schema" CASCADE;
+EOF
+  } | psql_dbuser || abort "Restoring database failed. Could not drop schema."
+fi
+
+# PostgreSQL 9.2 and earlier do not support CREATE SCHEMA IF NOT EXISTS
```
ricardas-buc, line 78: Do we still support 9.2?

ricardas-buc (changes requested):

src/packages/src/xroad/common/secret-store-local/usr/share/xroad/scripts/backup_openbao_db.sh
andresrosenthal: db_libpq.env is shared between of serverconf, messagelog & opmonitor databases during their setup (see [setup_database()](https://github.com/nordic-institute/X-Road/blob/develop/src/packages/src/xroad/common/base/usr/share/xroad/scripts/_setup_db.sh#L52-L59) which is invoked by [setup_messagelog_db.sh](https://github.com/nordic-institute/X-Road/blob/develop/src/packages/src/xroad/common/addon/proxy/setup_messagelog_db.sh#L18) & [setup_opmonitor_db.sh](https://github.com/nordic-institute/X-Road/blob/develop/src/packages/src/xroad/common/op-monitor/usr/share/xroad/scripts/setup_opmonitor_db.sh) both alike.
Although not fully sure but this does look like a intended feature to me.
If that's the case I do think db restore and init processes should act consistently in this regard.

src/packages/src/xroad/common/secret-store-local/usr/share/xroad/scripts/restore_openbao_db.sh
andresrosenthal: Same case as with db_libpq.env - PGOPTIONS_EXTRA are taken into account by all the databases during their setup - setup & restore processes should consistent.

src/packages/src/xroad/common/secret-store-local/usr/share/xroad/scripts/restore_openbao_db.sh
andresrosenthal: Nah, probably a good idea to clean it up from here and other DB scripts.

ricardas-buc (approved):

PR #2804, opened 2025-05-28
feat: As a Developer I want to add the capability of sharing operational monitoring data with the rest of the components via gRPC to the operational monitoring addon so that we can use it to visualise traffic

mikkbachmann:
refs: XRDDEV-2892, XRDDEV-2893


src/service/op-monitor/op-monitor-core/src/main/java/org/niis/xroad/opmonitor/core/OpMonitorRpcService.java
```diff
@@ -0,0 +1,93 @@
...
+            responseObserver.onNext(handleGetMonitoringDataRequest(request));
+        } catch (Exception e) {
+            log.error("getOperationalDataIntervals failed", e);
+            responseObserver.onError(e);
+        }
+        responseObserver.onCompleted();
+    }
+
+    private GetOperationalDataIntervalsResp handleGetMonitoringDataRequest(GetOperationalDataIntervalsReq request) throws Exception {
+        log.info("Getting operational data in {} minute intervals, between {} and {}",
```
mloitm: I think this should be on debug level.

PR #2809, opened 2025-06-03
feat: backup-manager module

justasnortal:
Refs: XRDDEV-2896

Depends on saving Signer data to DB (https://nordic-institute.atlassian.net/browse/XRDDEV-2883). Currently Signer restart is excluded when restoring backup, should be enabled after Signer DB changes.


deployment/security-server/docker/base/Dockerfile-backup-manager-baseline
```diff
@@ -0,0 +1,21 @@
...
+FROM ${REGISTRY_URL}/ss-baseline-runtime:latest
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get -qq update && \
+    apt-get -qq install --no-install-recommends curl postgresql-client && \
+    apt-get -qq clean && \
+    rm -rf /var/lib/apt/lists/*
+
+RUN ARCH=$(dpkg --print-architecture) && \
```
andresrosenthal, line 12: Just curious: why did you prefer to install kubectl binary with curl instead of using package manager?

src/packages/src/xroad/common/base/usr/share/xroad/scripts/_restore_xroad.sh
```diff
@@ -162,6 +163,29 @@ create_pre_restore_backup () {
...
+          die "Error occurred while creating pre-restore Openbao database backup" \
+              "to ${PRE_RESTORE_OPENBAO_DATABASE_DUMP_FILENAME}"
+        fi
+      fi
+      backed_up_files_cmd="${backed_up_files_cmd}; echo ${PRE_RESTORE_OPENBAO_DATABASE_DUMP_FILENAME}"
+    else
+      die "Failed to execute OpenBao database backup script at ${OPENBAO_DATABASE_BACKUP_SCRIPT} for" \
+          "doing pre-restore backup"
+    fi
+  fi
```
andresrosenthal, line 187: 👍

src/packages/src/xroad/ubuntu/generic/xroad-proxy.postinst
```diff
@@ -128,37 +116,38 @@ case "$1" in
...
+#    if [ $? -ne 0 ]; then
+#      echo "Error: SQL query execution failed"
+#    else
+#      echo "SQL query executed successfully. Renaming ${conf_anchor_file} to ${conf_anchor_file}.copy. It can be deleted."
+#      mv ${conf_anchor_file} ${conf_anchor_file}.copy
+#    fi
+#    rm "$tmp_sql"
+#  else
+#    echo "${conf_anchor_file} file not found, skipping migration to db."
+#  fi
```
andresrosenthal, line 150: Any particular reason why this was commented out & not moved to configuration-client's postinst?

deployment/security-server/docker/base/Dockerfile-backup-manager-baseline
justasnortal: Just because it was the first option in documentation :) Since it is not available in the default repos, adding the k8s repository and installing single package from it is about the same amount of steps.

src/packages/src/xroad/ubuntu/generic/xroad-proxy.postinst
justasnortal: It was in Proxy because the `read_db_properties.sh` script was in Proxy package as well. There are more things which will need the migration from file system to DB (properties, probably softtoken) and we do not have a migration plan yet.

PR #2845, opened 2025-06-15
feat: Introduce database based authentication for containerized deployment

andresrosenthal:


src/central-server/admin-service/ui/src/store/modules/adminUsers.ts
```diff
@@ -0,0 +1,137 @@
...
+ */
+import { useNotifications } from '@/store/modules/notifications';
+import { useRouter } from 'vue-router';
+import { useUser } from '@/store/modules/user';
+import * as api from '@/util/api';
+import {
+  AdminUser,
+  AdminUserPasswordChangeRequest,
+  AdminUsersHandler,
+  i18n,
```
ovidijusnortal, line 35: better to use `import { useI18n } from 'vue-i18n';`

src/security-server/admin-service/ui/src/store/modules/adminUsers.ts
```diff
@@ -0,0 +1,137 @@
...
+ */
+import { useNotifications } from '@/store/modules/notifications';
+import { useRouter } from 'vue-router';
+import { useUser } from '@/store/modules/user';
+import * as api from '@/util/api';
+import {
+  AdminUser,
+  AdminUserPasswordChangeRequest,
+  AdminUsersHandler,
+  i18n,
```
ovidijusnortal, line 34: same thing as above

src/common/common-admin-api/src/main/java/org/niis/xroad/restapi/service/AdminUserService.java
```diff
@@ -0,0 +1,112 @@
...
+
+    public List<AdminUser> getAll() {
+        return userRepository.getAll()
+                .stream()
+                .map(mapper::toDomainObject)
+                .toList();
+    }
+
+    public void create(AdminUser adminUser) {
+        adminUser = new AdminUser(
```
justasnortal: Doesn't look nice reassigning method parameter.

src/common/common-admin-api/src/main/java/org/niis/xroad/restapi/domain/AdminUser.java
```diff
@@ -0,0 +1,40 @@
...
+import lombok.RequiredArgsConstructor;
+
+import java.util.Set;
+
+@Getter
+@RequiredArgsConstructor
+public class AdminUser {
+    private final Long id;
+    private final String username;
+    private final String password;
```
justasnortal: It is not a good practice to use String for storing raw password, better to use char[]. Or have the password hashed.

src/central-server/admin-service/application/src/main/resources/application.yml
```diff
@@ -5,7 +5,7 @@ spring:
   jpa:
     open-in-view: false
     hibernate:
-      ddl-auto: validate
+      ddl-auto: none
```
ricardas-buc: was this intentional?

src/central-server/admin-service/ui/src/global.ts
```diff
@@ -64,6 +64,8 @@ export enum RouteName {
   SecurityServerAuthenticationCertificates = 'security-server-authentication-certificates',
   SecurityServerAuthenticationCertificate = 'security-server-authentication-certificate',
   SecurityServerClients = 'security-server-clients',
+  AdminUsers = 'admin-users',
```
ricardas-buc, line 67: Isn't calling users as admin users bit misleading? There will be observers, registration officers, etc. Users,BackofficeUsers, UiUsers, etc. might represent it better.

This is a general comment, if agreed, it should be applied to all places

src/common/common-admin-api/src/main/java/org/niis/xroad/restapi/auth/DatabaseAuthenticationProvider.java
```diff
@@ -0,0 +1,91 @@
...
+import org.springframework.security.authentication.UsernamePasswordAuthenticationToken;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.core.GrantedAuthority;
+import org.springframework.security.crypto.password.PasswordEncoder;
+
+import java.util.Set;
+
+@RequiredArgsConstructor
+public class DatabaseAuthenticationProvider implements AuthenticationProvider {
```
ricardas-buc, line 44: Have you considered extending DaoAuthenticationProvider?

src/common/common-admin-api/src/main/java/org/niis/xroad/restapi/auth/DatabaseAuthenticationProvider.java
```diff
@@ -0,0 +1,91 @@
...
+            throw e;
+        }
+    }
+
+    @Override
+    public boolean supports(Class<?> authentication) {
+        return authentication.equals(UsernamePasswordAuthenticationToken.class);
+    }
+
+    private void validateCredentialsLength(final String username, final String password) {
```
ricardas-buc: I am not sure if this is needed here. These does look like a requirements when creating a password/user.

src/common/common-admin-api/src/main/java/org/niis/xroad/restapi/service/AdminUserService.java
ricardas-buc: yea, looks like a bug.

src/common/common-admin-api/src/main/java/org/niis/xroad/restapi/service/ApiKeyService.java
```diff
@@ -70,6 +70,20 @@ public class ApiKeyService {
     private final SecurityHelper securityHelper;
     private final ApiKeyMapper apiKeyMapper;
 
+    public ApiKeyService(@Qualifier(API_KEY_ENCODER) PasswordEncoder passwordEncoder,
```
ricardas-buc: lombok should still work as Qualifier is defined in lombok.config

src/common/common-admin-api/src/main/resources/common-openapi-definition.yaml
```diff
@@ -133,6 +133,40 @@ components:
           items:
             type: string
             format: text
+    AdminUser:
+      type: object
+      properties:
+        id:
+          type: integer
+          description: Admin user ID
```
ricardas-buc, line 141: This would also be a good place to define username,password limitations

src/security-server/admin-service/infra-jpa/src/main/resources/liquibase/serverconf/007-admin-users.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.19.xsd">
+
+    <changeSet author="niis" id="006-admin-users">
+        <createTable tableName="admin_user">
```
ricardas-buc, line 9: would be good to have metada columns like 'create/modified timestamps', 'modified/created by' here, but on the other hand it does overlap with history tables. Still, can be considered

src/security-server/admin-service/infra-jpa/src/main/resources/liquibase/serverconf/007-admin-users.xml
```diff
@@ -0,0 +1,57 @@
...
+            </column>
+            <column name="role" type="VARCHAR(255)">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+
+        <addForeignKeyConstraint baseColumnNames="admin_user_id" baseTableName="admin_user_role" constraintName="fk_admin_user_role_admin_user_id" deferrable="false" initiallyDeferred="false" referencedColumnNames="id" referencedTableName="admin_user"/>
+        <addUniqueConstraint columnNames="admin_user_id, role" constraintName="unique_admin_user_role" tableName="admin_user_role" />
+
+        <sql splitStatements="false">
```
ricardas-buc, line 36: Is this related to admin user changeset? Can it be separate?

src/security-server/admin-service/infra-jpa/src/main/resources/liquibase/serverconf/007-admin-users.xml
```diff
@@ -0,0 +1,57 @@
...
+            </column>
+            <column name="admin_user_id" type="BIGINT">
+                <constraints nullable="false"/>
+            </column>
+            <column name="role" type="VARCHAR(255)">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+
+        <addForeignKeyConstraint baseColumnNames="admin_user_id" baseTableName="admin_user_role" constraintName="fk_admin_user_role_admin_user_id" deferrable="false" initiallyDeferred="false" referencedColumnNames="id" referencedTableName="admin_user"/>
```
ricardas-buc, line 33: if you separate admin_user_role into its own changeset you could also inline foreign key under constraint element

src/common/common-admin-api/src/main/java/org/niis/xroad/restapi/domain/AdminUser.java
andresrosenthal: Good catch, overlooked that

src/central-server/admin-service/application/src/main/resources/application.yml
andresrosenthal: Intentional indeed, it was either this or provide users liquibase to CS as well

src/central-server/admin-service/ui/src/global.ts
andresrosenthal: I do like "UiUsers" as a potential replacement.
I'll consider this in the follow-up PR regarding password policy

src/common/common-admin-api/src/main/java/org/niis/xroad/restapi/auth/DatabaseAuthenticationProvider.java
andresrosenthal: I did try it, but then the "AdminUserService" must implement UserDetailsService, which automatically autoconfigured an additional DaoAuthenticationProvider to the auth chain.

src/common/common-admin-api/src/main/java/org/niis/xroad/restapi/auth/DatabaseAuthenticationProvider.java
andresrosenthal: Will remove it.
Proper Username/password validations will be addressed in the follow-up PR.

src/common/common-admin-api/src/main/resources/common-openapi-definition.yaml
andresrosenthal: Yep, will adress this in the follow up PR.

src/security-server/admin-service/infra-jpa/src/main/resources/liquibase/serverconf/007-admin-users.xml
andresrosenthal: user role constraints? How can it be not related?

src/security-server/admin-service/infra-jpa/src/main/resources/liquibase/serverconf/007-admin-users.xml
andresrosenthal: Ok now I see what you mean. To be honest I even didn't think about this, but good to know 👍

src/common/common-admin-api/src/main/java/org/niis/xroad/restapi/service/ApiKeyService.java
andresrosenthal: Turns out that it's not present in lombok.config - will add and adjust

src/central-server/admin-service/ui/src/global.ts
ricardas-buc: We even have UiUserEntity that is a remnant of old impl :) can be reused. Resolving this assuming that it will be tackled in another PR

src/security-server/admin-service/infra-jpa/src/main/resources/liquibase/serverconf/007-admin-users.xml
ricardas-buc: I do see APIKEY_ROLES table here

andresrosenthal: Yea, the typo aspect was kinda a blind spot of mine, you could've led with that 😄

ricardas-buc (approved):

PR #2896, opened 2025-07-08
chore: Internal TLS key and certificate management via gRPC

andresrosenthal:


src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/InternalTlsCertificateService.java
```diff
@@ -68,49 +57,34 @@
 @Transactional
```
justasnortal: Is @Transactional needed?

src/service/proxy/proxy-application/src/main/resources/application.yaml
```diff
@@ -8,11 +8,20 @@ quarkus:
       - xroad-common.yaml
       - xroad-logging.yaml
 
+script:
+  generate-certificate:
+    path: /usr/share/xroad/scripts/generate_certificate.sh
+    args: -n internal -f -S -p 2>&1
```
justasnortal: Usually we put the default values in the @ConfigMapping interface.

justasnortal (approved):

PR #2898, opened 2025-07-09
chore: optimize operational data query for traffic visualization

mikkbachmann:
refs: XRDDEV-2892


raits (approved):
The documentation portion looks ok to me.

PR #2936, opened 2025-08-15
feat: add softHSM support to the development container

enelir:
Refs: XRDDEV-2727

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -345,7 +345,8 @@ jobs:
           push: true
           build-args: |
             PACKAGE_SOURCE=internal
-          tags: localhost:5000/xrd-securityserver:${{ github.sha }}
+          target: final-hwtoken
+          tags: localhost:5000/xrd-securityserver-hwtoken:${{ github.sha }}
       - name: Security Server test setup measurement
         uses: green-coding-solutions/eco-ci-energy-estimation@v5.0
         with:
@@ -353,7 +354,7 @@ jobs:
           label: 'Set up Security Server tests'
       - name: Run Security Server system tests
         working-directory: ./src
-        run: ./gradlew -Dorg.gradle.jvmargs=-Xmx1g :security-server:system-test:systemTest -PsystemTestSsImageName=localhost:5000/xrd-securityserver:${{ github.sha }}
+        run: ./gradlew -Dorg.gradle.jvmargs=-Xmx1g :security-server:system-test:systemTest -PsystemTestSsImageName=localhost:5000/xrd-securityserver-hwtoken:${{ github.sha }}
       - name: Test Security Server measurement
         uses: green-coding-solutions/eco-ci-energy-estimation@v5.0
         with:
@@ -446,7 +447,8 @@ jobs:
           push: true
           build-args: |
             PACKAGE_SOURCE=internal
-          tags: localhost:5000/xrd-securityserver:${{ github.sha }}
+          target: final-hwtoken
+          tags: localhost:5000/xrd-securityserver-hwtoken:${{ github.sha }}
       - name: E2E test setup measurement
         uses: green-coding-solutions/eco-ci-energy-estimation@v5.0
         with:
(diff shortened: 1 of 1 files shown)


Docker/securityserver/Dockerfile
```diff
@@ -86,6 +87,13 @@ RUN useradd xrd-sec && adduser xrd-sec xroad-security-officer && sh -c "echo 'xr
     && useradd xrd-sys && adduser xrd-sys xroad-system-administrator && sh -c "echo 'xrd-sys:$USER_PASSWD' | chpasswd" \
     && useradd xrd-obs && adduser xrd-obs xroad-securityserver-observer && sh -c "echo 'xrd-obs:$USER_PASSWD' | chpasswd"
 
+# Create token
+RUN adduser xroad softhsm \
+    && printf "\n[softhsm2]\nlibrary = /usr/lib/softhsm/libsofthsm2.so\nos_locking_ok = true\nlibrary_cant_create_os_threads = true\n" >> /etc/xroad/devices.ini \
+    && mkdir -p /var/lib/softhsm/tokens/ \
+    && softhsm2-util --init-token --slot 0 --label 'x-road-softhsm2' --so-pin 1234 --pin 'Secret1234' \
```
mloitm: Token should not be initialized in Dockerfile. This causes it to have the same token for every security server  container, that use the image.

Docker/xrd-dev-stack/local-dev-run.sh
```diff
@@ -29,6 +29,10 @@ fi
 docker compose $COMPOSE_FILE_ARGS --env-file "$ENV_FILE" up -d --build hurl
 
 if [[ -n "$INITIALIZE" ]]; then
+  docker compose $COMPOSE_FILE_ARGS --env-file "$ENV_FILE" \
+    exec ss0 sh -c "apt-get remove --purge -y softhsm2 xroad-addon-hwtokens && apt-get autoremove --purge -y"
```
mloitm: Instead of removing hwtokens here from containers that don't need it, I think it would be better to add it here to containers that need it.
Maybe create a separate docker image with hwtoken support?

.github/workflows/publish_images.yaml
```diff
@@ -118,8 +118,67 @@ jobs:
           context: ${{ github.workspace }}/Docker/securityserver
           build-args: |
             PACKAGE_SOURCE=internal
+          target: final
           push: true
           tags: ${{ matrix.architecture == 'arm64' && 'ghcr.io/nordic-institute/xrddev-security-server:arm-latest' || 'ghcr.io/nordic-institute/xrddev-security-server:amd-latest' }}
+  PublishHwTokenSS:
```
mloitm: Instead of copy-pasting to new job,  just adding new `Build and push Docker image` step to previous job would probably be enough. Less duplication. The same applies to `PublishMultiArchHwTokenSS` job

development/hurl/scenarios/setup.hurl
```diff
@@ -672,8 +672,42 @@ HTTP 201
 [Captures]
 ss1_auth_key_cert_hash: jsonpath "$.certificate_details.hash"
 
+# Get not available hardware token
```
mloitm: Is this the token softHSM generates by default? Maybe it could be deleted from entrypoint script instead?

Docker/securityserver/Dockerfile
```diff
@@ -57,17 +58,16 @@ RUN --mount=type=bind,source=build/packages,target=/tmp/packages \
     # Install
     && pg_ctlcluster 16 main start \
     && apt-get update \
-    && apt-get -y -qq install xroad-securityserver xroad-autologin xroad-addon-opmonitoring \
+    && apt-get -y -qq install xroad-securityserver xroad-autologin xroad-addon-hwtokens xroad-addon-opmonitoring \
```
mloitm: Could `xroad-addon-hwtokens` be installed final-hwtoken stage instead? This could probably mean, that clean-up steps would need to be moved and duplicated to final stages too.

mloitm (approved):

PR #2942, opened 2025-08-22
docs: Update service metadata protocol documents

petkivim:
- Update Security Server default HTTP and HTTPS ports


raits (approved):
Looks good to me.

PR #2945, opened 2025-08-26
chore: Store internal & opmonitor TLS credentials in OpenBao's KV

andresrosenthal:


Harini-royal12:
Could you share more details about the failing build step? It looks like the Build, test and package code job didn’t pass. Are there any dependencies or environment changes required for this update?

Harini-royal12:
I noticed that the Build, test and package code job is currently failing during the Gradle build process. It seems to stop around the :buildSrc tasks (compileKotlin/compileJava) and project configuration (common-db-identifiers). Could you take a look at the build logs to check if there are any missing dependencies or plugin configuration issues? Once the build passes, the other test jobs (Central Server, Security Server, and E2E tests) should run and give better feedback

Harini-royal12:
The latest commit triggered the CI, but the Build and test source step is still failing (Gradle 8.12 running tasks: build sonar test intTest jacocoTestReport). The logs show that the failure happens during the test phase, and the build scan wasn’t published. Could you please check the failing unit/integration tests or Gradle config to see what’s causing the build to stop? Once this step passes, the skipped jobs (Central Server, Security Server, and E2E tests) should run as well.

Harini-royal12:
Thanks for the CI feedback 🙏. I see that the Security Server system tests failed while other checks passed.
I’ll check the test logs to identify the cause and push a fix.
If there are known test environment specifics for the Security Server (e.g., configs or dependencies I should be aware of), any guidance would be appreciated.
I’ll update the PR once I resolve the failure.

Harini-royal12:
I checked the CI logs, and the failure comes from the Security Server system tests (JUnit) – Error: Failed test were found and 'fail-on-error' option is set to true.
I’ll review the detailed test report to identify which test case failed and attempt to reproduce it locally (gradle test in the Security Server module).
Once I confirm the root cause, I’ll push a fix.

If there are any project-specific steps for running the Security Server tests locally (e.g., configs, docker dependencies), please let me know so I can align my setup.

Harini-royal12:
Build and packaging are successful, but the Security Server system tests are failing (fail-on-error triggered). Could you please check the failing test cases and update the PR

Harini-royal12:
The pipeline failed because some Security Server system tests did not pass.
The error Failed test were found and 'fail-on-error' option is set to true means at least one JUnit test case failed.
To fix this, please review the detailed test report in the workflow logs, identify which test cases failed, and update either the code or the tests accordingly.
Alternatively, the fail-on-error flag in the test-reporter step could be set to false, but the recommended solution is to resolve the failing tests so the pipeline passes.

Harini-royal12:
The Central Server system tests and Security Server system tests are failing.
The error comes from the test-reporter step:
Failed test were found and 'fail-on-error' option is set to true.
This means some JUnit test cases did not pass.
To fix this, please check the detailed test report in the workflow logs, identify the failing test cases, and update the code or tests so they pass.
As an alternative (not recommended for production), the fail-on-error flag in the workflow could be set to false, but the correct approach is to fix the failing tests.

src/common/common-vault-spring/src/main/java/org/niis/xroad/common/vault/spring/SpringVaultClientConfig.java
```diff
@@ -0,0 +1,53 @@
...
+import org.springframework.context.annotation.Profile;
+import org.springframework.vault.core.VaultTemplate;
+
+import java.util.Optional;
+
+@Configuration
+public class SpringVaultClientConfig {
+
+    @Bean
+    @Profile("!test")
```
ricardas-buc, line 42: I'd rather avoid this test profile in production code ( I do know there few places with it already). It would be better to either have ConditionBean and some property or stubbing using mockBean annotations

sidecar/files/_entrypoint_common.sh
```diff
@@ -120,11 +120,7 @@ if dpkg --compare-versions "$CONFIG_VERSION" lt-nl "7.6.0"; then
   /usr/share/xroad/scripts/acme_contacts_and_keystore_pw_migra.sh
 fi
 
-# Generate internal and admin UI TLS keys and certificates if necessary
-if [ ! -f /etc/xroad/ssl/internal.crt ]; then
-  log "Generating new internal TLS key and certificate"
-  "$XROAD_SCRIPT_LOCATION/generate_certificate.sh" -n internal -f -S -p 2>&1 >/dev/null | sed 's/^/    /'
-fi
+# Generate admin UI TLS key and certificate if necessary
```
ricardas-buc, line 123: dangling comment?

andresrosenthal: Proxy UI cert is currently still generated with script, located two lines below

src/common/common-vault-spring/src/main/java/org/niis/xroad/common/vault/spring/SpringVaultClientConfig.java
andresrosenthal: I’ll keep this in mind for the next pull request, where I’ll be migrating the Admin UI TLS credentials into OpenBao.

PR #2947, opened 2025-08-28
chore: drop devices.ini in favour on centralised yaml property source

justasnortal:
Refs: XRDDEV-2799


src/service/signer/signer-core/src/main/java/org/niis/xroad/signer/core/config/ModuleProperties.java
```diff
@@ -0,0 +1,139 @@
...
+public interface ModuleProperties {
+
+    @WithName("enabled")
+    @WithDefault("true")
+    boolean enabled();
+
+    @WithName("library")
+    String library();
+
+    @WithName("library_cant_create_os_threads")
```
ricardas-buc, line 46: should be migrated to snake-case, but agreed that this will be done separately.

ricardas-buc (approved):

PR #2948, opened 2025-08-28
feat: As a Security Server Administrator I would like to be able to configure the ACME HTTP challenge port so that I can use a higher port and route to it

mikkbachmann:
refs: XRDDEV-2919


src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/config/AcmeConfig.java
```diff
@@ -60,7 +60,7 @@ public WebServerFactoryCustomizer<TomcatServletWebServerFactory> acmeChallengeCu
     private void acmeChallengeCustomizer(TomcatServletWebServerFactory factory) {
         var connector = new Connector(Http11NioProtocol.class.getName());
         connector.setScheme("http");
-        connector.setPort(80);
```
petkivim, line 63: Since the ACME challenge port can be customised now, it would be good to log it during startup so that it's possible to see from the logs what port is used.

doc/Sidecar/security_server_sidecar_user_guide.md
```diff
@@ -365,6 +366,8 @@ See [UG-SYSPAR](#Ref_UG-SYSPAR) for configuration details.
 Automated Certificate Management Environment (ACME) protocol enables partly automated certificate management of the authentication and sign
 certificates on the Security Server. More information about the required configuration is available in the [Security Server User Guide](../Manuals/ug-ss_x-road_6_security_server_user_guide.md#24-configuring-acme).
 
+For Sidecar, it is possible to use a different ACME challenge port from the default 80. For this, set the environment variable `XROAD_PROXY_UI_API_ACME_CHALLENGE_PORT` to the desired port number and map it to a host port 80 when starting the container (ACME Servers always use port 80 for HTTP-01 challenge).
```
petkivim, line 369: I think it would be good to either add the content of section "3.2 Enabling ACME Support" or at least a reference to it to the Sidecar Kubernetes user guide. Currently, the Kubernetes user guide doesn't include any information about ACME. And usually, problems with port 80 are related to running the Sidecar on Kubernetes.

andresrosenthal (approved):

petkivim (approved):

PR #2949, opened 2025-09-01
chore: Update exception handling over gRPC

ricardas-buc:
Refs: XRDDEV-2959

### DIFF
--- a/src/common/common-core/src/main/java/org/niis/xroad/common/core/exception/ErrorCode.java
+++ b/src/common/common-core/src/main/java/org/niis/xroad/common/core/exception/ErrorCode.java
@@ -258,7 +258,7 @@ public enum ErrorCode implements DeviationBuilder.ErrorDeviationBuilder {
 
     private static String prepareCode(String code) {
         if (code == null) {
-            log.warn("Error code is null, returning empty string");
+            log.warn("Error code is null, defaulting to {}", INTERNAL_ERROR.code());
             return INTERNAL_ERROR.code();
         }
         return code.toLowerCase();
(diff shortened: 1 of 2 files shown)


src/common/common-rpc/src/main/java/org/niis/xroad/common/rpc/server/RpcResponseHandler.java
```diff
@@ -0,0 +1,92 @@
...
+    public <T> void handleRequest(StreamObserver<T> responseObserver, Supplier<T> handler) {
+        try {
+            responseObserver.onNext(handler.get());
+            responseObserver.onCompleted();
+        } catch (Exception e) {
+            handleException(e, responseObserver);
+        }
+    }
+
+    @SuppressWarnings("checkstyle:MagicNumber")
```
justasnortal: seems not needed here.

src/common/common-core/src/main/java/org/niis/xroad/common/core/exception/ErrorCode.java
```diff
@@ -245,17 +245,16 @@ public String code() {
...
-                return errorCode;
-            }
-        }
-        return INTERNAL_ERROR; // Default fallback
+    public static DeviationBuilder.ErrorDeviationBuilder withCode(String code) {
+        return () -> code;
+    }
+
+    public DeviationBuilder.ErrorDeviationBuilder withPrefix(String prefix) {
+        return () -> prefix + "." + code;
```
justasnortal: maybe add prefix.toLowerCase(), to make sure prefix is lowercased?
there are mixed calls, like
TOKEN_NOT_FOUND.withPrefix("signer")
INTERNAL_ERROR.withPrefix("Signer")

PR #2960, opened 2025-09-09
docs: Documentation updates

petkivim:
- Add information about trust services to the CS user guide.

### DIFF
--- a/doc/Manuals/ug-cs_x-road_6_central_server_user_guide.md
+++ b/doc/Manuals/ug-cs_x-road_6_central_server_user_guide.md
@@ -1,6 +1,6 @@
 # X-Road: Central Server User Guide <!-- omit in toc -->
 
-Version: 2.49  
+Version: 2.50  
 Doc. ID: UG-CS
 
 ## Version history <!-- omit in toc -->
@@ -75,6 +75,7 @@ Doc. ID: UG-CS
 | 15.01.2025 | 2.47    | Minor updates                                                                                                                                                                                                                                                                                                                                                                                                                           | Petteri Kivimäki     |
 | 09.03.2025 | 2.48    | Naming/Renaming subsystems                                                                                                                                                                                                                                                                                                                                                                                                              | Ovidijus Narkevicius |
 | 21.03.2025 | 2.49    | Syntax and styling                                                                                                                                                                                                                                                                                                                                                                                                                      | Pauline Dimmek       |
+| 09.09.2025 | 2.50    | Add information about trust services                                                                                                                                                                                                                                                                                                                                                                                                    | Petteri Kivimäki     |
 
 ## Table of Contents <!-- omit in toc -->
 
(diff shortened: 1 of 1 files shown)


doc/Manuals/ug-cs_x-road_6_central_server_user_guide.md
```diff
@@ -1027,6 +1028,38 @@ To delete a global group, follow these steps.
 
 ## 10. Managing the Approved Certification Services
 
+A Certification Authority (CA) issues certificates to Security Servers (authentication certificates) and to X-Road member organizations (signing certificates). Authentication certificates secure the TLS connection between two Security Servers, while signing certificates are used to digitally sign messages sent by X-Road members. Only certificates issued by an approved CA, as defined on the Central Server by the X-Road operator, may be used. Information about approved CAs is distributed to the Security Servers via the global configuration. All approved CAs must comply with the [technical requirements](https://nordic-institute.atlassian.net/wiki/spaces/XRDKB/pages/16252930/What+Are+the+Technical+Requirements+for+Trust+Service+Providers#Certification-Authority) defined by X-Road.
```
raits, line 1031: I think we should migrate the technical requirements for the CA from the knowledge base to the core documentation since it is an important prerequisite for setting up the ecosystem and Central Server.

Perhaps we should consider having a separate document that covers the ecosystem requirements and general setup (linking to CS and SS documents).

With that said, I am fine with merging this change with the link to the knowledge base for now.

(This same comment applies to the OCSP requirements and TSA requirements in the following sections as well)

doc/Manuals/ug-cs_x-road_6_central_server_user_guide.md
```diff
@@ -1027,6 +1028,38 @@ To delete a global group, follow these steps.
...
+
+**Certificate Profile**
+
+Every CA uses a [certificate profile](https://nordic-institute.atlassian.net/wiki/spaces/XRDKB/pages/16449539/What+Is+a+Certificate+Profile) that defines how information is stored in the certificate — including which fields are required, which are optional, and what values are allowed. The information required by X-Road must be present in the certificates, although the specific fields used to store it may vary between CAs. The following information is mandatory in X-Road certificates:
```
raits, line 1049: We should also consider documenting the certificate profile in the core documentation rather than the knowledge base.

doc/Manuals/ug-cs_x-road_6_central_server_user_guide.md
petkivim: I agree, the ecosystem and technical requirements should be part of the official documentation and not stored separately in the Knowledge Base. I thought about that too when updating the documentation, but since that requires a little bit more work and restructuring the current documentation, let's include these links to the KB in the official documentation as the first step. The next step will be including the requirements directly in the official documentation.

doc/Manuals/ug-cs_x-road_6_central_server_user_guide.md
petkivim: Yes, should be part of the official documentation too. Let's include it when working with the other changes.

PR #2984, opened 2025-09-24
chore: SystemProperties refactoring

justasnortal:
Refs: XRDDEV-2974


src/common/common-acme/src/main/java/org/niis/xroad/common/acme/AcmeConfig.java
```diff
@@ -0,0 +1,72 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.niis.xroad.common.acme;
+
+public interface AcmeConfig {
+
+    // raw usage with default in org.niis.xroad.securityserver.restapi.config.AcmeBeanConfig.IsAcmeCertRenewalJobsActive
+    boolean isAcmeRenewalActive();
```
ricardas-buc, line 35: perhaps  use javadoc comments?

src/common/common-properties/build.gradle.kts
```diff
@@ -6,7 +6,7 @@ plugins {
 
 dependencies {
   implementation(libs.slf4j.api)
-  implementation(libs.microprofile.config.api)
+  implementation(libs.smallrye.config.core)
```
ricardas-buc, line 9: that might introduce jboss logmanager into the classpath. Are there any errors in tests or similar?

src/security-server/system-test/src/intTest/resources/compose.systemtest.yaml
```diff
@@ -49,7 +51,7 @@ services:
       - ../../../src/security-server/system-test/src/intTest/resources/container-files/etc/xroad/backup-keys:/etc/xroad/backup-keys
 
   db-opmonitor:
-    image: postgres:17
+    image: postgres:16
```
ricardas-buc: any particular reason to downgrade?

src/common/common-properties/build.gradle.kts
justasnortal: Needed for the 
@ConfigMapping(prefix = "xroad.common")
public interface CommonProperties { 
....

src/security-server/system-test/src/intTest/resources/compose.systemtest.yaml
justasnortal: To match with pg_dump version used in backup-manager application. ubuntu 24 has postgresql v16.

PR #3001, opened 2025-10-06
feat: add the central server connection test to diagnostics page

enelir:
Refs: XRDDEV-2922

### DIFF
--- a/src/common/common-domain/src/main/resources/request.xsd
+++ b/src/common/common-domain/src/main/resources/request.xsd
@@ -43,6 +43,14 @@
                     </documentation>
                 </annotation>
             </element>
+            <element name="dryRun" type="boolean" minOccurs="0" default="false">
+                <annotation>
+                    <documentation>
+                        Optional flag indicating whether the request should
+                        be validated and processed as a dry run only.
+                    </documentation>
+                </annotation>
+            </element>
             <element name="requestId" type="tns:RequestIdType" minOccurs="0"/>
         </sequence>
     </complexType>
--- a/src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/openapi/DiagnosticsApiController.java
+++ b/src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/openapi/DiagnosticsApiController.java
@@ -40,6 +40,7 @@ import org.niis.xroad.restapi.converter.ClientIdConverter;
 import org.niis.xroad.restapi.converter.ServiceIdConverter;
 import org.niis.xroad.restapi.openapi.ControllerUtil;
 import org.niis.xroad.securityserver.restapi.converter.AddOnStatusConverter;
+import org.niis.xroad.securityserver.restapi.converter.AuthCertStatusConverter;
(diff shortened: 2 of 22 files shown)


src/common/common-domain/src/main/resources/request.xsd
```diff
@@ -43,6 +43,14 @@
                     </documentation>
                 </annotation>
             </element>
+            <element name="dryRun" type="boolean" minOccurs="0" default="false">
+                <annotation>
+                    <documentation>
+                        Optional flag indicating whether the request should
+                        be validated and processed as a dry run only.
```
mikkbachmann: further down one code comment says about dryRun: "if true, the request is only validated by the central server, not processed", here it says "whether the request should be validated and processed as a dry run only", the wording was a bit confusing, when reading them. Is it meant hear: "Optional flag indicating whether the request should be validated only and not processed like a real request"?

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/DiagnosticConnectionService.java
```diff
@@ -0,0 +1,230 @@
...
+            }
+        }
+    }
+
+    public ConnectionStatus getAuthCertRegStatusInfo() {
+        CertValidation certValidation = validateAuthCert();
+        try {
+            // the error is expected, but we want to verify that the connection can be established
+            managementRequestSenderService.sendAuthCertRegisterRequest(null, new byte[0], true);
+            throw new IllegalStateException("should not get here");
```
mikkbachmann, line 140: this throw feels misplaced here. sendAuthCertRegisterRequest is always expected to throw exception? If yes, then I think the method name could reflect this better - is it so that this method doesn't actually get auth cert registration status info, but rather checks connection?

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/openapi/DiagnosticsApiController.java
```diff
@@ -181,6 +189,19 @@ public ResponseEntity<List<OperationalDataIntervalDto>> getOperationalDataInterv
         return new ResponseEntity<>(operationalInfoConverter.convert(opDataIntervals), HttpStatus.OK);
     }
 
+    @Override
+    @PreAuthorize("hasAuthority('DIAGNOSTICS')")
+    public ResponseEntity<ConnectionStatusDto> getAuthCertReqStatus() {
+        return new ResponseEntity<>(authCertStatusConverter.convert(diagnosticConnectionService.getAuthCertRegStatusInfo()), HttpStatus.OK);
```
mikkbachmann: Method name has req with a "q", but calls a method with reg with a "g". Is this correct?

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/DiagnosticConnectionService.java
enelir: Not only connection, there are other verification as well

PR #3029, opened 2025-10-14
feat: Add support for multiple tokens to the autologin script

raits:
This PR:

- Updates the autologin script to accept pins for multiple tokens
- Adds environmental variables for configuring multiple tokens in the Sidecar and development Docker images
- Keeps the old behavior for existing users
- Migrates `xrd-dev-stack` to use the autologin script instead of Hurl so we have at least some test coverage
- Fixes Sidecar issue where autologin reports success but does not actually log the token in if the token is still uninitialised


Docker/xrd-dev-stack/compose.dev.yaml
```diff
@@ -36,7 +36,8 @@ services:
   ss1:
     container_name: ss1
     environment:
-      - XROAD_TOKEN_PIN=Secret1234
+      - XROAD_TOKEN_0_PIN=Secret1234
+      - XROAD_TOKEN_31_PIN=Secret1234
```
mikkbachmann, line 40: is this XROAD_TOKEN_31_PIN and not XROAD_TOKEN_1_PIN on purpose or by accident?

Docker/xrd-dev-stack/compose.dev.yaml
```diff
@@ -91,4 +92,4 @@ networks:
   # Use implicitly named network so that is easier to add container outside the compose
   xroad-network:
     name: xroad-network
-    driver: bridge
```
mikkbachmann, line 94: this is a peculiar diff, maybe worth checking if there wasn't any unwanted invisible line end character changes. Same is at the end of init-token-and-run-entrypoint.sh and Docker/securityserver/Dockerfile

Docker/xrd-dev-stack/compose.dev.yaml
raits: It's on purpose, the way that the hardware token ID-s work is that it takes the token id (which we have hard-coded to `1` via the `token_id_format` option) as characters and converts them to their hex values. And the hex value for the ASCII character `1` is `31`.

Docker/xrd-dev-stack/compose.dev.yaml
raits: Good point, will have a look.

raits: Looks like it is just the `\n` character added to the end with the formater, should be ok for this case. From what I can find this is considered best practice due to the [POSIX definition of a line](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206).

doc/Sidecar/kubernetes_security_server_sidecar_user_guide.md
```diff
@@ -289,6 +289,7 @@ For example the following configuration could be stored as a Kubernetes secret:
 * Sensitive Sidecar environment variables:
```
petkivim, line 290: The document version number and version history table should be updated.

PR #3078, opened 2025-11-05
feat: As a Central Server Administrator I want to have the possibility to define which CSR format is used for CA-s so that user wouldn't need to look the information up

mikkbachmann:
refs: XRDDEV-2977


src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/service/CertificationServicesServiceImpl.java
```diff
@@ -177,6 +180,9 @@ public CertificationService update(CertificationService approvedCa) {
                 .ifPresent(persistedApprovedCa::setAuthCertProfileId);
         Optional.ofNullable(approvedCa.getSigningCertificateProfileId())
                 .ifPresent(persistedApprovedCa::setSignCertProfileId);
+        Optional.ofNullable(approvedCa.getDefaultCsrFormat())
+                .map(Enum::name)
+                .ifPresent(persistedApprovedCa::setDefaultCsrFormat);
         final ApprovedCaEntity updatedApprovedCa = approvedCaRepository.save(persistedApprovedCa);
         addAuditData(updatedApprovedCa);
```
enelir, line 188: in audit log need also add defaultCsrFormat?

enelir (changes requested):

PR #3080, opened 2025-11-05
fix: use subsystem name as label where subsystem name is displayed in…

toomy4:
…stead of member name

refs: XRDDEV-3033


src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/converter/comparator/ServiceClientSortingComparator.java
```diff
@@ -38,7 +38,7 @@
 public class ServiceClientSortingComparator implements Comparator<ServiceClientDto> {
 
     /**
-     * Compare ServiceClient objects using member name / group description as the primary sort key, and client id
+     * Compare ServiceClient objects using subsystem name / group description as the primary sort key, and client id
```
enelir, line 41: prb here is better "client name"

src/security-server/admin-service/ui/src/locales/en.json
```diff
@@ -5,7 +5,7 @@
     "addServiceClientsTitle": "Add Subjects",
     "addSubjectsSuccess": "Access rights added successfully",
     "id": "ID",
-    "memberName": "Member Name / Group Description",
+    "memberName": "Subsystem Name / Group Description",
```
enelir: can here also rename  "name": "Subsystem Name / Group Description"?

src/security-server/admin-service/ui/src/locales/en.json
```diff
@@ -854,7 +854,7 @@
     "enableSuccess": "Service description enabled",
     "idGroupCode": "ID / Group code",
     "lastRefreshed": "Last refreshed: ",
-    "memberNameGroupDesc": "Member name / Group description",
+    "memberNameGroupDesc": "Subsystem name / Group description",
```
enelir: "subsystemNameGroupDesc"

PR #3090, opened 2025-11-10
XRDDEV-3030

ricardas-buc:
Migrate to quarkus based property source and move all the messagelog properties to proxy

Refs: XRDDEV-3030

### DIFF
--- a/src/security-server/system-test/src/intTest/resources/container-files/etc/xroad/conf.d/local.yaml
+++ b/src/security-server/system-test/src/intTest/resources/container-files/etc/xroad/conf.d/local.yaml
@@ -23,7 +23,7 @@ xroad:
         messagelog-keystore-password: somepassword
         messagelog-key-id: key1
       archiver:
-        encryption-enabled: false
+        encryption-enabled: true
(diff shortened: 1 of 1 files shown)


ricardas-buc:
[rerun-ci]

src/security-server/admin-service/management-rpc-client/src/main/java/org/niis/xroad/management/rpc/ManagementRpcClient.java
```diff
@@ -0,0 +1,88 @@
...
+
+    public MessageLogArchivalResp triggerArchival(MessageLogArchivalRequest request) {
+        return exec(() -> archiverServiceBlockingStub
+                .triggerArchival(request));
+    }
+
+    public MessageLogCleanupResp triggerCleanup(MessageLogCleanupRequest request) {
+        return exec(() -> archiverServiceBlockingStub
+                .triggerCleanup(request));
+    }
```
justasnortal: Archival and cleanup might take time, probably it should be async calls.

PR #3113, opened 2025-11-21
refactor: SystemProperties refactor

ovidijusnortal:
Refs: XRDDEV-3029


src/security-server/admin-service/application/src/main/resources/META-INF/spring.factories
```diff
@@ -1 +1 @@
-org.springframework.boot.env.EnvironmentPostProcessor=org.niis.xroad.securityserver.restapi.config.LegacySystemPropertyEnvironmentPostProcessor
```
justasnortal, line 1: Maybe delete the file if it is empty?

PR #3120, opened 2025-11-25
feat: As a Security Server Administrator I want the data exchange to account for the TSA and OCSP strategy I have selected so that I can manage cost

mikkbachmann:
refs: XRDDEV-2981


src/common/common-domain/src/main/java/org/niis/xroad/common/CostTypeSorter.java
```diff
@@ -0,0 +1,81 @@
...
+                    if (CostType.PAID.equals(tbs.getCostType())) {
+                        this.paidValues.add(sortableValue);
+                    }
+                    if (tbs.getCostType() == null || CostType.UNDEFINED.equals(tbs.getCostType())) {
+                        this.undefinedCostValues.add(sortableValue);
+                    }
+                });
+    }
+
+    public List<String> sort(SystemProperties.ServicePrioritizationStrategy prioritizationStrategy) {
```
mloitm: This method is also filtering, so method name is a bit misleading. Maybe `filterSort` or `select` would describe it better? 
This also applies to class name.

src/lib/globalconf-core/src/main/resources/globalconf/v6/shared-parameters.xsd
```diff
@@ -293,7 +293,7 @@
         <annotation>
             <documentation>Describes whether the service is free or paid</documentation>
             <appinfo>
-                <jxb:javaType name="org.niis.xroad.globalconf.model.CostType" parseMethod="valueOf" printMethod="name"/>
+                <jxb:javaType name="org.niis.xroad.common.CostType" parseMethod="valueOf" printMethod="name"/>
```
mloitm, line 296: This wasn't added in this PR, but bindings should go into separate file as can be seen in `identifiers-bindings.xml`. Schema should be implementation agnostic.

src/lib/globalconf-impl/src/test/java/org/niis/xroad/globalconf/impl/GlobalConfVer6Test.java
```diff
@@ -106,20 +107,77 @@ private static void deleteConfigurationFiles() {
...
+        assertEquals(2, addresses.size());
+        assertEquals("http://127.0.0.1:8082/ocsp", addresses.get(0));
+        assertEquals("http://www.example.net/ocsp4", addresses.get(1));
+
+        System.setProperty(SystemProperties.PROXY_OCSP_PRIORITIZATION_STRATEGY,
+                SystemProperties.ServicePrioritizationStrategy.FREE_FIRST.name());
+        addresses =
+                globalConfProvider.getOrderedOcspResponderAddresses(TestCertUtil.getCertChainCert("user_1.p12"));
+
+        assertEquals(5, addresses.size());
```
mloitm: These lists could be asserted better with assertj-s containsExactly

src/lib/globalconf-impl/src/main/java/org/niis/xroad/globalconf/impl/GlobalConfImpl.java
```diff
@@ -287,27 +291,59 @@ public ClientId.Conf getSubjectName(SignCertificateProfileInfo.Parameters parame
...
+        OptionalInt gcVersion = getVersion();
+        boolean globalConfSupportsCostTypes = gcVersion.isPresent() && gcVersion.getAsInt() >= GLOBAL_CONF_VERSION_WITH_COST_TYPE;
+
+        if (globalConfSupportsCostTypes) {
+            CostTypeSorter<SharedParameters.OcspInfo> sorter = new CostTypeSorter<>(responders);
+            SystemProperties.ServicePrioritizationStrategy prioritizationStrategy = SystemProperties.getOcspPrioritizationStrategy();
+            log.debug("OCSP responder urls will be sorted based on prioritization strategy: {}", prioritizationStrategy);
+            return sorter.sort(prioritizationStrategy);
+        } else {
+            return responders.stream().map(SharedParameters.OcspInfo::getUrl).collect(Collectors.toList());
```
mloitm: One way to address the sonar warning on this line would be returning Stream instead of list from this method. Then in the calling mehtod, member ocsp responder could be appended with `Stream.concat(..., Optional.ofNullable(uri).stream())`

src/common/common-domain/src/main/java/org/niis/xroad/common/CostTypeSorter.java
mloitm: Maybe it would better to split filtering and sorting into separate methods, so they could be used directly on stream and grouping in divide() would not be necessary.

mloitm (approved):

PR #3126, opened 2025-11-28
feat: better internal ip resolution for opmonitoring data processor

toomy4:
Refs: XRDDEV-2734


src/addons/op-monitoring/src/main/java/org/niis/xroad/proxy/core/opmonitoring/OpMonitoringDataProcessor.java
```diff
@@ -96,4 +100,27 @@ private static boolean isNonLoopback(NetworkInterface ni) {
...
+
+    private static boolean isUsableAddress(InetAddress addr) {
+        return !addr.isLoopbackAddress()
+                && !addr.isLinkLocalAddress()
+                && addr.getHostAddress() != null;
+    }
+
+    private static int compareInetAddresses(InetAddress a1, InetAddress a2) {
+        // prefer IPv4 addresses
+        return Boolean.compare(a1.getHostAddress().contains(":"), a2.getHostAddress().contains(":"));
```
mloitm: Maybe using instanceof check instead of string search would be more reliable?

```suggestion
        return Boolean.compare(a1 instanceof Inet6Address, a2 instanceof Inet6Address);
```

PR #3133, opened 2025-12-02
fix: change default_csr_format column to be nullable and required in api

MarcDavid155:
Refs: XRDDEV-3046

### DIFF
--- a/src/central-server/openapi-model/src/main/resources/openapi-definition.yaml
+++ b/src/central-server/openapi-model/src/main/resources/openapi-definition.yaml
@@ -3576,8 +3576,6 @@ components:
           format: text
           type: string
       type: object
-      required:
-        - default_csr_format
     DistinguishedName:
       type: object
       properties:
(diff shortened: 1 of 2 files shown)


src/central-server/admin-service/infra-jpa/src/main/resources/liquibase/centerui/202511271200-ca-csr-nullable.xml
raits, line 1: Since this is a change to an unreleased version, I wonder if it would make more sense to just modify the original changelog file. This change will otherwise look strange in the actual production version, since we will have a changelog file that adds the column as not nullable and then in the same version upgrade do this change to make it nullable.

It will potentially create some issues in the future (e.g. upgrading the open beta environment we use for the security audit) and potentially for local development environments, but I feel would make the resulting software package clearer.

src/central-server/openapi-model/src/main/resources/openapi-definition.yaml
```diff
@@ -3547,6 +3547,9 @@ components:
         - $ref: '#/components/schemas/CertificationServiceSettings'
       description: certification service file and settings
       type: object
+      required:
+        - certificate
+        - default_csr_format
```
raits, line 3552: Lets make sure to add this to the implementation notes field of the ticket so that we remember to mention this in the release notes as it is a breaking change to the Central Server API.

src/central-server/admin-service/infra-jpa/src/main/resources/liquibase/centerui/202511271200-ca-csr-nullable.xml
MarcDavid155: I thought about potentially doing it that way as well, however, liquibase themselves recommend not changing any already commited changelogs and instead do a "roll-forward", which means adding a new changelog to achieve the same result - https://www.liquibase.com/blog/dealing-with-changing-changesets

Since liquibase's update will fail if the database already has a changed changelog applied to it, this seems like the safest way to revert this non-nullable requirement.

PR #3140, opened 2025-12-04
fix: an issue where encoded path traversal could bypass service ACL c…

ricardas-buc:
…heck

Refs: XRDDEV-3054

### DIFF
--- a/src/common/common-core/src/test/java/ee/ria/xroad/common/util/UriUtilsTest.java
+++ b/src/common/common-core/src/test/java/ee/ria/xroad/common/util/UriUtilsTest.java
@@ -98,6 +98,7 @@ public class UriUtilsTest {
         assertEquals("/a/b/c", UriUtils.decodeAndNormalize("/a/b/%2e/c"));
         assertEquals("/a/b/c", UriUtils.decodeAndNormalize("/a/b/d/%2e%2e/c"));
         assertEquals("/c", UriUtils.decodeAndNormalize("/a/b/%2e%2e/%2e%2e/c"));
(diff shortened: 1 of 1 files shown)


src/service/proxy/proxy-core/src/main/java/org/niis/xroad/proxy/core/util/MessageProcessorBase.java
```diff
@@ -172,7 +172,7 @@ protected void updateOpMonitoringDataByRestRequest(OpMonitoringData opMonitoring
     }
 
     private String getNormalizedServicePath(String servicePath) {
-        return Optional.of(UriUtils.uriPathPercentDecode(URI.create(servicePath).normalize().getRawPath(), true))
+        return Optional.of(UriUtils.decodeAndNormalize(servicePath))
```
justasnortal: Seems there is no point of Optional here.

src/common/common-core/src/main/java/ee/ria/xroad/common/util/UriUtils.java
```diff
@@ -35,6 +38,16 @@ public final class UriUtils {
     private UriUtils() {
     }
 
+    /**
+     * Percent-decodes and normalizes a URI path, assuming UTF-8 character set.
+     *
+     * @see #uriPathPercentDecode(String, boolean)
+     */
+    public static String decodeAndNormalize(@Nonnull final String path) {
```
justasnortal, line 46: Maybe add unit tests?

justasnortal (approved):

PR #3158, opened 2025-12-08
feat: Introduce horizontal scalability for softtoken signing

andresrosenthal:


src/service/softtoken-signer/softtoken-signer-application/src/main/java/org/niis/xroad/signer/softtoken/config/SoftTokenSignerKeysProperties.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Configuration properties for softtoken-signer keys' synchronization from signer.
+ */
+@ConfigMapping(prefix = "xroad.softtoken-signer.keys")
+public interface SoftTokenSignerKeysProperties {
+
+    @WithName("sync-rate")
+    @WithDefault("30")
+    int refreshRateInSeconds();
```
ricardas-buc: Use Duration. There is a common pattern for quarkus props.  In this case default would be 30S

src/service/softtoken-signer/src/main/java/org/niis/xroad/signer/softtoken/sync/KeysSynchronizationJob.java
```diff
@@ -0,0 +1,154 @@
...
+                keyCache.removeKey(keyId);
+                removedCount++;
+            }
+
+            log.info("Successfully synchronized software token keys (added: {}, updated: {}, unchanged: {}, removed: {})",
+                    addedCount, updatedCount, unchangedCount, removedCount);
+
+        } catch (Exception e) {
+            log.error("Software token key synchronization failed: {}", e.getMessage(), e);
+            throw XrdRuntimeException.systemInternalError("Key synchronization failed", e);
```
ricardas-buc: what happens if it fails? Is it an outage for the signer?

When on the same subject, perhaps this should relate to k8 healtchecks?

src/service/softtoken-signer/src/main/java/org/niis/xroad/signer/softtoken/config/SofttokenSignerKeysProperties.java
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * Configuration properties for softtoken-signer keys' synchronization from signer.
+ */
+@ConfigMapping(prefix = "xroad.softtoken-signer.keys")
+public interface SofttokenSignerKeysProperties {
+
+    @WithName("refresh-rate")
+    @WithDefault("30")
+    int refreshRateInSeconds();
```
ricardas-buc: should be Duration as well

src/service/softtoken-signer/softtoken-signer-int-test/src/intTest/resources/softtoken-signer-container-files/Dockerfile
```diff
@@ -0,0 +1,13 @@
+ARG SOFTTOKEN_SIGNER_IMG
+
+FROM $SOFTTOKEN_SIGNER_IMG
+
+USER root
+
+RUN apt-get -qq update \
+  && apt-get -qq install --no-install-recommends curl \
```
ricardas-buc: cant we avoid curl usage here?

src/service/softtoken-signer/softtoken-signer-application/src/main/java/org/niis/xroad/signer/softtoken/sync/SoftwareTokenKeyCache.java
```diff
@@ -0,0 +1,71 @@
...
+ *   <li>Read operations (during signing): Lock-free via volatile map reference</li>
+ *   <li>Write operations (during sync): Atomic map replacement, no read blocking</li>
+ *   <li>Thread safety: ConcurrentHashMap handles concurrent reads</li>
+ * </ul>
+ */
+@Slf4j
+@ApplicationScoped
+public class SoftwareTokenKeyCache {
+
+    private volatile Map<String, CachedKeyInfo> keys = new ConcurrentHashMap<>();
```
ricardas-buc: I dont like this pattern. We check and update at an item level, but we've already have all the items computed. A simpler approach would be just to swap out whole map here.  Something like:
```  
private final AtomicReference<Map<String, CachedKeyInfo>> keys = new AtomicReference<>(Map.of());

    public void updateKeys(Map<String, CachedKeyInfo> newKeys) {
        keys.set(Map.copyOf(newKeys));
    }
```

src/service/softtoken-signer/softtoken-signer-application/src/main/java/org/niis/xroad/signer/softtoken/sync/KeysSynchronizationJob.java
```diff
@@ -0,0 +1,153 @@
...
+    public void destroy() {
+        scheduler.shutdown();
+    }
+
+    /**
+     * Synchronizes software token keys from the signer service.
+     * <p>
+     * This method:
+     * <ul>
+     *   <li>Fetches all software token keys from the signer via gRPC</li>
```
ricardas-buc: instead of relying on javadoc, lets refactor so that each step would be a separate method.

PR #3171, opened 2025-12-11
fix: Incorrect CSR generation for ACME when using the PEM format

mikkbachmann:
refs: XRDDEV-3062


src/common/common-core/src/main/java/ee/ria/xroad/common/util/CertUtils.java
```diff
@@ -532,4 +534,20 @@ public static void createPkcs12(String filenameKey, byte[] certBytes, String fil
     public static String identify(X509Certificate certificate) {
         return certificate.getIssuerX500Principal() + " " + certificate.getSerialNumber();
     }
+
+    public static byte[] convertPemCsrToDer(byte[] pemCsr) {
+        try {
+            String pemString = new String(pemCsr, StandardCharsets.UTF_8);
+            String base64Content = pemString
```
toomy4: wouldn't bouncycastle's org.bouncycastle.openssl.PEMParser be a more reliable option for this?

PR #3185, opened 2025-12-15
feat: As a Security Expert I want that the ACME account keystore password is generated automatically so that the implementation is more secure.

mikkbachmann:
refs: XRDDEV-3064


doc/Manuals/ug-ss_x-road_6_security_server_user_guide.md
```diff
@@ -3605,7 +3606,7 @@ This parameter can be overridden by an environment variable `XROAD_PROXY_UI_API_
...
 
 1. Credentials (kid and hmac secret) for external account binding. Some CAs require these for added security. They tie the X-Road member to an external account on the Certificate Authority's side and so need to be acquired externally from the CA.
-2. `account-keystore-password` -  a password of the ACME Server account PKCS #12 keystore that is populated automatically by the Security Server, when communicating with the ACME Server.
+2. `account-keystore-password` -  a password of the ACME Server account PKCS #12 keystore that is populated automatically by the Security Server, when communicating with the ACME Server. When ACME is used for the first time, then the keystore is generated automatically with this password. If this field is left empty, a random password is generated by the Security Server and added to the `acme.yml` file.
```
petkivim: Consider updating this a little bit:

> `account-keystore-password` - the password for the ACME Server account PKCS #12 keystore. The password is populated automatically by the Security Server when communicating with the ACME Server. When ACME is used for the first time, the keystore is generated automatically using this password. If the value of this property is left empty, the Security Server generates a random password and stores it in the `acme.yml` file. If the value of this property is not empty, the provided value is used as the password for the generated keystore file.

petkivim (approved):

PR #3195, opened 2025-12-18
feat: Add xroad-addon-hwtokens to security server sidecar image

MarcDavid155:
Refs: XRDDEV-3048

### DIFF
--- a/doc/Sidecar/security_server_sidecar_user_guide.md
+++ b/doc/Sidecar/security_server_sidecar_user_guide.md
@@ -380,9 +380,8 @@ The memory allocation for the Proxy Service can be configured using helper scrip
 Security Server Sidecar provides built-in support for hardware security tokens, requiring only configuration.
 
 1. Make the PKCS\#11 provider library (and any required additional files) available in the sidecar container by mounting them as volumes.
-2. Add the path to the PKCS\#11 library to the file `/etc/xroad/devices.ini` in the sidecar container.
+2. Create the `devices.ini` file under your mount for `/etc/xroad` and add the path for the PKCS\#11 library inside it (note that the library path should match the path inside the container you chose in the last step).
     * More information on how to configure the `devices.ini` file itself can be found in the [Security Server Installation Guide](../Manuals/ig-ss_x-road_6_security_server_installation_guide.md#210-installing-the-support-for-hardware-tokens).
-    * It is recommended to also mount the configured `devices.ini` file to avoid having to manually edit it inside the container.
 3. Restarting the Security Service Sidecar container might be required:
 
     ```bash


raits:
@MarcDavid155

Can the driver itself be provided via a volume instead so that they wouldn't need to change the container itself or does it need to be installed via the package manager?

raits (changes requested):
Re-requesting related to the general comment I made.

doc/Sidecar/security_server_sidecar_user_guide.md
```diff
@@ -373,7 +375,21 @@ For Sidecar, it is possible to use a different ACME challenge port from the defa
 
 The memory allocation for the Proxy Service can be configured using helper script `/usr/share/xroad/scripts/proxy_memory_helper.sh`. More information about the usage of this script is available in the [Security Server User Guide](../Manuals/ug-ss_x-road_6_security_server_user_guide.md#211-updating-proxy-services-memory-allocation-command-line-arguments).
 
-### 3.4 Autologin
+### 3.4 Installing Support for Hardware Tokens
+
+Security Server Sidecar provides built-in support for hardware security tokens, requiring only configuration.
+
+1. Make the PKCS\#11 provider library (and any required additional files) available in the sidecar container by mounting them as volumes.
+2. Add the path to the PKCS\#11 library to the file `/etc/xroad/devices.ini` in the sidecar container.
```
raits:
```suggestion
2. Create the `devices.ini` file under your mount for `/etc/xroad` and add the path for the PKCS\#11 library inside it (note that the library path should match the path inside the container you chose in the last step).
```

doc/Sidecar/security_server_sidecar_user_guide.md
```diff
@@ -373,7 +375,21 @@ For Sidecar, it is possible to use a different ACME challenge port from the defa
...
+    * More information on how to configure the `devices.ini` file itself can be found in the [Security Server Installation Guide](../Manuals/ig-ss_x-road_6_security_server_installation_guide.md#210-installing-the-support-for-hardware-tokens).
+    * It is recommended to also mount the configured `devices.ini` file to avoid having to manually edit it inside the container.
```
raits: This could probably be removed, as we already suggest mounting `/etc/xroad` in general.

raits (changes requested):
A few suggestions related to configuration instructions.

doc/Sidecar/security_server_sidecar_user_guide.md
raits: For some reason it picked the whole block, my comment refers to just removing the "* It is recommended to also mount the ...." part as that is already a general recommendation.

raits (approved):
Looks good to me.

PR #3211, opened 2025-12-29
feat: Update autologin retry logic from script to service

MarcDavid155:
Refs: XRDDEV-3061

### DIFF
--- a/src/autologin/source/redhat/xroad-autologin.service
+++ b/src/autologin/source/redhat/xroad-autologin.service
@@ -10,7 +10,7 @@ ExecStart=/usr/share/xroad/autologin/xroad-autologin-retry.sh
 StandardOutput=journal+console
 
 Restart=on-failure
-RestartSec=10s
+RestartSec=3s
 RestartPreventExitStatus=127
 
 [Install]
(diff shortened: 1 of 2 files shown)


src/autologin/source/redhat/xroad-autologin.service
```diff
@@ -9,6 +9,10 @@ Group=xroad
 ExecStart=/usr/share/xroad/autologin/xroad-autologin-retry.sh
 StandardOutput=journal+console
 
+Restart=on-failure
+RestartSec=10s
```
justasnortal: It think it is better to keep the 3 seconds like it was in the script. unless there is a reason to change.

src/packages/src/xroad/ubuntu/generic/xroad-autologin.service
```diff
@@ -9,6 +9,10 @@ Group=xroad
 ExecStart=/usr/share/xroad/autologin/xroad-autologin-retry.sh
 StandardOutput=journal+console
 
+Restart=on-failure
+RestartSec=10s
```
justasnortal: Lets keep the 3 sec, like in the script.

PR #3222, opened 2026-01-07
fix: restore security server token login in hurl script

enelir:


enelir:
Will be fixed here https://github.com/nordic-institute/X-Road/pull/3233

PR #3226, opened 2026-01-08
Xrddev 3079 fix(cs): Include subsystem used servers in Member Details view 

egidijusmwork:


src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/service/MemberServiceImpl.java
```diff
@@ -158,12 +157,19 @@ public List<SecurityServer> getMemberOwnedServers(ClientId memberId) {
...
-                .map(ServerClientEntity::getSecurityServer)
-                .map(securityServerMapper::toTarget)
-                .collect(toList());
+        return xRoadMemberRepository.findOneBy(memberId).map(member -> {
+            var memberServers = member.getServerClients().stream();
+            var subsystemServers = member.getSubsystems().stream()
+                    .map(SecurityServerClientEntity::getServerClients)
+                    .flatMap(Collection::stream);
+
+            return java.util.stream.Stream.concat(memberServers, subsystemServers)
```
ovidijusnortal: Why not just Stream.concat(...)?

src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/service/MemberServiceImpl.java
```diff
@@ -158,12 +157,19 @@ public List<SecurityServer> getMemberOwnedServers(ClientId memberId) {
...
+                    .map(ServerClientEntity::getSecurityServer)
+                    .distinct()
+                    .map(securityServerMapper::toTarget)
+                    .collect(toList());
```
ovidijusnortal: in java 21  Stream has built-in method for toList

src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/service/MemberServiceImpl.java
```diff
@@ -158,12 +157,19 @@ public List<SecurityServer> getMemberOwnedServers(ClientId memberId) {
...
+        })
+                .orElse(List.of());
```
ovidijusnortal: in this case maybe nothing to gain, but this could be done also like `orElseGet(List::of)` to get lazy evaluation

ovidijusnortal (approved):

PR #3238, opened 2026-01-14
feat: Add input validations to central and security servers descripti…

MarcDavid155:
…on fields

Refs: XRDDEV-3072


src/security-server/admin-service/ui/src/plugins/vee-validate.ts
```diff
@@ -132,6 +132,18 @@ export function createValidators() {
           return true;
         },
       );
+      defineRule(
+        'validDescription',
```
ovidijusnortal: maybe it would be good idea validation function and error message move into shared-ui module, to avoid duplication, also same thing can be done for already existing  'address' validation above

PR #3243, opened 2026-01-15
feat: block OPTIONS HTTP method XRDDEV-3075

egidijusmwork:


src/common/common-admin-api/src/main/java/org/niis/xroad/restapi/auth/securityconfigurer/ManageApiKeysWebSecurityConfig.java
```diff
@@ -51,45 +52,44 @@
...
-    /**
-     * Check if an alive session exists
-     */
-    private static boolean sessionExists(HttpServletRequest request) {
-        return request.getSession(false) != null;
-    }
+        /**
+         * Check if an alive session exists
+         */
+        private static boolean sessionExists(HttpServletRequest request) {
```
ricardas-buc: seems that formatting is off, is .editorconfig working/enabled in intellij?

src/service/proxy/proxy-core/src/main/java/org/niis/xroad/proxy/core/clientproxy/ClientRestMessageHandler.java
```diff
@@ -93,16 +96,23 @@ public ClientRestMessageHandler(MessageProcessorFactory messageProcessorFactory,
...
         if (target != null && target.startsWith("/r" + RestMessage.PROTOCOL_VERSION + "/")) {
-            verifyCanProcess();
+            verifyCanProcess(request);
             return messageProcessorFactory.createClientRestMessageProcessor(request, response, opMonitoringData);
         }
         return null;
     }
 
-    private void verifyCanProcess() {
+    private void verifyCanProcess(RequestWrapper request) {
```
ricardas-buc: this is xroad proxy, component, I am not sure that OPTIONS removal is expected here. We should explicitly confirm that.

src/service/proxy/proxy-core/src/test/java/org/niis/xroad/proxy/core/clientproxy/ClientRestMessageHandlerTest.java
```diff
@@ -0,0 +1,61 @@
+package org.niis.xroad.proxy.core.clientproxy;
```
ricardas-buc: licenses are missing. you can generate em over gradle commands or just copy/paste from another file.

PR #3258, opened 2026-01-21
feat: Move to OpenBao-based password store and internalize signer auto-login

andresrosenthal:


src/lib/vault-spring/src/main/java/org/niis/xroad/common/vault/spring/SpringVaultClient.java
```diff
@@ -139,6 +139,21 @@ public Map<String, String> getMLogDBEncryptionSecretKeys() {
         );
     }
 
+    @Override
```
ricardas-buc, line 142: Perhaps its worth separating into a separate vault client?

src/service/signer/signer-core/src/main/java/org/niis/xroad/signer/core/tokenmanager/module/ModuleManager.java
```diff
@@ -69,6 +74,7 @@ public void start() {
         } catch (Exception e) {
             log.error("Failed to initialize token worker!", e);
         }
+        autoLoginExecutor.execute(autoLoginService::execute);
```
ricardas-buc: if start fails, should it try loggiging in?

src/service/signer/signer-core/src/main/java/org/niis/xroad/signer/core/tokenmanager/token/AbstractTokenWorker.java
```diff
@@ -76,25 +77,27 @@ public abstract class AbstractTokenWorker extends SignReqHandler implements Toke
...
 
     @Override
     public void handleActivateToken(ActivateTokenReq message) {
         try {
             if (!message.getActivate()) {
-                PasswordStore.storePassword(message.getTokenId(), null);
+                tokenPinStoreProvider.clearPin(message.getTokenId());
             } else if (message.hasPin()) {
-                PasswordStore.storePassword(message.getTokenId(), message.getPin().toByteArray());
+                tokenPinStoreProvider.addPin(message.getTokenId(), byteToChar(message.getPin().toByteArray()));
```
ricardas-buc: perhaps "setPin" would be semantically better? Add kinda states that something is always added, but in our case its either set or put

src/service/signer/signer-core/src/main/java/org/niis/xroad/signer/core/tokenpinstore/AutoLoginService.java
```diff
@@ -0,0 +1,179 @@
...
+
+    /**
+     * Attempts to login to a token with retry logic.
+     *
+     * @param tokenId the token ID
+     * @param pin the token PIN
+     * @return the login result
+     */
+    private LoginResult loginTokenWithRetry(String tokenId, char[] pin) {
+        while (true) {
```
ricardas-buc: This seems unsafe. How about long timeouts? It should stop retrying eventually.  I think we should reuse already existing retry patterns (we had poc for HC) or use libraries for that.

src/service/signer/signer-core/src/main/java/org/niis/xroad/signer/core/tokenpinstore/TokenPinStoreProvider.java
```diff
@@ -0,0 +1,35 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.niis.xroad.signer.core.tokenpinstore;
+
+import java.util.Optional;
+
+public interface TokenPinStoreProvider {
+    Optional<char[]> getPin(String tokenId);
+    void addPin(String id, char[] password);
```
ricardas-buc: I'd suggest set or put instead of add

src/service/signer/signer-core/src/main/java/org/niis/xroad/signer/core/tokenpinstore/VaultTokenPinStoreProvider.java
```diff
@@ -23,35 +24,40 @@
...
-        storePassword("bar", "barPwd".getBytes(UTF_8));
+    @Override
+    public Optional<char[]> getPin(String tokenId) {
+        return vaultClient.getTokenPin(tokenId);
+    }
 
-        assertArrayEquals("fooPwd".toCharArray(), getPassword("foo").orElseThrow());
-        assertArrayEquals("barPwd".toCharArray(), getPassword("bar").orElseThrow());
+    @Override
+    public void addPin(String tokenId, char[] pin) {
```
ricardas-buc: might have missed somewhere in flow, but do we check anywhere for nulls? empty pin?

deployment/native-packages/src/xroad/common/secret-store-local/usr/share/xroad/scripts/_openbao.sh
```diff
@@ -172,6 +172,9 @@ path "xrd-secret/message-log/archival/pgp/*" {
 path "xrd-secret/message-log/database-encryption/keys/*" {
   capabilities = ["read", "list", "create", "update"]
 }
+path "xrd-secret/signer/token-pins/*" {
```
ricardas-buc, line 175: once again caugth myself seeing deployment/security-server/k8s/charts/openbao-init/scripts/_openbao.sh . perhaps symlink should be replace with a copy script, otherwise it does get rather confusing. IMO this is a bad practice to had symlink in codebase

ricardas-buc (changes requested):

deployment/native-packages/src/xroad/common/secret-store-local/usr/share/xroad/scripts/_openbao.sh
andresrosenthal: I agree that using a symlink isn’t ideal and can be misleading. The issue is that Kubernetes can’t reference files outside its build context, so if we want a true “single source of truth,” it has to live within the Kubernetes scripts. The OpenBao image build in `build-dev-images.sh` already depends on scripts from native-packages, which means we’d need to introduce a second build context so it can also access `_openbao.sh` from the Kubernetes scripts. To be honest this approach would not be ideal as well.

ricardas-buc: Lets put a pin on it for now. We can get back to this at a later date.

PR #3274, opened 2026-01-29
feat: initialization API v2 with typed DTOs and HasGpgKey RPC

egidijusmwork:
-    Add InitializationApiControllerV2 implementing generated OpenAPI interface (InitializationV2Api) with typed request/response DTOs.
-    Add HasGpgKey RPC to backup manager to replace volatile boolean tracking. 
-    Add step-specific audit events. 
-    Service methods now throw typed exceptions handled by ApplicationExceptionHandler.


src/service/backup-manager/backup-manager-rpc-client/src/main/proto/backup_service.proto
```diff
@@ -91,3 +92,7 @@ message BackupEncryptionStatusResp {
   bool backup_encryption_status = 1;
   repeated string backup_encryption_keys = 2;
 }
+
+message HasGpgKeyResp {
```
ricardas-buc: I'd suggest a more generic name like BackupGpgStatus,  bool either 'initialized' or 'exists'. We might reuse and expand this message later

src/security-server/openapi-model/src/main/resources/META-INF/openapi-definition.yaml
```diff
@@ -2731,6 +2733,195 @@ paths:
           description: an existing item already exists
         '500':
           description: internal server error
+
+  # V2 Initialization Endpoints - Granular step-based initialization
+  /v2/initialization/status:
```
ricardas-buc: I'd like to see separate openapi definition for v2, this will make deprecation easier.

src/security-server/openapi-model/src/main/resources/META-INF/openapi-definition.yaml
```diff
@@ -2731,6 +2733,195 @@ paths:
...
+        '401':
+          description: authentication credentials are missing
+        '403':
+          description: request has been refused
+        '409':
+          description: step already completed or prerequisite not met
+        '500':
+          description: internal server error
+
+  /v2/initialization/run-all:
```
ricardas-buc: v1 endpoint technically will do the same? I'd rather not see convenience apis if they're not really in a main flow.

src/security-server/admin-service/application/src/test/java/org/niis/xroad/securityserver/restapi/openapi/InitializationApiControllerV2Test.java
```diff
@@ -0,0 +1,347 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.niis.xroad.securityserver.restapi.openapi;
+
+import org.junit.Test;
```
ricardas-buc: We do have an unwritten rule to write any new and updated tests with junit6. Please transition this one as well

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/InitializationStepService.java
```diff
@@ -0,0 +1,400 @@
...
+    private final TokenPinValidator tokenPinValidator;
+    private final SecurityServerBackupService securityServerBackupService;
+    private final EncryptionInitializationService encryptionInitializationService;
+    private final VaultClient vaultClient;
+    private final BackupManagerRpcClient backupManagerRpcClient;
+
+    /**
+     * Get the complete initialization status with granular step tracking.
+     */
+    public InitializationStatusV2 getInitializationStatusV2() {
```
ricardas-buc: drop V2 from method name

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/InitializationStepService.java
```diff
@@ -0,0 +1,400 @@
...
+
+            switch (stepInfo.getStatus()) {
+                case COMPLETED -> completedSteps.add(step);
+                case FAILED -> failedSteps.add(step);
+                case NOT_STARTED, IN_PROGRESS, SKIPPED, UNKNOWN -> pendingSteps.add(step);
+                default -> log.warn("Unhandled step status: {}", stepInfo.getStatus());
+            }
+        }
+
+        OverallStatus overallStatus = determineOverallStatus(anchorImported, completedSteps, failedSteps);
```
ricardas-buc: Try using java21 features whenever it seems logical. For example var usages for inner variables does make code bit cleaner.

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/InitializationStepService.java
```diff
@@ -0,0 +1,400 @@
...
+        boolean fullyInitialized = completedSteps.size() == InitializationStep.values().length;
+
+        String securityServerId = null;
+        if (serverConfService.isServerCodeInitialized() && serverConfService.isServerOwnerInitialized()) {
```
ricardas-buc, line 121: sonarqube did not trigger here, but it should be on a border of having too much complexity in a single method. Ideally securityServerId, tokenPinPolicyEnforced could be split into their own helper methods

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/dto/InitializationStepInfo.java
```diff
@@ -0,0 +1,128 @@
...
+
+import java.time.Instant;
+import java.util.List;
+
+/**
+ * DTO containing detailed status information for a single initialization step.
+ */
+@Data
+@Builder
+public class InitializationStepInfo {
```
ricardas-buc: could probably be a record?

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/dto/InitializationStatusV2.java
```diff
@@ -0,0 +1,152 @@
...
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * DTO containing the complete initialization status with granular step tracking.
+ */
+@Data
+@Builder
+public class InitializationStatusV2 {
```
ricardas-buc: could be record?

doc/design/ss-initialization-v2-design.md
```diff
@@ -0,0 +1,372 @@
+# Security Server Initialization v2 - Design Document
```
ricardas-buc: Lets not push these types of files.

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/openapi/InitializationApiControllerV2.java
```diff
@@ -0,0 +1,188 @@
...
+    @PreAuthorize("isAuthenticated()")
+    public ResponseEntity<InitializationStatusV2Dto> getInitializationStatusV2() {
+        InitializationStatusV2 status = initializationStepService.getInitializationStatusV2();
+        return ResponseEntity.ok(convertStatus(status));
+    }
+
+    @Override
+    @PreAuthorize("hasAuthority('INIT_CONFIG')")
+    @AuditEventMethod(event = INIT_SERVER_CONF)
+    public synchronized ResponseEntity<InitStepResultDto> initializeServerConf(
```
ricardas-buc: I'd rather see synchronization at service level, especially when we want v1 to reuse same services

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/openapi/InitializationApiControllerV2.java
```diff
@@ -0,0 +1,188 @@
...
+    @AuditEventMethod(event = INIT_MLOG_ENCRYPTION)
+    public synchronized ResponseEntity<InitStepResultDto> initializeMessageLogEncryption() {
+        InitializationStepInfo result = initializationStepService.executeMessageLogEncryptionStep();
+        return createStepResultResponse(result);
+    }
+
+    @Override
+    @PreAuthorize("hasAuthority('INIT_CONFIG')")
+    @AuditEventMethod(event = INIT_ALL_STEPS)
+    public synchronized ResponseEntity<InitializationStatusV2Dto> runAllInitializationSteps(
```
ricardas-buc: mentioned in openapi spec, but will mention here as well.  This kinda does the same as legacy api does, so we should just use v1 endpoint with newer services.  V2 flow by default should call each method separately from UI so there is no need to keep this here.

src/service/backup-manager/backup-manager-rpc-client/src/main/proto/backup_service.proto
```diff
@@ -44,6 +44,7 @@ service BackupService {
 
   rpc GenerateGgpKey(GenerateGpgKeyReq) returns (Empty) {}
   rpc GetBackupEncryptionStatus(Empty) returns (BackupEncryptionStatusResp) {}
+  rpc HasGpgKey(Empty) returns (BackupGpgStatus) {}
```
ricardas-buc: endpoint should be  GetGpgStatus

src/security-server/openapi-model/src/main/resources/META-INF/openapi-definition.yaml
```diff
@@ -27,6 +27,8 @@ tags:
     description: services' endpoints related endpoints
   - name: initialization
     description: initialization endpoints
+  - name: initialization-v2
```
ricardas-buc: whats this?

src/security-server/admin-service/ui/src/router/routes.ts
```diff
@@ -104,6 +105,15 @@ const routes: RouteRecordRaw[] = [
         },
         meta: { permissions: [Permissions.INIT_CONFIG] },
       },
+      {
+        name: RouteName.InitialConfigurationV2,
+        path: '/initial-configuration-v2',
```
ricardas-buc: UI should support only new path so there is no need to prefix v2. V1 should only remain in back-end. Front end should be purely v2.

src/security-server/admin-service/ui/src/locales/en.json
```diff
@@ -17,7 +17,8 @@
   "action": {
     "disable": "Disable",
     "enable": "Enable",
-    "order": "Order"
+    "order": "Order",
+    "retry": "Retry"
```
ricardas-buc, line 21: other locales should be addressed as well

development/hurl/scenarios/test-init-v2.hurl
```diff
@@ -0,0 +1,124 @@
+# V2 Initialization API Tests
```
ricardas-buc: Instead of keeping it separate, we can just improve existing hurl. Make SS0 using V1 and SS1 V2

src/security-server/admin-service/application/src/test/java/org/niis/xroad/securityserver/restapi/openapi/InitializationApiControllerV2Test.java
```diff
@@ -0,0 +1,251 @@
...
+    void initializeSoftTokenWeakPin() {
+        when(initializationStepService.executeSoftTokenStep(anyString()))
+                .thenThrow(new WeakPinException("PIN is too weak", Collections.emptyList()));
+
+        SoftTokenInitRequestDto request = new SoftTokenInitRequestDto("weak");
+
+        try {
+            initializationApiControllerV2.initializeSoftToken(request);
+        } catch (WeakPinException e) {
+            assertEquals("PIN is too weak", e.getMessage());
```
ovidijusnortal: what if logic is changed in `initializeSoftToken` so that `initializationStepService.executeSoftTokenStep` isn't called or exception is handled in some way then this assert isn't called and test is green.

ovidijusnortal: I guess also this apply to other similar tests here.
either add assertFail in try block or better use assertThrows plus some other asserts after that

src/security-server/admin-service/ui/src/openapi-types-v2/models/ErrorInfo.ts
```diff
@@ -0,0 +1,8 @@
+/* generated using openapi-typescript-codegen -- do not edit */
```
ovidijusnortal: generated code should not be included in committed sources

src/security-server/admin-service/ui/src/views/InitialConfiguration/AutoInitStep.vue
```diff
@@ -0,0 +1,152 @@
...
+      <div v-for="sub in subSteps" :key="sub.step" class="sub-step mb-4 d-flex align-center">
+        <v-icon v-if="sub.state === 'completed'" color="success" class="mr-3">mdi-check-circle</v-icon>
+        <v-progress-circular v-else-if="sub.state === 'in-progress'" indeterminate size="20" width="2" class="mr-3" />
+        <v-icon v-else-if="sub.state === 'failed'" color="error" class="mr-3">mdi-alert-circle</v-icon>
+        <v-icon v-else color="grey" class="mr-3">mdi-circle-outline</v-icon>
+
+        <span class="body-regular">{{ $t(sub.label) }}</span>
+      </div>
+
+      <v-alert v-if="errorMessage" type="error" variant="outlined" class="mt-4" closable @click:close="errorMessage = ''">
```
ovidijusnortal: Unless this error is needed to be shown here with particular styling then we have module to show errors in unified fashion with `addError` method from notifications module

this applies to some other components.

src/security-server/admin-service/ui/src/views/InitialConfiguration/AutoInitStep.vue
```diff
@@ -0,0 +1,152 @@
...
+ -->
+<template>
+  <XrdWizardStep>
+    <div class="auto-init-container">
+      <p class="mb-6 body-regular">{{ $t('initialConfigurationV2.autoInit.info') }}</p>
+
+      <div v-for="sub in subSteps" :key="sub.step" class="sub-step mb-4 d-flex align-center">
+        <v-icon v-if="sub.state === 'completed'" color="success" class="mr-3">mdi-check-circle</v-icon>
+        <v-progress-circular v-else-if="sub.state === 'in-progress'" indeterminate size="20" width="2" class="mr-3" />
+        <v-icon v-else-if="sub.state === 'failed'" color="error" class="mr-3">mdi-alert-circle</v-icon>
```
ovidijusnortal: also we don't use mdi-* icons, they are replaced by material symbols 
more about usage can be found in: `src/assets/_material-symbols-rounded.scss`

src/security-server/admin-service/ui/src/views/InitialConfiguration/AutoInitStep.vue
```diff
@@ -0,0 +1,152 @@
...
+
+type SubStepState = 'pending' | 'in-progress' | 'completed' | 'failed';
+
+interface SubStep {
+  step: string;
+  label: string;
+  state: SubStepState;
+}
+
+export default defineComponent({
```
ovidijusnortal: for new components/views and etc we should use composition syntax instead of old Options syntax.


this applies to other components also.

src/security-server/admin-service/ui/src/views/InitialConfiguration/ServerConfStep.vue
```diff
@@ -0,0 +1,210 @@
...
+import { useInitializationV2 } from '@/store/modules/initializationV2';
+
+const props = withDefaults(
+  defineProps<{
+    showPreviousButton?: boolean;
+  }>(),
+  {
+    showPreviousButton: true,
+  },
+);
```
ovidijusnortal: cold be written in a simpler way, like:

```suggestion
const props = defineProps({
    showPreviousButton: {
      type: Boolean,
      default: true,
    },
  });
```

src/security-server/admin-service/ui/src/views/InitialConfiguration/ConfigurationAnchorStep.vue
```diff
@@ -67,63 +67,50 @@
...
-    const { addError } = useNotifications();
-    const { fetchConfigurationAnchor: apiFetchConfigurationAnchor } = useSystem();
-    return { addError, apiFetchConfigurationAnchor };
-  },
-  data() {
-    return {
-      showAnchorDialog: false as boolean,
-      configurationAnchor: undefined as Anchor | undefined,
-    };
+withDefaults(
```
ovidijusnortal: similar case as https://github.com/nordic-institute/X-Road/pull/3274/changes#r2774089981

src/security-server/admin-service/ui/src/openapi-types-v2/models/ErrorInfo.ts
ovidijusnortal: this openapi-types-v2/ folder should be included into .gitignore file

ovidijusnortal (changes requested):
few comments

PR #3325, opened 2026-02-17
feat: Implement readiness probes to enhance k8s orchestration

andresrosenthal:


src/service/proxy/proxy-core/src/main/java/org/niis/xroad/proxy/core/healthcheck/readiness/MessagelogDatabaseReadinessCheck.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Override
+    protected String getCheckName() {
+        return "PROXY_MESSAGELOG_DATABASE_READINESS_CHECK";
+    }
+
+    @Override
+    protected String getDatabaseName() {
+        return "messagelog";
```
justasnortal: The DatabaseCtx has the name, maybe use it instead of hardcoding? The other places as well.

src/service/proxy/proxy-core/src/main/java/org/niis/xroad/proxy/core/healthcheck/readiness/ConfClientChannelReadinessCheck.java
```diff
@@ -0,0 +1,61 @@
...
+import org.niis.xroad.confclient.rpc.ConfClientRpcClient;
+
+/**
+ * Readiness check for the gRPC channel to the Configuration-Client service.
+ * Verifies that the channel is in an acceptable connectivity state.
+ */
+@Readiness
+@ApplicationScoped
+@RequiredArgsConstructor
+public class ConfClientChannelReadinessCheck extends GrpcChannelReadinessCheck {
```
justasnortal: All classes extending GrpcChannelReadinessCheck are very similar and appears in multiple modules. Wouldn't it be possible to create @Readiness checks for rpc clients programatically from a single place for each RpcClient extending AbstractRpc client? This would reduce duplications and in case new rpc client is added, the readiness probe would be create automatically. Otherwise it is easy to forget to create it.

src/service/proxy/proxy-core/src/main/java/org/niis/xroad/proxy/core/healthcheck/readiness/GlobalConfReadinessCheck.java
```diff
@@ -0,0 +1,102 @@
...
+ * - Configuration-client needs time to download globalconf on first boot
+ * - Marking as DOWN would cause unnecessary pod restarts and prevent initial setup
+ * <p>
+ * Only actual failure states are considered failures.
+ */
+@Slf4j
+@Readiness
+@ApplicationScoped
+@RequiredArgsConstructor
+public class GlobalConfReadinessCheck implements HealthCheck {
```
justasnortal, line 55: Why we have it only in proxy? Shouldn't we have it in other modules (monitor, op-monitor, signer,...) as well, where the globalconf is used. When REMOTE source is used, it is fetched from ConfigurationClient via grpc.

src/service/proxy/proxy-core/src/main/java/org/niis/xroad/proxy/core/healthcheck/readiness/ConfClientChannelReadinessCheck.java
andresrosenthal: Looks like a sensible proposition, will look into that

src/service/proxy/proxy-core/src/main/java/org/niis/xroad/proxy/core/healthcheck/readiness/GlobalConfReadinessCheck.java
andresrosenthal: * To my eyes Signer needs the globalconf only when importing certificate - do we really want to make signer (and consequentially the whole SS) unavailable when just the certificate import admin action is not working but the signer itself is actually operational?
* Monitor component was excluded from readiness probing explained in the analysis
* Opmonitor perhaps should include this indeed - globalconf validity is checked on every request

src/service/backup-manager/backup-manager-application/src/main/java/org/niis/xroad/backupmanager/application/healthcheck/KubernetesApiReadinessCheck.java
```diff
@@ -0,0 +1,169 @@
...
+            throws GeneralSecurityException, IOException {
+        Path caCertPath = Path.of(k8sProps.caCertPath());
+        if (!Files.exists(caCertPath)) {
+            // Fall back to default SSL context if CA cert not available
+            return SSLContext.getDefault();
+        }
+
+        CertificateFactory cf = CertificateFactory.getInstance("X.509");
+        X509Certificate caCert;
+        try (var is = Files.newInputStream(caCertPath)) {
```
justasnortal, line 157: How often the ca cert can change? Now the SSLContext is created or every readiness request. If the caCert doesn't change, we could cache the SSLContext after creating KubernetesApiReadinessCheck instance.

justasnortal (approved):

src/service/backup-manager/backup-manager-application/src/main/java/org/niis/xroad/backupmanager/application/healthcheck/KubernetesApiReadinessCheck.java
```diff
@@ -0,0 +1,169 @@
...
+        }
+    }
+
+    @SuppressWarnings("checkstyle:MagicNumber")
+    private boolean checkKubernetesApiAccess(BackupManagerReadinessCheckProperties.KubernetesApiProperties k8sProps)
+            throws IOException, GeneralSecurityException {
+        String k8sHost = k8sProps.serviceHost().orElseThrow();
+        String k8sPort = k8sProps.servicePort().orElseThrow();
+
+        String apiUrl = "https://" + k8sHost + ":" + k8sPort + "/healthz";
```
justasnortal: The healthz endpoint is deprecated (since Kubernetes v1.16)

andresrosenthal: Good catch

PR #3376, opened 2026-03-09
 feat: Implemented endpoints in order to retrieve and update configurable properties

MarcDavid155:
Refs: XRDDEV-2953


src/lib/serverconf-impl/src/main/java/org/niis/xroad/serverconf/impl/dao/ConfigurationPropertyDAOImpl.java
```diff
@@ -0,0 +1,50 @@
...
+     * @param session     Hibernate session
+     * @param propertyKey the property key
+     * @return Entity object containing the property if found
+     */
+    public ConfigurationPropertyEntity getConfigurationProperty(Session session, String propertyKey) {
+        return session.createQuery(
+                        "FROM ConfigurationPropertyEntity WHERE propertyKey = :propertyKey",
+                        ConfigurationPropertyEntity.class)
+                .setParameter("propertyKey", propertyKey)
+                .uniqueResult();
```
justasnortal, line 51: The same property key can be defined with scopes and without scope, uniqueResult might fail here.

src/lib/serverconf-impl/src/main/java/org/niis/xroad/serverconf/impl/entity/ConfigurationPropertyEntity.java
```diff
@@ -0,0 +1,71 @@
...
+    private String propertyValue;
+
+    @Column(name = "scope")
+    private String scope;
+
+    @Column(name = "created_at", nullable = false)
+    private Instant createdAt;
+
+    @Column(name = "updated_at", nullable = false)
+    private Instant updatedAt;
```
justasnortal: You can extend AuditableEntity, it has the createdAt, updatedAt.

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/repository/ConfigurationPropertyRepository.java
```diff
@@ -0,0 +1,76 @@
...
+    private final PersistenceUtils persistenceUtils;
+
+    /**
+     *
+     * Find a configuration property by its key.
+     *
+     * @param propertyKey the property key
+     * @return Optional containing the property if found
+     */
+    public Optional<ConfigurationPropertyEntity> findConfigurationPropertyByPropertyKey(String propertyKey) {
```
justasnortal: There might be multiple properties by key - with scope and without.

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/repository/ConfigurationPropertyRepository.java
```diff
@@ -0,0 +1,76 @@
...
+        return Optional.ofNullable(serviceDescriptionDAO.getConfigurationProperty(persistenceUtils.getCurrentSession(), propertyKey));
+    }
+
+    /**
+     * Save or update a configuration property entity.
+     *
+     * @param entity the entity to save or update
+     */
+    public void saveOrUpdate(ConfigurationPropertyEntity entity) {
+        entity.setUpdatedAt(Instant.now());
```
justasnortal: If I remember correctly, it is handled on DB level.

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/ConfigurablePropertiesService.java
```diff
@@ -0,0 +1,132 @@
...
+     * <p>
+     * If a configuration property with the given key is found, its value is updated.
+     * Otherwise, a new {@link ConfigurationPropertyEntity} is created and persisted.
+     *
+     * @param propertyKey   unique key of the system property
+     * @param propertyValue new value for the system property
+     * @param scope         scope of the property (service name or "common")
+     */
+    public void updateConfigurableProperty(String propertyKey, String propertyValue, String scope) {
+        auditDataHelper.put(RestApiAuditProperty.SYSTEM_PARAMETER, Map.of(propertyKey, propertyValue));
```
justasnortal: Maybe it is not good idea to save propertyValue to audit. Some properties might contain sensitive information, it shouldn't appear in logs.

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/ConfigurablePropertiesService.java
```diff
@@ -0,0 +1,132 @@
...
+@Transactional
+@PreAuthorize("isAuthenticated()")
+@RequiredArgsConstructor
+public class ConfigurablePropertiesService {
+
+    private final ConfigurableProperties configurableProperties;
+    private final ConfigurationPropertyRepository repository;
+    private final AuditDataHelper auditDataHelper;
+
+    private static final String PARAMETER_DESCRIPTION_KEY = "systemParameterDescription.";
```
justasnortal: Maybe the hardcoded description prefix should go to frontend? It is not used in backend, no point returning the same value with every property.

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/ConfigurablePropertiesService.java
```diff
@@ -0,0 +1,132 @@
...
+     * If a configuration property with the given key is found, its value is updated.
+     * Otherwise, a new {@link ConfigurationPropertyEntity} is created and persisted.
+     *
+     * @param propertyKey   unique key of the system property
+     * @param propertyValue new value for the system property
+     * @param scope         scope of the property (service name or "common")
+     */
+    public void updateConfigurableProperty(String propertyKey, String propertyValue, String scope) {
+        auditDataHelper.put(RestApiAuditProperty.SYSTEM_PARAMETER, Map.of(propertyKey, propertyValue));
+        repository.findConfigurationPropertyByPropertyKey(propertyKey)
```
justasnortal: We have a list of allowed to edit properties, but no validation before actually changing the value.

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/ConfigurablePropertiesService.java
```diff
@@ -0,0 +1,132 @@
...
+                .ifPresentOrElse(entity -> {
+                    log.debug("system parameter with propertyKey {} found and updated to {}.", propertyKey, propertyValue);
```
justasnortal: Value might be sensitive, do not log.

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/ConfigurablePropertiesService.java
```diff
@@ -0,0 +1,132 @@
...
+                );
+        return systemPropertyDto;
+    }
+
+    private static ConfigurationPropertyEntity createConfigurationProperty(String propertyKey, String propertyValue, String scope) {
+        var entity = new ConfigurationPropertyEntity();
+        entity.setPropertyKey(propertyKey);
+        entity.setPropertyValue(propertyValue);
+        entity.setScope(!scope.equals("common") ? scope : null);
+        entity.setCreatedAt(Instant.now());
```
justasnortal: Handled at db level.

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/ConfigurablePropertiesService.java
```diff
@@ -0,0 +1,132 @@
...
+                        () -> systemPropertyDto.setScope(serviceName)
+                );
+        return systemPropertyDto;
+    }
+
+    private static ConfigurationPropertyEntity createConfigurationProperty(String propertyKey, String propertyValue, String scope) {
+        var entity = new ConfigurationPropertyEntity();
+        entity.setPropertyKey(propertyKey);
+        entity.setPropertyValue(propertyValue);
+        entity.setScope(!scope.equals("common") ? scope : null);
```
justasnortal: Drop the "!scope.equals("common") ? scope : null"

src/security-server/admin-service/application/src/test/java/org/niis/xroad/securityserver/restapi/repository/ConfigurationPropertyRepositoryIntegrationTest.java
```diff
@@ -0,0 +1,101 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.niis.xroad.securityserver.restapi.repository;
+
+import org.junit.Test;
```
justasnortal, line 29: Lets use newer junit for new tests.

src/security-server/admin-service/application/src/test/java/org/niis/xroad/securityserver/restapi/service/ConfigurablePropertiesServiceTest.java
```diff
@@ -0,0 +1,177 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.niis.xroad.securityserver.restapi.service;
+
+import org.junit.Before;
+import org.junit.Test;
```
justasnortal: Use newer junit.

src/security-server/admin-service/application/src/test/java/org/niis/xroad/securityserver/restapi/service/ConfigurablePropertiesServiceTest.java
```diff
@@ -0,0 +1,177 @@
...
+@RunWith(MockitoJUnitRunner.class)
+public class ConfigurablePropertiesServiceTest {
+
+    private static final String PROPERTY_NAME = "xroad.proxy-ui-api.client-timeout";
+    private static final String PROPERTY_VALUE = "10000";
+    private static final String PROPERTY_VALUE_2 = "11000";
+    private static final String DEFAULT_VALUE = "5000";
+    private static final String SCOPE = "proxy-ui-api";
+    private static final String DESCRIPTION_KEY = "systemParameterDescription." + PROPERTY_NAME;
+    private static final String COMMON_SCOPE = "common";
```
justasnortal: A bit confusing.  The scope is meant to be the application name, we don't have "common"

justasnortal (changes requested):

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/repository/ConfigurationPropertyRepository.java
```diff
@@ -0,0 +1,76 @@
...
+
+    /**
+     *
+     * Find a configuration property by its key.
+     *
+     * @param propertyKey the property key
+     * @return Optional containing the property if found
+     */
+    public Optional<ConfigurationPropertyEntity> findConfigurationPropertyByPropertyKey(String propertyKey) {
+        ConfigurationPropertyDAOImpl serviceDescriptionDAO = new ConfigurationPropertyDAOImpl();
```
justasnortal: Why to create instance on every request? 
And variable name should be fixed :)

src/security-server/admin-service/application/src/main/resources/configurable-properties.yaml
```diff
@@ -0,0 +1,13 @@
+configurableProperties:
```
justasnortal, line 1: The jira ticket also mentioned different types of properties, e.g. boolean, string, integer, enums

MarcDavid155: Since these values all get saved to the database as VARCHAR, then this sort of distinction didn't seem required

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/ConfigurablePropertiesService.java
MarcDavid155: Here I didn't implement any validations since we already have documentation that states that values are not directly validated - https://github.com/nordic-institute/X-Road/blob/develop/doc/Manuals/ug-syspar_x-road_v6_system_parameters.md?plain=1#L230

This seems like something that could be improved upon with a follow-up enhancement task in the future though

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/ConfigurablePropertiesService.java
MarcDavid155: The idea with this is that if we define a "common" scoped property (e.g xroad.common.temp-files-path) in [configuration-properties.yaml](https://github.com/nordic-institute/X-Road/pull/3376/changes#diff-d279f2ab7d4882ed03d5a8b63fed26d62067e41064d32aa81b017c4912943d65R10), then we would need to save this property in the database with empty scope value, instead of with other services where the scope will be set with the service name.

However, this could be something that is done on front-end side instead (so in the GET response, if the scope value is set as common, then the update endpoint is called with empty scope parameter instead)

src/security-server/admin-service/application/src/test/java/org/niis/xroad/securityserver/restapi/repository/ConfigurationPropertyRepositoryIntegrationTest.java
MarcDavid155: AbstractFacadeMockingTestContext uses junit 4, so currently kept this the same

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/ConfigurablePropertiesService.java
justasnortal: Users can only modify existing properties from the predefined list — adding new properties is not allowed.

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/ConfigurablePropertiesService.java
```diff
@@ -0,0 +1,120 @@
...
+ * Service that handles configurable system parameters
+ */
+@Slf4j
+@Service
+@Transactional
+@PreAuthorize("isAuthenticated()")
+@RequiredArgsConstructor
+public class ConfigurablePropertiesService {
+
+    private static final String COMMON_SCOPE = "common";
```
justasnortal: I suggest to remove "common" in all places, it might get misleading. The idea for the scope in properties - to be able to define property only for single application. For example all quarkus applications use property "quarkus.log.level". In case I need to change log level for signer only, I should set scope = "signer", other quarkus applications will not be affected.

src/security-server/admin-service/application/src/main/resources/configurable-properties.yaml
```diff
@@ -0,0 +1,13 @@
...
+  signer:
+    - propertyName: "xroad.signer.client-timeout"
+      defaultValue: "60000"
```
justasnortal: Any ideas how not to duplicate default value here and in code @WithDefault(..). There is a high chance that the default value will be changed in one place and forgotten in other.

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/ConfigurablePropertiesService.java
```diff
@@ -0,0 +1,113 @@
...
+     * @param propertyKey   unique key of the system property
+     * @param propertyValue new value for the system property
+     * @param scope         scope of the property (service name or null)
+     */
+    public void updateConfigurableProperty(String propertyKey, String propertyValue, String scope) {
+        repository.findConfigurationPropertyByPropertyKeyAndScope(propertyKey, scope)
+                .ifPresentOrElse(entity -> {
+                    entity.setPropertyValue(propertyValue);
+                    repository.saveOrUpdate(entity);
+                }, () -> repository.saveOrUpdate(createConfigurationProperty(propertyKey, propertyValue, scope)));
```
justasnortal, line 98: Still missing validation. User should be able to add/modify properties only from the predefined list (configurable-properties.yaml)

src/security-server/admin-service/application/src/main/resources/configurable-properties.yaml
```diff
@@ -0,0 +1,13 @@
+configurableProperties:
+  signer:
```
justasnortal: Remove the grouping by application name in this file. I suggest using using flat format, like
```
configurableProperties:
  - propertyName: "xroad.common.client-timeout"
    defaultValue: "30000"
    scope: .... <-- optional, in most cases not needed.
```

With the current format there is no way how to put properties with prefixes "xroad.common-rpc.*", "xroad.common-global-conf.*", "xroad.common-server-conf.*", etc.

justasnortal (changes requested):

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/ConfigurablePropertiesService.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    private SecurityServerConfigurablePropertyDto toSecurityServerSystemParameterDto(ConfigurableProperty parameter) {
+        var systemPropertyDto = new SecurityServerConfigurablePropertyDto();
+        systemPropertyDto.setPropertyName(parameter.getPropertyName());
+        systemPropertyDto.setDefaultValue(parameter.getDefaultValue());
+        systemPropertyDto.setScope(parameter.getScope());
+        repository.findConfigurationPropertyByPropertyKeyAndScope(parameter.getPropertyName(), parameter.getScope())
```
justasnortal, line 106: This db query executed for each property, so ~280 db calls when creating the response. Maybe preload all props into mem before mapping and check the value in mem.

justasnortal (approved):

PR #3407, opened 2026-03-25
feat: add checks for porstgres15 version check, update docs

egidijusmwork:


deployment/native-packages/src/xroad/redhat/SOURCES/database-local/xroad-init-local-postgres.sh
justasnortal, line 1: This is script is for the local database. I think we should also check the Postgres version when remote db is used.

doc/Architecture/arc-tec_x-road_technologies.md
justasnortal, line 1: There is a "Version history" section in every document, we used to update it when making changes to the documents.

justasnortal (commented):
1. Can we also add constraints on the packaging level (in spec files for rhel, and control for ubuntu). So the installation would not even start if requirements are not met.
2. The xroad-installer should be updated as well. For rhel it installs the Postgres packages. For both OS types we could also check if the version is 15+.

deployment/native-packages/src/xroad/redhat/SOURCES/database-local/xroad-init-local-postgres.sh
egidijusmwork: _setup_db.sh already handles both local and remote via psql_master

ricardas-buc (approved):

PR #3427, opened 2026-04-06
feat: As a Developer I want to use "minumumreleaseage" with PNPM to help mitigate supply chain attacks

mikkbachmann:
+ dependabot cooldown for similar reasons

refs: XRDDEV-3159

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
     directory: "/src"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 3
     groups:
       spring-patch:
         applies-to: version-updates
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -4,6 +4,8 @@ updates:
     directory: "/src"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 3
```
ricardas-buc, line 8: perhaps 3 days is too much? actual security vulnerability might be missed

mikkbachmann: based on dependabot docs this should only affect regular version updates but not security updates. From https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#cooldown- : " The `cooldown` option is only available for version updates, not security updates."

ovidijusnortal (approved):

PR #3436, opened 2026-04-09
feat: refactor dynamic request handlers

ricardas-buc:


src/service/proxy/proxy-core/src/main/java/org/niis/xroad/proxy/core/addon/opmonitoring/serverproxy/OpMonitoringServiceHandlerImpl.java
```diff
@@ -54,32 +61,54 @@
...
 /**
  * Service handler for operational monitoring.
+ * This is a top-level {@link ApplicationScoped} CDI singleton — all per-request state is
+ * kept in method-local variables and returned via {@link ServiceHandlerResult}.
+ * The {@code opMonitorHttpClient} field is set after construction via
+ * {@link #setOpMonitorHttpClient(HttpClient)} because it requires SSL keys
+ * from {@link ServerConfProvider} which are only available after startup.
  */
 @Slf4j
+@ApplicationScoped
```
justasnortal: Why not create addon handlers it in ServiceHandlerLoader? Now the bean is created but if addon is disabled, it is not initialized and might not function properly, but it doesn't prevent injecting it somewhere.

justasnortal (approved):

PR #3445, opened 2026-04-13
feat: implement new configurable parameter editing in UI

MarcDavid155:
Refs: XRDDEV-2954


src/security-server/admin-service/ui/src/views/Settings/SystemParameters/EditConfigurablePropertyDialog.vue
```diff
@@ -0,0 +1,134 @@
...
+
+<script lang="ts">
+import { defineComponent, PropType } from 'vue';
+import { useForm } from 'vee-validate';
+import { mapActions } from 'pinia';
+import { XrdFormBlock, XrdFormBlockRow, useNotifications, XrdSimpleDialog } from '@niis/shared-ui';
+import { SecurityServerConfigurableProperty } from '@/openapi-types';
+import { useSystem } from '@/store/modules/system';
+
+export default defineComponent({
```
ovidijusnortal, line 64: we should use composition style syntax(more modern way and also a bit cleaner) for components, specially new ones

ovidijusnortal (changes requested):

ovidijusnortal (approved):

PR #3447, opened 2026-04-14
chore: remove gradle lock files and fix verification metadata

mikkbachmann:
refs: XRDDEV-3159


MarcDavid155 (approved):
Approved, but maybe there should be some documentation on how to update these lockfiles/verification metadata?

development/docs/gradle-lockfiles-and-verification.md
MarcDavid155, line 1: Maybe some script could be made, that would execute these commands in the correct order? This would make it simpler to update these lockfiles in the future

MarcDavid155 (changes requested):

MarcDavid155 (approved):

PR #3457, opened 2026-04-17
chore: Rename issuance service to issuer service

andresrosenthal:


ricardas-buc (approved):
seems good, just jackson v2 might be upped to v3 in step defs

PR #3481, opened 2026-04-22
feat(proxy): strict identifier checks

nortal-stefanc:
Restrict identifiers (member code, subsystem code etc.) to match ^[a-zA-Z0-9'()+,-.=?]*

Introduced system property `xroad.proxy.strict-identifier-checks`: true - reject invalid identifiers, false - logs warning.

System property `xroad.proxy.strict-identifier-checks` is configurable in UI.

Refs: XRDDEV-3069


nortal-stefanc:
I'll focus on the post install / upgrade part while this is in review

src/service/proxy/proxy-core/src/main/java/org/niis/xroad/proxy/core/util/IdentifierValidationService.java
```diff
@@ -0,0 +1,64 @@
...
+
+    private final ProxyProperties proxyProperties;
+
+    public IdentifierValidationService(final ProxyProperties proxyProperties) {
+        this.proxyProperties = proxyProperties;
+    }
+
+    public void checkIdentifier(XRoadId id) {
+        boolean isStrictCheckEnabled = proxyProperties.strictIdentifierChecks();
+        var identifierValidator = IdentifierValidator.get(isStrictCheckEnabled);
```
MarcDavid155: Is there a need for separate legacy/strict validators? Seems to me that we could just always validate using the logic in StrictIdentifierValidator, and then in case it is invalid, check whether we have configured the strict check and either throw exception or log a warning according to that.

src/service/proxy/proxy-core/src/main/java/org/niis/xroad/proxy/core/util/IdentifierValidationService.java
```diff
@@ -0,0 +1,64 @@
...
+        if (identifierValidator.isValid(id)) {
+            return;
+        }
+
+        if (isStrictCheckEnabled) {
```
MarcDavid155: Maybe some unit tests could be created to validate this logic?

MarcDavid155 (changes requested):

src/service/proxy/proxy-core/src/main/java/org/niis/xroad/proxy/core/util/IdentifierValidationService.java
nortal-stefanc: Just following the same pattern as it was implemented in [ee.ria.xroad.common.validation.IdentifierValidator](https://github.com/nordic-institute/X-Road/blob/develop/src/common/common-admin-api/src/main/java/ee/ria/xroad/common/validation/IdentifierValidator.java).

Also LegacyIdentifierValidation (old IdentifierValidation) allows member code  to have empty space [IdentifierValidatorTest](https://github.com/nordic-institute/X-Road/blob/develop/src/service/proxy/proxy-core/src/test/java/org/niis/xroad/proxy/core/util/IdentifierValidatorTest.java#L46), but strict validation doesn't.

nortal-stefanc: After observations, a member code cannot contain empty space and since the old validator (now LegacyIdentifierValidator) was just use with informative role to log a warning, I agree that can be replace completely with the strict pattern: ^[a-zA-Z0-9'()+,-.=?]*

src/service/proxy/proxy-core/src/test/java/org/niis/xroad/proxy/core/util/IdentifierValidationServiceTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+    @Mock
+    private ProxyProperties proxyProperties;
+
+    @InjectMocks
+    private IdentifierValidationService identifierValidationService;
+
+    @ParameterizedTest
+    @CsvSource({"true, false"})
+    void testCheckIdentifier(boolean isStrictCheckEnabled) {
```
MarcDavid155: isStrictCheckEnabled is not used here

MarcDavid155 (changes requested):

src/service/proxy/proxy-core/src/test/java/org/niis/xroad/proxy/core/util/IdentifierValidationServiceTest.java
nortal-stefanc: Yup, leftover, removed.

MarcDavid155 (approved):

andresrosenthal (approved):

PR #3500, opened 2026-05-04
fix: use PostgreSQL 15 as default on Ubuntu Jammy

enelir:
Refs: XRDDEV-3164


deployment/security-server/xroad-installer/tasks/setup_repositories.sh
```diff
@@ -91,6 +112,22 @@ setup_repositories_ubuntu() {
...
+        postgresql-15 \
+        postgresql-client-15 \
+        postgresql-contrib-15; then
+      log_info "PostgreSQL 15 packages installed successfully"
+    else
+      log_die "Failed to install PostgreSQL 15 packages"
+    fi
+
+    log_message ""
+  fi
```
justasnortal: Do we need to install postgres-15 here? This file is meant to be only for setting up required repositories.
If the postgres repo is set up, can't the deb package manager install the required version (need to add version requirements to `control` file)?

enelir: Probl not. Right now in control file are:
```
Depends: xroad-base (=${binary:Version}), postgresql, postgresql-contrib
Depends: xroad-base (=${binary:Version}), postgresql-client
```
It means for noble postgrSQL 16 and for jammy postgreSQL 14 will be install default. We whant for jammy postreSQL 15 at least.

If we put
```
Depends: xroad-base (=${binary:Version}), postgresql (>= 15), postgresql-contrib (>= 15)
Depends: xroad-base (=${binary:Version}), postgresql-client (>= 15)
```
Then for noble postgreSQL 16 and for jammy postgreSQL 18 will be installed

justasnortal (approved):

PR #3516, opened 2026-05-11
fix: remove race condition for management and registration service tls certificate creation

MarcDavid155:
Refs: XRDDEV-3177

### DIFF
--- a/src/central-server/registration-service/src/main/java/org/niis/xroad/cs/registrationservice/config/RegistrationServiceSslBundleRegistrar.java
+++ b/src/central-server/registration-service/src/main/java/org/niis/xroad/cs/registrationservice/config/RegistrationServiceSslBundleRegistrar.java
@@ -26,16 +26,38 @@
  */
 package org.niis.xroad.cs.registrationservice.config;
 
-import lombok.Getter;
-import lombok.Setter;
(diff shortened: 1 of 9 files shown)


src/central-server/registration-service/src/main/java/org/niis/xroad/cs/registrationservice/config/RegistrationServiceSslBundleRegistrar.java
```diff
@@ -26,16 +26,38 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.niis.xroad.common.managementservice.AbstractManagementServiceSslBundleRegistrar;
+import org.niis.xroad.common.vault.VaultClient;
+
+import java.io.IOException;
+import java.security.NoSuchAlgorithmException;
+import java.security.spec.InvalidKeySpecException;
+
+@Slf4j
+public class RegistrationServiceSslBundleRegistrar extends AbstractManagementServiceSslBundleRegistrar {
```
ricardas-buc, line 41: Since registration service now does not create certs, we should mark it as a systemd dependency so that it would always start after management. Otherwise this retry would show quite often

ricardas-buc (changes requested):

ricardas-buc (approved):

PR #3539, opened 2026-05-18
chore: update password

ovidijusnortal:
Refs: XRDDEV-3170


src/security-server/system-test/src/intTest/resources/container-files/etc/xroad/conf.d/local.yaml
```diff
@@ -38,7 +38,7 @@ xroad:
   proxy-ui-api:
     rate-limit-enabled: true
     rate-limit-requests-per-second: 10000
-    rate-limit-requests-per-minute: 10000
+    rate-limit-requests-per-minute: 20000
```
egidijusmwork, line 41: as we have moniotoring in place locally, can we try to check how much we actually do on test run?

MarcDavid155 (approved):

PR #3579, opened 2026-06-02
feat: add message log enable/disable option to security server installer

egidijusmwork:


deployment/security-server/xroad-installer/tasks/configure_messagelog.sh
```diff
@@ -0,0 +1,54 @@
...
+
+set -euo pipefail
+
+# Get the directory where this script is located
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# Source common functions and logging
+source "$SCRIPT_DIR/../lib/common.sh"
+
+CONFIG_FILE="/etc/xroad/conf.d/local.yaml"
```
justasnortal: Since we are moving from file based configuration to DB properties, it would be better to save value directly to DB (note that DB is not present at the moment this script runs)

deployment/security-server/xroad-installer/tasks/configure_messagelog.sh
```diff
@@ -0,0 +1,65 @@
...
+    log_die "$DB_PROPS not found; the Security Server package must be installed before this step"
+  fi
+
+  if ! command -v java >/dev/null 2>&1; then
+    log_die "java not found on PATH; cannot run migration-cli"
+  fi
+
+  if [[ ! -f "$MIGRATION_CLI_JAR" ]]; then
+    log_message "migration-cli not present, downloading..."
+    bash "$DOWNLOAD_MIGRATION_CLI" || log_die "Failed to download migration-cli"
```
justasnortal: There is already an util for setting DB property `/usr/share/xroad/scripts/db_property.sh`, no need to download the heavy migration cli.

justasnortal (dismissed):

deployment/security-server/xroad-installer/xroad-installer.sh
```diff
@@ -515,6 +539,17 @@ main() {
   fi
   log_message ""
 
+  # Step: Apply Message Log choice to the configuration database (now provisioned)
+  if [[ -f "$SCRIPT_DIR/tasks/configure_messagelog.sh" ]]; then
+    if ! XROAD_MESSAGELOG_ENABLED="$XROAD_MESSAGELOG_ENABLED" \
+       bash "$SCRIPT_DIR/tasks/configure_messagelog.sh"; then
```
justasnortal, line 558: In deb, the packages are started when installed, so at this point the proxy might have been started and the property might not be effective until the next restart

egidijusmwork: Fixed deb auto-starts xroad-proxy on install, before the DB write happens. Now when disabling, we mask xroad-proxy before install, set the property, then unmask+start. Enable path needs no deferral since true is already the default (no DB write there). Also clear any stale mask from a prior failed run first.

ovidijusnortal (approved):

PR #3580, opened 2026-06-03
chore: add audit logging for configurable property updates

nortal-stefanc:
Refs: XRDDEV-3192


src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/openapi/SystemApiController.java
```diff
@@ -145,14 +145,21 @@ public ResponseEntity<Set<SecurityServerConfigurablePropertyDto>> getConfigurabl
...
-                securityServerSystemParameterUpdateDto.getPropertyValue(),
-                securityServerSystemParameterUpdateDto.getScope()
-        );
+        var name = securityServerSystemParameterUpdateDto.getPropertyName();
+        var value = securityServerSystemParameterUpdateDto.getPropertyValue();
+        var scope = securityServerSystemParameterUpdateDto.getScope();
+
+        auditDataHelper.put(RestApiAuditProperty.SYSTEM_PROPERTY_NAME, name);
+        auditDataHelper.put(RestApiAuditProperty.SYSTEM_PROPERTY_VALUE, value);
+        if (scope != null) {
```
ovidijusnortal: scope is pretty static and can be mapped from property name, maybe more interesting thing would be here the old value to see what change was made

ovidijusnortal (changes requested):

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/openapi/SystemApiController.java
nortal-stefanc: scope is nullable (based on configurable-properties.yaml and database)

changes: 
- resolving scope with `ObjectUtils.getIfNull(scope, StringUtils.EMPTY)` 
- added audit listener to log property old value, I want to know the new values before update (to have knowledge in case there are exceptions) and to know old value after udpate

ovidijusnortal (approved):

PR #3584, opened 2026-06-05
fix: correct operational monitoring disabled state detection and update translations

enelir:
Refs: XRDDEV-3166


src/security-server/admin-service/ui/src/locales/en.json
```diff
@@ -347,7 +347,7 @@
     "traffic": {
       "client": "Client",
       "date": "Date",
-      "disabledMessage": "Operational data monitoring is disabled for this Security Server. To enable it, please install the operational data monitoring add-on (xroad-addon-opmonitoring).",
+      "disabledMessage": "Operational data monitoring is disabled for this Security Server. To enable it, please install the operational data monitoring add-on (xroad-opmonitor) and enable operational monitoring in the Security Server configuration.",
```
justasnortal: The message is not entirely correct. There is no separate package for opmonitoring addon anymore - the fuctionality merged into code and it should be enabled in configuration if needed.
`xroad-opmonitor` is op monitoring application package, it can be deployed on the same or remote machine.

justasnortal (approved):

PR #3591, opened 2026-06-11
chore: Update ORT workflow and Gradle verification metadata

petkivim:
- Set Java version to Java 25 for CI builds
- Update Gradle verification metadata


raits (approved):
Looks good to me.

PR #3596, opened 2026-06-12
feat: add backup version compatibility verification

MarcDavid155:
Refs: XRDDEV-2533


src/service/auxiliary-service/auxiliary-service-core/src/main/java/org/niis/xroad/auxiliaryservice/core/backup/BackupMetadataService.java
```diff
@@ -0,0 +1,132 @@
...
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.niis.xroad.auxiliaryservice.core.config.BackupProperties;
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Optional;
+
+@Startup
```
justasnortal: Is @Startup required here?

justasnortal:
The Central Server user crontab for creating automatic backups and for deleting the old backup files. Verify the metadata files are handled correctly there (created and deleted later)

deployment/native-packages/src/xroad/common/base/usr/share/xroad/scripts/_create_backup_metadata.sh
```diff
@@ -0,0 +1,77 @@
...
+# Usage: _create_backup_metadata.sh <backup-file>
+# Exit 0 on success, whether or not the backup turns out to be compatible.
+# Exit non-zero only on usage/file errors.
+
+source /usr/share/xroad/scripts/_backup_restore_common.sh
+
+set -e
+
+TEMP_GPG_DIR=/var/tmp/xroad/gpgtmp
+TEMP_TAR_FILE=${TEMP_GPG_DIR}/decrypted_temporary.tar
```
justasnortal: What if concurrent uploads happen? They will collide into the same folder/file

src/service/auxiliary-service/auxiliary-service-core/src/main/java/org/niis/xroad/auxiliaryservice/core/backup/job/repository/FileSystemBackupRepository.java
```diff
@@ -114,7 +119,10 @@ public BackupItem storeBackup(String name, byte[] content) {
         var path = getAbsoluteBackupFilePath(name);
         try {
             Files.write(path, content);
-            return new BackupItem(name, getCreatedAt(path));
+            return new BackupItem(
+                    name,
+                    getCreatedAt(path),
+                    backupMetadataService.determineBackupCompatibility(path));
```
justasnortal: The side effect is hidden in backupMetadataService.determineBackupCompatibility(path). Without going into details, it looks like read only operation. But it actually creates the metadata file and reads is from there. 
Better to separate responsibilities.

src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/service/BackupMetadataService.java
```diff
@@ -0,0 +1,135 @@
...
+        } catch (IOException e) {
+            log.warn("Failed to read backup format version file {}: {}", formatVersionFilePath, e.getMessage());
+        }
+    }
+
+    public boolean isBackupCompatible(Path backupPath) {
+        return readMetadataVersion(backupPath).map(this::isCompatible).orElse(false);
+    }
+
+    public void deleteMetadata(Path backupPath) {
```
justasnortal, line 52: Currently there is no way to pass path not in backups dir (it is checked before this call), but it is a public method deleting the file, maybe it is worth adding check for path traversal, etc.. and allowing delete only in backups folder.

justasnortal:
@greptileai review

src/service/auxiliary-service/auxiliary-service-core/src/main/java/org/niis/xroad/auxiliaryservice/core/backup/BackupMetadataService.java
```diff
@@ -0,0 +1,156 @@
...
+
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Optional;
+
+@ApplicationScoped
+@RequiredArgsConstructor
+@Slf4j
+public class BackupMetadataService {
```
justasnortal, line 39: Lots of code duplication with BackupMetadataService from CS,

src/shared-ui/src/locales/en.json
```diff
@@ -202,6 +203,9 @@
         "title": "File already exists"
       },
       "success": "Backup {file} uploaded successfully"
+    },
+    "compatibility": {
+      "notCompatibleHint": "Backup version is not compatible with the current server version"
```
justasnortal: Is this used?

deployment/security-server/images/base-images/auxiliary-service/files/usr/share/xroad/scripts/containerised/create_backup_metadata.sh
```diff
@@ -0,0 +1,117 @@
...
+
+parse_backup_label "$TAR_LABEL"
+
+if [ -z "$BACKUP_LABEL_VERSION" ]; then
+  log "Could not parse a backup format version from tar label: $TAR_LABEL"
+  exit 0
+fi
+
+log "Parsed backup format version: $BACKUP_LABEL_VERSION (server type: $BACKUP_LABEL_SERVER_TYPE)"
+echo "{\"version\":\"${BACKUP_LABEL_VERSION}\",\"server_type\":\"${BACKUP_LABEL_SERVER_TYPE}\"}" > "${BACKUP_FILE}.metadata"
```
justasnortal: There are multiple places writing json to metadata file. In case format updates, it should be changed in all places. For the maintenance if would be easier if the file format would be in single place (script) and then reused from other scripts, if it doesn't bring extra complexity.

justasnortal (approved):

PR #3675, opened 2026-07-15
fix(hashchain): reject over-large value list before allocating digest…

egidijusmwork:
… array

### DIFF
--- a/src/common/common-message/src/main/java/ee/ria/xroad/common/hashchain/HashChainVerifier.java
+++ b/src/common/common-message/src/main/java/ee/ria/xroad/common/hashchain/HashChainVerifier.java
@@ -290,14 +290,15 @@ public final class HashChainVerifier {
 
             List<AbstractValueType> values = hashStep.getLeft().getHashValueOrStepRefOrDataRef();
 
+            if (values.size() > MAX_VALUES - totalValuesResolved) {
+                throw XrdRuntimeException.systemException(MALFORMED_HASH_CHAIN,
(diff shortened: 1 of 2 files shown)


src/common/common-message/src/main/java/ee/ria/xroad/common/hashchain/HashChainVerifier.java
```diff
@@ -290,14 +290,15 @@ private byte[] resolveHashStep(String uri, HashChainType currentChain)
...
+            if (values.size() > MAX_VALUES - totalValuesResolved) {
+                throw XrdRuntimeException.systemException(MALFORMED_HASH_CHAIN,
+                        "Hash chain exceeds maximum value count of %d".formatted(MAX_VALUES));
+            }
+
             DigestValue[] digests = new DigestValue[values.size()];
 
             for (int i = 0; i < digests.length; ++i) {
                 totalValuesResolved++;
-                if (totalValuesResolved > MAX_VALUES) {
```
justasnortal, line 297: The MAX_VALUES limit is not enforced anymore. Now the check run once when the step starts, but the totalValuesResolved counter is increased by one on every step. When the call recurses into the child the new guard passes (the counter is not increased yet) and when returning back to parent, it can finish the remaining values even if the totalValuesResolved > MAX_VALUES

justasnortal (changes requested):

ovidijusnortal (approved):

PR #3682, opened 2026-07-16
chore: map DSP errors to common error codes

nortal-stefanc:
Map DSP failures to common XRD error codes on the server side to avoid exposing DSP/EDC-specific error codes to consumers.

- Refs: XRDDEV-3207


nortal-stefanc:
These changes are obsolete, main work done in #3706

PR #3713, opened 2026-08-10
feat: add salting to software pin hashing

MarcDavid155:
Refs: XRDDEV-3257


src/service/signer/signer-common/src/main/java/org/niis/xroad/signer/common/config/SignerConfigKeys.java
```diff
@@ -268,6 +268,13 @@ public final class SignerConfigKeys implements ConfigKeyProvider {
             .exposedInUi()
             .build();
 
+    /** {@code xroad.signer.pin-hasher.salt-length} */
+    public static final ConfigKey<Integer> PIN_HASHER_SALT_LENGTH = PIN_HASHER
+            .integer("salt-length")
+            .withDefaultValue(16)
+            .exposedInUi()
```
ovidijusnortal, line 281: there is precreated validator `.withValidator(Validator.positive())`

ovidijusnortal (approved):

PR #3720, opened 2026-08-12
fix: migrate xroad-secret-store-local to OpenBao >= 2.6.0 (removed pin)

nortal-stefanc:
Refs: XRDDEV-3284

This solution is the cleanest and doesn't introduce additional complexity.

1. The script for configuring openbao mirrors are removed from .deb and .rpm packages. The mirrors should be and are configured before `xroad-secret-store-local` is even installed. The the .deb and .rpm package must not have logic for configuring mirrors for it's dependencies. They must be configured before installation.

2. The pin is removed, not required anymore, as of openbao version 2.6.1 there are no problems.

3. On rpm systems, when dnf was resolving openbao, both openbao and openbao-hsm have `Provide: openbao` and openbao-hsm was picked by dnf. This is now solved by adding `excludepkgs=openbao-hsm*` **for local development**. There seems to be no problems using either openbao or openbao-hsm (the -hsm has additional libraries).

5. The minimum version of openbao is now >= 2.6.0.


deployment/native-packages/src/xroad/redhat/SPECS/xroad-secret-store-local.spec
```diff
@@ -10,7 +10,7 @@ Release:            %{rel}%{?snapshot}%{?dist}
 Summary:            Meta-package for local secret store dependencies
 Group:              Applications/Internet
 License:            MIT
-Requires:           jq, openbao < 2.6.0
+Requires:           jq, openbao >= 2.6.0
```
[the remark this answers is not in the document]
nortal-stefanc: Double check, it should now.

development/docker/security-server/openbao/files/entrypoint.sh
```diff
@@ -1,8 +1,9 @@
 #!/bin/sh
 
-bao server -dev \
-  -dev-root-token-id="$BAO_TOKEN" \
-  -dev-listen-address="${BAO_DEV_LISTEN_ADDRESS:-"0.0.0.0:8200"}" &
+# Drops to the unprivileged openbao user for the network-reachable server
```
ricardas-buc: drop the comment, it explains the change itself, but once its done - does not add much.

deployment/.scripts/configure-mirror-openbao-rpm.sh
```diff
@@ -72,6 +72,7 @@ baseurl=${OPENBAO_MIRROR_URL}
 repo_gpgcheck=0
 gpgcheck=1
 enabled=1
+excludepkgs=openbao-hsm*
```
ricardas-buc: where does this openbao-hsm comes from? first time hearing it

ricardas-buc (approved):

PR #3736, opened 2026-08-17
Restructure and update security hardening guide

petkivim:


doc/Manuals/ug-sec_x_road_security_hardening.md
```diff
@@ -48,17 +62,54 @@ This document is licensed under the Creative Commons Attribution-ShareAlike 3.0
...
+For each measure the guide states the decision to be made and why it matters. Where the procedure for carrying it out is already documented in an installation or user guide, that guide is referenced rather than repeated here, so that there is one authoritative description of each procedure and this document does not diverge from it.
+
+### 1.1 Target audience
+
+The intended audience of this User Guide are X-Road administrators (Central or Security Server) who are responsible for X-Road instance set-up and/or everyday management of the X-Road infrastructure.
+
+The guide addresses two distinct roles, whose responsibilities and available controls differ:
+
+* the **X-Road operator**, the governing authority responsible for the Central Server and, where deployed, the Configuration Proxy — that is, for the security policy of the whole X-Road instance;
+* the **Security Server administrator**, responsible for a single member's Security Server.
```
raits: Is this wording maybe too specific? Since hosting providers and potentially other organizations might host Security Servers that themselves host multiple clients, would it be better to say "responsible for maintaining the organisations Security Server"?

doc/Manuals/ug-sec_x_road_security_hardening.md
```diff
@@ -48,17 +62,54 @@ This document is licensed under the Creative Commons Attribution-ShareAlike 3.0
...
+
+* Section [2](#2-x-road-operator-and-security-server-administrator-controls) — hardening addressed to both roles: operating system accounts, the Admin UI, the token PIN policy, backups and the audit log.
+* Section [3](#3-security-server-administrator-controls) — controls available to the administrator of an individual Security Server.
+* Section [4](#4-x-road-operator-controls) — controls available to the governing authority operating the Central Server and the Configuration Proxy.
+
+The section titles name the **role** the controls are addressed to, so that an administrator can tell at a glance which sections are theirs to read. Each section and subsection then states an **Applies to** line naming the **components** its controls apply to. The line is repeated on every subsection rather than left to be inferred from the section above it, so that a reader arriving at a subsection directly — from the table of contents, from a search, or from a link in another document — sees its scope without having to look elsewhere. The innermost subsections, which only elaborate on the control they sit under, carry the line only where their scope is narrower than that control's.
```
raits: I think this section goes a bit too in-depth with the explanation and reasoning for repeating the metadata.

doc/Manuals/ug-sec_x_road_security_hardening.md
```diff
@@ -48,17 +62,54 @@ This document is licensed under the Creative Commons Attribution-ShareAlike 3.0
...
+
+The two are stated separately because they do not coincide: the X-Road operator administers two components that differ in what can be hardened. The Configuration Proxy has no Admin UI and no web application users, so the user management and Admin UI controls of section [2](#2-x-road-operator-and-security-server-administrator-controls) are addressed to the operator but apply only to the Central Server, while the token PIN policy in the same section applies to the Configuration Proxy too. Read the sections that name your role, and within them apply only what the **Applies to** line names for the components you run.
+
+Every section also carries a stable anchor that does not change when sections are renumbered, so that other documents — in particular the X-Road threat model \[[ARC-TM](#Ref_ARC-TM)\] — can cite a control without the citation breaking at the next revision.
```
raits: Not sure if this is relevant information for the reader.

doc/Manuals/ug-sec_x_road_security_hardening.md
```diff
@@ -48,17 +62,54 @@ This document is licensed under the Creative Commons Attribution-ShareAlike 3.0
...
 
-### 1.2 Terms and abbreviations
+* The Security Server Sidecar Security Guide \[[UG-SS-SEC-SIDECAR](#Ref_UG-SS-SEC-SIDECAR)\] covers the Docker host, the Docker daemon and the container runtime;
+* The Kubernetes Security Server Sidecar Security User Guide \[[UG-K-SS-SEC-SIDECAR](#Ref_UG-K-SS-SEC-SIDECAR)\] covers the Kubernetes cluster — secrets, cluster access, network policies and pod security. A Kubernetes deployment still runs containers, so the Docker guide applies there as well.
+
+Two consequences are worth stating plainly. Applying a platform guide alone leaves the X-Road controls in this document unapplied, because nothing in either platform guide sets a token PIN policy, enables backup encryption or forwards an audit log. And section [2.1](#21-user-management) assumes a Linux host that administrators log in to; where a Security Server runs as a container, the equivalent concerns — how administrator credentials are supplied to the container, and who may reach the container runtime or the cluster — are addressed in the platform guides instead.
```
raits: In my opinion this deserves rephrasing as well. Maybe just state "Please note that simply following either the platform guide or this document alone does not cover all bases. In order to properly secure the software, both the applicable platform document and this security hardening document should be reviewed."

doc/Manuals/ug-sec_x_road_security_hardening.md
```diff
@@ -179,18 +257,126 @@ usermod -s /sbin/nologin user
...
+```
+
+When the policy is enforced, a software token PIN must be
+
+* at least 10 characters long, and
+* composed of characters from at least three of the four character classes: lower-case letters, upper-case letters, digits, and special characters.
+
+Only printable ASCII characters are accepted. A PIN containing any character outside that range is rejected regardless of its length.
+
+Some country-specific meta-packages set the parameter already. The Finnish and Estonian Security Server packages both ship `enforce-token-pin-policy = true`, so on those installations the policy is in force without further configuration. Check the effective value before assuming it is unset. The parameter is described for each component in [UG-SYSPAR](#Ref_UG-SYSPAR) sections "Signer parameters: [signer]".
```
raits: I think this paragraph should be moved up to before the talk about how to change the value. The reason is that the introduction paragraph leads users to believe that it is always false in the package and they might not read this until they have already done the modification.

doc/Manuals/ug-sec_x_road_security_hardening.md
```diff
@@ -179,18 +257,126 @@ usermod -s /sbin/nologin user
...
+
+A backup contains the contents of `/etc/xroad` together with a database dump. That includes the software token private keys, the internal TLS key material and the OpenPGP keyring, so a single backup file is enough to reconstruct the server's identity and to use its keys. A backup that leaves the host, or that can be read by anyone with access to the backup storage, is worth as much to an attacker as the key material it contains.
+
+Backups are always signed and the signature is verified on restore, so their integrity is protected out of the box. Encryption is a separate setting and is **off** by default: `backup-encryption-enabled` defaults to `false`. It is recommended to enable it on both the Central Server and the Security Server, and to set at least one additional recipient in `backup-encryption-keyids` — without one, a backup is encrypted only to the server's own key and cannot be recovered if that key is lost.
+
+The procedure is documented in full elsewhere and is not repeated here. For where the parameters are set, how to generate an additional key pair, how to import and trust it in the `/etc/xroad/gpghome` keyring, and how to decrypt a backup, see:
```
raits:
```suggestion
For where the parameters are set, how to generate an additional key pair, how to import and trust it in the `/etc/xroad/gpghome` keyring, and how to decrypt a backup, see:
```

doc/Manuals/ug-sec_x_road_security_hardening.md
```diff
@@ -179,18 +257,126 @@ usermod -s /sbin/nologin user
...
+
+X-Road applies no strength or validity checks to the keys named in the list. Confirming that each key is strong enough, that its fingerprint is the expected one, and that its private key is held securely away from the server is the administrator's responsibility.
+
+Encryption protects the backup file, not the place it is kept. Backups downloaded through the Admin UI or copied to external or long-term storage leave the protection of the host behind, so the storage location, the transfer channel, and who may retrieve a backup all need to be controlled to the same standard as the server itself.
+
+Enabling encryption changes the restore path. Confirm that a backup taken after the change can actually be decrypted and restored before relying on it, and repeat that check periodically. An encrypted backup that cannot be decrypted is a loss of availability, which for backups is as damaging as a loss of confidentiality.
```
raits:
```suggestion
Enabling encryption changes the restore path. Confirm that a backup taken after the change can actually be decrypted and restored before relying on it, and repeat that check periodically. 
```

doc/Manuals/ug-sec_x_road_security_hardening.md
```diff
@@ -179,18 +257,126 @@ usermod -s /sbin/nologin user
...
+
+<a id="ug-sec-audit-log" class="anchor"></a>
+
+### 2.5 Forwarding the audit log
+
+**Applies to:** Central Server, Security Server
 
-### 4.1 Minimum Supported Client Security Server Version
+The audit log records every change an administrator makes to the system state or configuration through the Admin UI or the management REST API, whether the attempt succeeded or failed, together with the user name, the authentication type used and a correlation identifier. The events are enumerated in \[[SPEC-AL](#Ref_SPEC-AL)\]. It is the record of who changed what, and on a Central Server it covers the registry and trust decisions that the whole ecosystem relies on.
```
raits: I would consider rewording the last sentance because it applies to the Security Server as well not just the Central Server. Also the "trust decisions that the whole ecosystem relies on." sounds a bit too much for me personally :)

doc/Manuals/ug-sec_x_road_security_hardening.md
```diff
@@ -179,18 +257,126 @@ usermod -s /sbin/nologin user
...
+
+Kept only on the host, that record is no more trustworthy than the host. Anyone who obtains administrative access — an intruder, or an administrator acting outside their remit — can alter or delete the local file, and the actions that led to the access disappear with it. Forwarding the audit log to an independent system such as a SIEM or a central log server places the record beyond the reach of whoever controls the X-Road host, which is what makes it dependable as evidence.
```
raits:
```suggestion
Kept only on the host, that record is no more trustworthy than the host. Anyone who obtains administrative access — an intruder, or an administrator acting outside their expected role — can alter or delete the local file, and the actions that led to the access disappear with it. Forwarding the audit log to an independent system such as a SIEM or a central log server places the record beyond the reach of whoever controls the X-Road host, which is what makes it dependable as evidence.
```

doc/Manuals/ug-sec_x_road_security_hardening.md
```diff
@@ -179,18 +257,126 @@ usermod -s /sbin/nologin user
...
+
+The audit log is not a record of personal data, but it is sensitive in its own right. The user names in it are technical accounts rather than the names of individuals, and the identifiers and URLs describe members, subsystems and administrative operations rather than people. What it does expose is the administrative surface of the deployment: which accounts exist, what they do, and which of their actions succeed and fail. That is useful to anyone preparing an attack, and its integrity is what its value as evidence rests on. Forwarding the log to a SIEM does not hand that responsibility to the SIEM's operator; it extends it to a second system that now holds the same record.
+
+Correlating events across hosts depends on their clocks. The correlation identifier links records belonging to the same request on one server, but reconstructing a sequence that spans a Security Server, a Central Server and the SIEM relies on those systems agreeing about the time. Keep the hosts synchronised to a trusted time source and monitor for drift, or the order of events in an investigation cannot be trusted.
```
raits: In most cases, I think synchronizing audit logs perfectly between Security Servers, Central Servers and SIEM-s is pretty unlikely. Perhaps this could be a topic for the X-Road operator, but in my opinion this is a bit of a stretch for the security hardening of most Security Servers, who are likely the main audience.

raits (approved):
Overall looks good to me. I would just change the wording in some places.

doc/Manuals/ug-sec_x_road_security_hardening.md
petkivim: Fixed.

doc/Manuals/ug-sec_x_road_security_hardening.md
petkivim: Fixed.

doc/Manuals/ug-sec_x_road_security_hardening.md
petkivim: Fixed.

doc/Manuals/ug-sec_x_road_security_hardening.md
petkivim: Fixed.

doc/Manuals/ug-sec_x_road_security_hardening.md
petkivim: Fixed.

doc/Manuals/ug-sec_x_road_security_hardening.md
petkivim: Fixed.

doc/Manuals/ug-sec_x_road_security_hardening.md
petkivim: Fixed.

doc/Manuals/ug-sec_x_road_security_hardening.md
petkivim: Fixed.

doc/Manuals/ug-sec_x_road_security_hardening.md
petkivim: Fixed.

PR #3738, opened 2026-08-17
feat: generate and persist ACME account keystore password to the database

nortal-stefanc:
- AcmeConfigDocumentCodec — pure parse/mutate/serialize of the xroad.acme document, shared by the existing read path and the new write path.
- DatabaseAccountKeystorePasswordProvider — new AccountKeystorePasswordProvider implementation wired into AcmeBeanConfig, replacing the always-throwing stub. Reads the current document, generates a password, persists it, and refreshes the in-memory AcmeProperties so it's usable immediately without a restart.
- Removed dead version-7 acme.yml deviation codes and their i18n strings across all locales.
- Updated the Security Server user guide's ACME section (account-keystore-password and EAB credentials) to describe the database-backed behavior, and annotated acme.example.yml as migration-only.

Refs: XRDDEV-3070


nortal-stefanc:
Good catch, and correct — thanks for flagging this.
This is a known, pre-existing race (the same unlocked check-then-act shape existed against the shared `acme.yml` file in version 7, not something this PR introduces): `AcmeService.getAccountKeyPair` checks the configured password, asks the provider for a new one, and writes the keystore file as three separate, unsynchronized steps. Two concurrent first-time calls (e.g. ordering the auth and sign certs together) can each generate a different password and end up with a keystore file and a persisted/in-memory password that permanently disagree.
Fixing it properly means adding locking around that check-generate-create sequence in `AcmeService` (ideally per-account, not global), which is a separately-scoped change from database-backed password generation. Tracking it as a follow-up rather than folding it into this PR to keep this one focused.

nortal-stefanc:
This pull request is closed. ACME account keys will be stored in OpenBao #3748

PR #3746, opened 2026-08-24
feat: replace EDC's jetty-core with an owned, vault-backed, hot-reloadable module

andresrosenthal:
Adds lib:edc-jetty-tls, an X-Road-owned drop-in replacement for EDC's
jetty-core: XRoadJettyExtension/XRoadJettyService source the HTTPS
keystore for every connector from OpenBao (tls/ds-https) at boot and
reload it via Jetty's SslContextFactory.reload on a schedule, using one
eagerly-started, shared SslContextFactory per JVM so a single reload
rotates every connector including framework-internal ones (STS). An
empty vault slot or a key with an empty certificate chain ("certificate
pending") fails startup with an actionable error instead of falling
back to a file keystore. The servlet-registration contract (context
aliases resolved to a PortMapping's path, never the alias itself) was
verified against the decompiled stock jetty-core artifact and is
covered by a regression test using distinct alias/path strings, plus a
real-TLS-handshake test proving certificate rotation without restart
across multiple connectors.

Adds lib:edc-tls-reload, a reusable periodic reload seam
(PeriodicMaterialReloader): fingerprint-compare, bounded retries within
a cycle, keep-serving-old-material on exhaustion, and a schedule that
is never cancelled even by an unanticipated exception. The trust slice
is its next consumer.

Completes the swap: removes the stock edc-core-jetty dependency
everywhere it was declared directly (bootstrap-edc-quarkus,
ds-control-plane-application, ds-issuer-service-customization), adds
the xroad.edc-owned-jetty-conventions Gradle convention (applied to all
three ds-* application modules) that excludes org.eclipse.edc:jetty-core
from every compile/runtime classpath configuration — including where it
still arrives transitively via the edc.http facade artifact — and fails
`check` if it ever resolves again. Also excludes stock JettyExtension
from EDC's service loading via xroad.edc.boot.excluded-service-extensions
as defense in depth. XRoadIssuerRequestHeaderSizeExtension now targets
XRoadJettyService instead of the now-excluded stock JettyService.

Wires ds-identity-hub and ds-issuer-service with a VaultClient CDI bean
(ds-control-plane already had one), resolved by the owned extension
through QuarkusServiceExtensionContext's existing CDI fallback.

Fixes QuarkusVaultClient.readSecret, which mapped every
VaultClientException to "no secret" regardless of cause: now only a 404
maps to empty, mirroring SpringVaultClient's contract, so a vault outage
during DS TLS keystore load (or any other secret read) surfaces as an
infra failure rather than "not provisioned yet".

Refs: XRDDEV-3292


src/lib/edc-jetty-tls/src/main/java/org/niis/xroad/edc/extension/jetty/DsTlsKeyStoreLoadException.java
```diff
@@ -0,0 +1,43 @@
...
+ * THE SOFTWARE.
+ */
+package org.niis.xroad.edc.extension.jetty;
+
+/**
+ * Signals that the DataSpace TLS certificate could not be loaded from OpenBao into a servable keystore,
+ * carrying a message that already tells the operator what to do (provision the vault slot, wait for the
+ * certificate, or check vault connectivity) rather than leaving that to the caller.
+ */
+public class DsTlsKeyStoreLoadException extends RuntimeException {
```
ricardas-buc, line 34: Just to confirm, XrdRuntimeException was not used as this is transient exception just to use within jetty core module? It will not leave its confines

src/lib/edc-jetty-tls/src/main/java/org/niis/xroad/edc/extension/jetty/XRoadJettyExtension.java
```diff
@@ -0,0 +1,136 @@
...
+            defaultValue = "60")
+    private long reloadIntervalSeconds;
+
+    @Inject
+    private VaultClient vaultClient;
+
+    private final PortMappingRegistry portMappingRegistry = new XRoadPortMappingRegistry();
+
+    private XRoadJettyService jettyService;
+    private PeriodicMaterialReloader<?> reloader;
```
ricardas-buc, line 86: naming is bit too vague

src/lib/edc-tls-reload/build.gradle.kts
```diff
@@ -0,0 +1,12 @@
+plugins {
+  id("xroad.java-conventions")
```
ricardas-buc, line 2: any point of keeping this as a separate module? just a single class

ricardas-buc (approved):
just few nitpicks, approving, but I would drop that extra gradle module if there will be nothing else in it

nortal-stefanc (approved):

src/lib/edc-tls-reload/build.gradle.kts
andresrosenthal: This separation is deliberate - it will be reused in XRDDEV-3293 to build hot-reloadable outbound trust for the `ds-*` clients

PR #3747, opened 2026-08-24
Xrddev 3311 per member participant provisioning

ovidijusnortal:


src/service/ds-identity-hub/ds-identity-hub-provisioning-protocol/src/main/proto/identity_hub_provisioning_service.proto
```diff
@@ -78,4 +78,5 @@ message GetParticipantContextExistsReq {
 
 message GetParticipantContextExistsResp {
```
ricardas-buc: Isnt that a simple get and not exists now? And since it might be missing then it should be 
```suggestion
message GetParticipantContextExistsResp {
optional string did =1
```

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/DataspaceProvisioningService.java
```diff
@@ -232,38 +351,106 @@ private String resolveCredentialStatus(String participantId, boolean contextCrea
...
+    public String readIdentityStatus(ClientId memberId) {
+        return assessMemberIdentity(memberId).status();
+    }
+
+    /**
+     * Pin assessment for one member: the status string plus the DID this server intends to publish
+     * (the pinned DID, or the fresh derivation when unpinned; {@code null} when the pin itself is in
+     * an error state and no intended DID can be stated).
+     */
+    private record MemberIdentity(String status, @Nullable String intendedDid) {
```
ricardas-buc: cant we have status as enum?

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/DataspaceProvisioningService.java
```diff
@@ -232,38 +351,106 @@ private String resolveCredentialStatus(String participantId, boolean contextCrea
...
+    }
+
+    /**
+     * The authority (host:port) embedded in derived DIDs. Interim source: the identity-hub host
+     * plus its DID-serving port, because that is where DID documents are actually served. Target
+     * source, once registered-address DID serving exists: the GlobalConf-registered security
+     * server address ({@code GlobalConfProvider#getSecurityServerAddress}), with no port.
+     */
+    private String didAuthority(String identityHubHost) {
+        return identityHubHost + ":" + DID_PORT;
```
ricardas-buc: not your change, but we should not hardcode ports

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/InitializationService.java
```diff
@@ -217,6 +219,19 @@ public void initialize(String securityServerCode, String ownerMemberClass, Strin
     }
 
     private void eagerProvisionDataspaceParticipant() {
+        if (TransactionSynchronizationManager.isSynchronizationActive()) {
```
ricardas-buc: rather uncommon pattern, verify that there were no simpler way.

ricardas-buc (changes requested):

ricardas-buc (approved):

PR #3752, opened 2026-08-26
feat: fail-closed outbound DS TLS trust from approvedDsTlsCa in EDC clients

andresrosenthal:
Replaces the trust EDC's singleton OkHttp client carries with one built
solely from globalconf's approvedDsTlsCa list: the JVM default trust
store is no longer merged in, and the member approvedCA list is never
consulted.

New lib/edc-tls-trust (depends on edc-tls-reload, not edc-jetty-tls):
- RejectAllTrustManager: explicit reject-all manager for an empty or
  not-yet-loaded DS TLS CA list, independent of JSSE's
  JDK-vendor-dependent empty-truststore behaviour.
- DelegatingTrustManager: swappable-delegate X509ExtendedTrustManager
  so the reload seam can replace trust behaviour on an OkHttpClient
  that, once built, never gets rebuilt.
- VaultEndpointTrust: a trust exception scoped to the exact host and
  port parsed from edc.vault.hashicorp.url (scheme-default ports
  resolved when omitted), built from the same CA file the Quarkus
  vault client already trusts; every host-indeterminate or
  unconfigured/unparseable case degrades to list-only, never to a
  broader match.
- DsTlsCompositeTrustManager: the static composite installed once into
  the OkHttp client; the vault exception applies only on an exact
  host+port match on a real SSLSocket/SSLEngine, everything else
  (including the no-socket overload) falls to the reloadable
  list-derived delegate.
- DsTlsCaTrustManagerLoader: distinguishes "globalconf answered: list
  empty" (successful load, reject-all material) from "globalconf
  unreadable" (throws, so the reload seam keeps last-known-good and
  retries).

lib/bootstrap-edc-quarkus's XRoadTlsOkHttpClientExtension — the
distribution's existing client-wiring seam, decompile-confirmed as the
sole https://github.com/Provider OkHttpClient and shared by EDC's own Hashicorp vault
client — is rewritten to install the composite once and start the
periodic reloader; an unreadable DS TLS CA list at boot degrades to
reject-all rather than aborting startup, unlike the serving keystore.

Decompile-verified client-path inventory (222 EDC library artifacts
across ds-control-plane, ds-identity-hub, ds-issuer-service): the
singleton OkHttpClient/EdcHttpClient is the only trust-relevant seam.
DID resolution, OAuth2/STS, DSP dispatch, and credential/status-list
fetches all route through it; no class in the dependency set
constructs its own OkHttpClient.Builder, SSLContext, TrustManager, or
alternate HTTP client. The proxy's consumer-side DSP path is gRPC to
ds-control-plane, not HTTP — no proxy trust wiring needed.

ds-identity-hub-application gains the globalconf/confclient wiring the
new extension needs (GlobalConfConfigKeys, OcspVerifierConfigKeys,
ConfClientRpcChannelProperties, quarkus.arc.unremovable-types) that
ds-control-plane and ds-issuer-service already carried for other
reasons.

Refs: XRDDEV-3293


src/security-server/e2e-test/src/intTest/resources/compose.aux.yaml
```diff
@@ -3,6 +3,17 @@ services:
     image: ${CS_IMG}
     environment:
       - XROAD_HOST=cs
+      # Root cause of the did.json 204 that survives even with the DS TLS CA registered:
```
ricardas-buc: avoid having comments which  act as a history for the whole reasoning.  Try checking out latest consitution files, I did add extra rules around that

ricardas-buc (approved):

nortal-stefanc (approved):

PR #3770, opened 2026-09-02
fix: unblock Initial Configuration wizard on already-initialized servers

nortal-stefanc:
On a Security Server whose owner member, server code, and anchor are already set (e.g. right after a 7.8 -> 8 upgrade, before the software token is logged back in), the wizard's Owner Member step read its prefill data from a store field that was never populated before the step mounted, so the fields rendered blank and Continue never enabled. InitialConfigurationView now fetches the current security server before the step needs it, and OwnerMemberStep falls back to fetching it itself if the fields are still empty on mount.

Separately, the same already-initialized case left the Owner Member step effectively first in the wizard (the anchor step is hidden once imported), but its Previous button still rendered unconditionally. Clicking it navigated to the anchor step's stepper pane, which has no content in that state, blanking the page. Previous is now only shown when an anchor step actually precedes Owner Member.

Refs: XRDDEV-3344

Screenshot from 2026-09-02 19-17-01

Previous btn bug:
Screenshot from 2026-09-02 19-38-22

### DIFF
--- a/src/security-server/admin-service/ui/src/views/InitialConfiguration/InitialConfigurationView.vue
+++ b/src/security-server/admin-service/ui/src/views/InitialConfiguration/InitialConfigurationView.vue
@@ -49,7 +49,7 @@
         </v-stepper-window-item>
 
         <v-stepper-window-item :value="memberStep">
-          <OwnerMemberStep :value="memberStep" @previous="previousStep" @done="nextStep" />
+          <OwnerMemberStep :value="memberStep" :show-previous-button="!!anchorStep" @previous="previousStep" @done="nextStep" />
         </v-stepper-window-item>
         <v-stepper-window-item :value="pinStep">
           <TokenPinStep :save-busy="pinSaveBusy" @previous="previousStep" @done="tokenPinReady" />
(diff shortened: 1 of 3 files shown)


src/security-server/admin-service/ui/src/views/InitialConfiguration/InitialConfigurationView.vue
```diff
@@ -49,7 +49,7 @@
         </v-stepper-window-item>
 
         <v-stepper-window-item :value="memberStep">
-          <OwnerMemberStep :value="memberStep" @previous="previousStep" @done="nextStep" />
+          <OwnerMemberStep :value="memberStep" :show-previous-button="!!anchorStep" @previous="previousStep" @done="nextStep" />
```
ovidijusnortal: maybe it is possible somehow avoid !!? replace with something more readable

ovidijusnortal (changes requested):

src/security-server/admin-service/ui/src/views/InitialConfiguration/InitialConfigurationView.vue
nortal-stefanc: there's no such things as "maybe", everything is possible, changed

src/security-server/admin-service/ui/src/views/InitialConfiguration/InitialConfigurationView.vue
```diff
@@ -128,6 +128,9 @@ export default defineComponent({
     this.fetchInitializationStatus().catch((error) => {
       this.addError(error);
     });
+    // The current security server only exists once the owner member is set up,
+    // so this fails silently on a server that has not been initialized yet.
+    this.fetchCurrentSecurityServer().catch(() => undefined);
```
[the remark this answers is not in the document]
nortal-stefanc: Can you close this one?

ovidijusnortal (approved):

PR #3772, opened 2026-09-03
feat: Retires ds-https-keystore stop-gap

andresrosenthal:
Retires the `ds-https-keystore` stop-gap everywhere: the openssl/keytool recipes copied across dev/test environments, the hardcoded `changeit` password, the EDC file-keystore config, the JVM truststore options
  and cacerts imports, and every direct OpenBao write of `tls/ds-https`. DS TLS certificates are now provisioned exclusively through the admin-service API — the same manual-upload flow an operator uses — so every
  environment exercises the real certificate path end to end.

  ## How provisioning works now

  One flow everywhere, using the shared admin API available on both SS and CS: generate key → download DN-only CSR → sign at the test CA → upload the certificate. Since the CSR intentionally carries no SANs (the
  private key never crosses the API), the test CA's sign endpoint accepts an optional `san` field, injected through its existing extfile mechanism; an explicit SAN wins over the CSR's own, and the docker test-CA
  copy gains the CSR-SAN re-injection the ansible copy already had. Per-environment DN/SAN lists are supplied through hurl variables.

  - **Compose e2e**: three new `setup.hurl` blocks provision CS, SS0 and SS1 right after the DS TLS CA registration / token logins. The keystore init/seed services, the shared volume and all mounts are gone; ds-*
  containers restart on failure until provisioned.
  - **SS api-test** (no hurl): `SsBaselineSeeder.ensureDsTlsCertificate()` provisions through the same API, check-then-skip so it never rolls back a rotated or ACME-enrolled certificate. `DsTlsAcmeEnrollmentTest`
  still forces its own enrollment.
  - **CS api-test**: the co-located issuer program is removed from this stack — its admin-service vault is MockServer, so an API upload can never reach the issuer's real OpenBao, and nothing in the suite consumes
  the issuer's endpoints. The DS TLS API tests drive the mocked slot directly.
  - **LXD**: the Ansible keystore recipes, preseed file-keystore blocks, xrd-ca system trust-store install and vault-seed tasks are deleted; the post-playbook hurl step provisions all hosts. A systemd drop-in
  keeps ds-* units retrying until then.
  - **kind k8s**: the `ds-https-keystore` chart and role (and the interim curl seeding) are deleted; the e2e-fixtures hurl job provisions through the API. The security-server role's availability wait excludes the
  intentionally crash-looping ds-* Deployments; a post-fixtures wait asserts their convergence.

  ## Initialization no longer blocks on dataspace provisioning
  
  The first CI run exposed an ordering dependency the stop-gap had been masking: `POST /initialization` synchronously provisioned dataspace contexts over gRPC into ds services that — correctly, by design — fail
  fast until their DS TLS slot is filled. On the CS this deadlocked initialization outright (the issuer cannot start before the certificate exists, and the ADR flow only produces the certificate after
  initialization, for the manual CSR path in production too); on the SS it stalled the init request for three full gRPC deadlines.

  Fixed following the SS's own established pattern:

  - **CS**: `initialize()` no longer calls the issuer. A new level-triggered `DataspaceIssuerProvisioningWorker` (30s ticks, skips until initialized, stops after one full success, failures non-fatal) retries
  until the issuer becomes available, plus a non-blocking eager attempt right after initialization — mirroring the SS participant provisioning worker.
  - **SS**: the eager post-init provisioning tick runs on a background thread (`provisionParticipantAsync`); the scheduled worker remains the convergence guarantee.
  - **SS api-test**: the seeder re-authenticates after initialization (the init call could outlive the stack's 1-minute UI session timeout).

  ## Post-state
  
  Zero occurrences repo-wide of `changeit`, `javax.net.ssl.trustStore` options, `EDC_WEB_HTTPS` file-keystore env, and vault-seed machinery; the only surviving `ds-https-keystore` string is the owned Jetty
  reloader's thread name (real-path code). The fail-closed design constant holds everywhere: ds-* components fail fast on an empty slot and retry, nothing health-gates on ds-* before the upload, and convergence
  is asserted by pollers that run after provisioning (DSP ISSUED gates in hurl, docker/k8s waits).
  
  ## Out of scope
  
  - Kubernetes ingress TCP/SNI passthrough (XRDDEV-3295)
  - Upgrade migration of existing DS TLS material and guide updates (XRDDEV-3296)
  - CS ACME auto-enrollment (XRDDEV-3291) — the CS provisioning block in `setup.hurl` is the single seam that collapses once it lands
  - Production packaging changes
  
  Refs: XRDDEV-3294


src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/dataspace/DataspaceIssuerProvisioningWorker.java
```diff
@@ -0,0 +1,96 @@
...
+    /**
+     * Non-blocking eager trigger for the initialization request path: runs one best-effort
+     * provisioning attempt on a background thread; the scheduled tick remains the convergence
+     * guarantee.
+     */
+    public void provisionAsync() {
+        CompletableFuture.runAsync(this::provisionBestEffort);
+    }
+
+    private void provisionBestEffort() {
```
ricardas-buc: I think race condition with main initialize is possible? Synchronizing here or inside service would make sense

src/central-server/admin-service/api-test/src/intTest/resources/container-files/etc/supervisor/conf.d/xroad.conf
```diff
@@ -34,8 +34,8 @@ user=xroad
 autorestart=true
 priority=300
 
-[program:xroad-ds-issuer-service]
-command=/usr/share/xroad/scripts/wait-for-secret-store.sh sh -c "/usr/share/xroad/scripts/seed-ds-https-vault-secret.sh && exec /usr/share/xroad/bin/xroad-ds-issuer-service"
-user=xroad
-autorestart=true
-priority=350
+; No xroad-ds-issuer-service on this stack: the admin-service's vault here is MockServer
```
ricardas-buc: reasoning is fine, but lets not comment it here. if needed it should be in readme for api tests

development/docker/central-server/files/cs-xroad.conf
```diff
@@ -52,10 +52,15 @@ user=xroad
 autorestart=true
 priority=300
 
+; The Issuer Service serves the DS TLS certificate from OpenBao's tls/ds-https slot,
+; provisioned through the admin API (setup.hurl or the manual-upload flow). Until the
+; slot is populated it fails fast by design; startsecs=0 makes every exit count as a
+; restartable run so supervisord retries it indefinitely instead of going FATAL.
 [program:xroad-ds-issuer-service]
-command=/usr/share/xroad/scripts/wait-for-secret-store.sh sh -c "/usr/share/xroad/scripts/seed-ds-https-vault-secret.sh && exec /usr/share/xroad/bin/xroad-ds-issuer-service"
+command=/usr/share/xroad/scripts/wait-for-secret-store.sh /usr/share/xroad/bin/xroad-ds-issuer-service
```
ricardas-buc, line 56: How would that work in actual production? Since issuer service has our quarkus wrapper, perhaps we can stall EDC startup until our preconditions are met?

development/ansible/roles/xroad-ss/tasks/pre-ss-install.yml
```diff
@@ -126,8 +135,8 @@
 # 5. DS control-plane runtime overrides. Same reasoning as identity-hub above.
 #
 # TEMPORARY (dataspaces dev): remove once xroad-ds-control-plane debconf ships
-#   participant-context-id, trusted-issuer.id, sts.oauth.token.url,
-#   dsp.callback.address, and keystore config.
+#   participant-context-id, trusted-issuer.id, sts.oauth.token.url and
+#   dsp.callback.address config.
 # ---------------------------------------------------------------------------
 - name: preseed ds-control-plane runtime overrides
```
ricardas-buc, line 121: do these still make sense? pre-ss-install and post-ss-install yml weere stop-gap sets of actions, but perhaps we can move everything to correct places now? If its still blocked - we should draft a separate story for it then

development/ansible/roles/xroad-ss/tasks/post-ss-install.yml
ricardas-buc: I assume this temporary safety net can be dropped (relevant parts moved to old patterns)

development/ansible/roles/xroad-cs/tasks/pre-cs-install.yml
```diff
@@ -1,17 +1,16 @@
 ---
 # TEMPORARY (dataspaces dev): everything in this file is a workaround for the fact
```
ricardas-buc, line 2: can we drop it under this pr or not yet?

development/docker/central-server/files/cs-xroad.conf
andresrosenthal: Yea looking into whether stalling EDC startup until TLS material exists in OpenBao is sensible

ricardas-buc (approved):

PR #3778, opened 2026-09-07
Xrddev 3325

ricardas-buc:


sidecar/slim/Dockerfile
```diff
@@ -70,6 +111,8 @@ RUN rm -f \
 ##
 # create a backup of the configuration for version migration and
 # for initializing an empty config volume
+  && test -f /etc/xroad/conf.d/local.ini || touch /etc/xroad/conf.d/local.ini \
```
justasnortal, line 114: why local.ini is created?

ricardas-buc: a remnant that also exists in native packaging. will address under slim refactor

sidecar/files/slim-ss-xroad.conf
```diff
@@ -42,11 +64,14 @@ stdout_logfile_maxbytes=0
...
 user=xroad
 environment=HOME="/var/lib/xroad"
+autostart=false
 autorestart=true
 stopwaitsecs=30
 stdout_logfile=/dev/stdout
 stdout_logfile_maxbytes=0
 
+[group:xroad-services]
+programs=xroad-signer,xroad-confclient,xroad-proxy
```
justasnortal, line 74: shouldn't it include all the xroad apps? proxy-ui-api, opmonitor, etc

ricardas-buc: fair point, but the slim variant is pretty much dead right now. Tests do use full image, this service config is for slim variant.

justasnortal (approved):

PR #3785, opened 2026-09-08
feat: Central Server configurable system parameters (XRDDEV-3190)

nortal-stefanc:
## Summary

Adds a REST API and web UI for the Central Server to view and modify its database-based system parameters, matching the Security Server's existing configurable-properties feature. Reuses the Security Server's role and audit mechanism as-is. The REST contract, persistence layer, and service are all shared between both products via `common-admin-api` — the `configuration_properties` table is schema-identical in both databases (one Liquibase changelog), so a single entity, repository, and service now serve both, with each product supplying only its own key providers and category-to-scope mapping.

## Changes

- Share the configurable-properties REST contract between Security Server and Central Server via `common-admin-api`, migrating Security Server's existing operations onto it.
- Extract the configurable-properties list/edit-dialog UI components into `shared-ui`, with translations for every locale it carries.
- Central Server: new REST API for its `configuration_properties` table.
- Central Server: new dedicated "System Parameters" settings page, gated to the System Administrator role.
- Fix: both products' frontend code now imports the shared DTO from `shared-ui`'s
- Share `ConfigurablePropertiesService`, `ConfigurationPropertyRepository`, and `ConfigurationPropertyEntity` between both products in `common-admin-api`, replacing each product's own near-duplicate copy. Each product now only implements `ConfigurablePropertySource` (`CsConfigurablePropertySource` / `SsConfigurablePropertySource`) to supply its own `ConfigKeyProvider`s and category-to-scope mapping.
- The shared entity drops the `AuditableEntity` mapping: `configuration_properties` already has a DB trigger populating `created_at`/`updated_at` on both products' databases, and neither product's API exposes those columns.
- `INVALID_PROPERTY_VALUE` moves from each product's own `ErrorMessage` enum to the shared `ErrorCode`, alongside `NOT_FOUND`, since the validation path is now shared code.
- Fix: boolean `ConfigKey`s (`BooleanKeyBuilder`) silently accepted any unrecognized string as `false` (`Boolean::parseBoolean` never throws). The converter now delegates to `BooleanUtils.toBooleanObject` and rejects values that don't parse, reporting the accepted literals (`true`/`false`/`1`/`0`/`Y`/`N`/`yes`/`no`/`on`/`off`, any case).

## Testing

- Backend: new unit tests for the Central Server service/repository/catalogue, mirroring the Security Server's equivalent test depth; a new Central Server `api-test` integration test.
- Frontend: existing Security Server test suites pass unchanged; Central Server's own test suites pass.
- New `ConfigurablePropertiesServiceTest` (`common-admin-api`) covering the shared service generically, independent of either product's own catalogue.
- New `BooleanKeyBuilderTest` and `Validator.bool()` coverage for the strict boolean parsing fix.

Refs: XRDDEV-3190


src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/service/ConfigurationPropertyServiceImpl.java
```diff
@@ -0,0 +1,185 @@
...
+     */
+    private static <T> void validateValue(ConfigKey<T> key, String rawValue) {
+        try {
+            var result = key.validate(key.convert(rawValue));
+            if (!result.valid()) {
+                throw new IllegalArgumentException(result.message());
+            }
+        } catch (RuntimeException e) {
+            throw new BadRequestException("Invalid value for property '%s': %s".formatted(key.key(), e.getMessage()),
+                    e, INVALID_PROPERTY_VALUE.build());
```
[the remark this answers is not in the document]
nortal-stefanc: This is a real gap in the shared ConfigKey/Validator mechanism, but it's not something this PR introduces — it's inherited from Security Server's own already-shipped ConfigurablePropertiesService, which has the identical gap for the identical kind of key, including two of the exact same literal keys:
```
  // AdminServiceConfigKeys.java (Security Server, already shipped)
  .string("request-size-limit-regular")
      .withDefaultValue("50KB")
      .exposedInUi()
      .build();                     // no validator — DataSize.parse() can still fail

  .string("authentication-provider")
      .withDefaultValue("PAM")
      .exposedInUi()
      .build();                     // no validator — AuthenticationProviderType.valueOf() can still fail
```
Security Server also exposes several other unvalidated string-backed keys the same way (authentication-key-algorithm, signing-key-algorithm, mail-notification-locale, key-management-api-whitelist, etc.).

ConfigurationPropertyServiceImpl.validateValue() does the right thing here — it delegates to the key's own convert()/validate(), the same mechanism Security Server's service uses. The actual gap is that string-typed ConfigKeys don't have a validator attached on either product, not that Central Server's write path skips validation it should be doing.

Given it's a pre-existing, cross-product limitation rather than a regression in this PR, we'd rather track "add real validators to string-backed ConfigKeys" as its own follow-up affecting both products, rather than scope-creep it into this ticket.

src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/service/ConfigurationPropertyServiceImpl.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    /**
+     * The owning process converts and validates every stored override eagerly at startup and refuses to
+     * start on failure, so a value must never reach the database without passing the key's own converter
+     * and validator here.
+     */
+    private static <T> void validateValue(ConfigKey<T> key, String rawValue) {
+        try {
+            var result = key.validate(key.convert(rawValue));
```
ovidijusnortal, line 172: a bit late and issue on SS to I guess, probably it's better validate string before convert, all though it is easier to test converted, in most cases this probably OK, but for example in case of boolean "string "foo" will become false without errors

src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/service/ConfigurationPropertyServiceImpl.java
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * Service that handles Central Server configurable system parameters, merging the declared
+ * configuration-key registry with the values stored in {@code configuration_properties}.
+ */
+@Slf4j
+@Service
+@Transactional
+@RequiredArgsConstructor
+public class ConfigurationPropertyServiceImpl implements ConfigurationPropertyService {
```
ovidijusnortal: no way to externalize this service? a lot of duplication

ovidijusnortal (changes requested):

src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/service/ConfigurationPropertyServiceImpl.java
nortal-stefanc: The boolean converted can be replaced with apache commons-lang [BooleanUtils#toBooleanObject](https://commons.apache.org/proper/commons-lang/javadocs/api-3.9/org/apache/commons/lang3/BooleanUtils.html#toBooleanObject-java.lang.String-).

There is no validator for boolean properties, will be added and default value.

src/common/common-admin-api/src/main/java/org/niis/xroad/restapi/entity/ConfigurationPropertyEntity.java
```diff
@@ -44,7 +42,7 @@
 @Entity
 @Table(name = ConfigurationPropertyEntity.TABLE_NAME)
 @Access(FIELD)
-public class ConfigurationPropertyEntity extends AuditableEntity {
+public class ConfigurationPropertyEntity {
```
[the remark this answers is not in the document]
nortal-stefanc: The ConfigurationPropertyEntity is removed from serverconf.hibernate.cfg.xml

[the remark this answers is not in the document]
nortal-stefanc: If nothing further, resolve this conversation and update the overall comment.

src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/service/CsConfigurablePropertySource.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.List;
+
+/**
+ * Central Server's {@link ConfigurablePropertySource}: the providers whose exposed keys make up the
+ * Central Server's system-parameters catalogue, and their category-to-scope mapping.
+ */
+@Component
+public class CsConfigurablePropertySource implements ConfigurablePropertySource {
+
+    private static final List<ConfigKeyProvider> CS_PROVIDERS = List.of(
```
ovidijusnortal, line 47: isn't there some common keys used in SS and CS?

ovidijusnortal (approved):

PR #3790, opened 2026-09-09
feat: distribute dataspace issuer trust anchor via globalconf (XRDDEV-3317)

justasnortal:
Security Servers currently learn the dataspace credential issuer through per-environment seeding (a trusted-issuer row and issuer-host setting across Compose, Kubernetes, Ansible and test fixtures). This PR distributes the issuer trust anchor in globalconf instead, signed by the Central Server, so a fresh Security Server needs zero manual dataspace configuration.

**What changed**

- Shared-parameters v7 gains an optional `dataspaceParameters` element carrying the instance's issuer DIDs (one per Central Server node in HA — issuer DIDs are persisted to the shared CS database at provisioning so any node's generator emits the full set). Older schema versions are untouched; federation partners receive the data through shared-parameters as usual.
- Security Servers consume the set directly from globalconf: every distributed DID becomes a trusted issuer, and issuance targets an endpoint resolved from a DID's document, failing over across candidates at targeting time. All seeding is removed from deployment artifacts.
- Trust also *revokes*: a custom swap-set `TrustedIssuerRegistry` (replacing EDC's additive-only default via DI precedence) drops a DID removed from globalconf within one refresh interval, without restart.
- Issuer reachability probing is bounded (5s per probe, 30s unreachable backoff); a lone issuer — the common case — is used directly without probing.

Refs: XRDDEV-3317


src/central-server/admin-service/core/src/main/java/org/niis/xroad/cs/admin/core/dataspace/DataspaceIssuerProvisioningServiceImpl.java
```diff
@@ -66,6 +68,8 @@ public void provisionIssuer() {
         rpcClient.createCredentialDefinition(ISSUER_PARTICIPANT_ID, CREDENTIAL_DEFINITION_ID, CREDENTIAL_TYPE,
                 CREDENTIAL_FORMAT, CREDENTIAL_JSON_SCHEMA, properties.getCredentialJsonSchemaUrl(),
                 CREDENTIAL_VALIDITY_SECONDS, List.of(ATTESTATION_DEFINITION_ID), membershipMappings());
+
+        dataspaceIssuerDidService.register(did);
```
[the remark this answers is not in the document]
justasnortal: management of issuer DIDs will be in followup stories.

src/service/ds-identity-hub/ds-identity-hub-provisioning-api/src/main/java/org/niis/xroad/edc/identityhub/provisioning/IdentityHubProvisioningGrpcService.java
```diff
@@ -136,6 +164,98 @@ private RequestCredentialResp requestCredentialInternal(RequestCredentialReq req
...
+            }
+            return false;
+        }
+    }
+
+    private boolean resolvesToIssuerService(String did) {
+        var resolved = didResolverRegistry.resolve(did);
+        return resolved.succeeded() && resolved.getContent().getService().stream()
+                .anyMatch(service -> service.getType().equalsIgnoreCase(
+                        CredentialRequestManager.ISSUER_SERVICE_ENDPOINT_TYPE));
```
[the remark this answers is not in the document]
justasnortal: The same application (issuer-service) serves the did and provides issuer service endpoint. If the did document was retrieved - the application is alive. Probing the service endpoint separately wouldn't add much.

src/service/ds-control-plane/ds-xroad-control-plane-policy/build.gradle.kts
```diff
@@ -13,8 +13,13 @@ dependencies {
 
     implementation(project(":lib:globalconf-core"))
     implementation(project(":lib:serverconf-core"))
+    implementation(project(":lib:edc-tls-reload"))
```
ricardas-buc: was this because of reloader helpers? Perhaps move em to a a more generic place then?

src/service/ds-control-plane/ds-xroad-control-plane-policy/src/main/java/org/niis/xroad/edc/extension/policy/controlplane/issuertrust/XRoadIssuerTrustAnchorExtension.java
```diff
@@ -0,0 +1,140 @@
...
+        if (reloader != null) {
+            reloader.close();
+        }
+    }
+
+    private Set<String> loadTrustedIssuerDids(Monitor monitor) {
+        var instanceIdentifier = globalConfProvider.getInstanceIdentifier();
+        var dids = Set.copyOf(globalConfProvider.getIssuerDids(instanceIdentifier));
+        if (dids.isEmpty()) {
+            if (notEnabledLogged.compareAndSet(false, true)) {
```
ricardas-buc: notEnabledLogged is used in few places, now sure how I feel about this pattern. At first it seems like its overly protective. I would evaluate this again, but its not a blocker/bug.

src/service/ds-identity-hub/ds-identity-hub-provisioning-protocol/src/main/proto/identity_hub_provisioning_service.proto
```diff
@@ -51,7 +51,10 @@ message CreateParticipantContextResp {
 
 message RequestCredentialReq {
   string participant_context_id = 1;
-  string issuer_did = 2;
+  // The trusted-issuer candidate set to target, in no particular order; the identity hub resolves each
```
ricardas-buc: seems to be a wrong place to leave this comment

src/service/ds-control-plane/ds-xroad-control-plane-policy/src/main/java/org/niis/xroad/edc/extension/policy/controlplane/issuertrust/XRoadIssuerTrustAnchorExtension.java
justasnortal: dropped the log once pattern.

src/service/ds-identity-hub/ds-identity-hub-provisioning-protocol/src/main/proto/identity_hub_provisioning_service.proto
justasnortal: removed

src/service/ds-control-plane/ds-xroad-control-plane-policy/build.gradle.kts
justasnortal: dropped the dependency, using simple scheduledExecutorService

ricardas-buc (approved):

PR #3796, opened 2026-09-10
feat(dataspace): derive counterparty targets from globalconf (XRDDEV-3314)

egidijusmwork:
Replaces the hardcoded, deprecated per-member `CounterPartyTarget` map with pure derivation per XRDADR-41: provider member id + serving Security Server's GlobalConf-registered address produce the counter-party DID and ctx-id. Multi-SS selection: each member@SS is a distinct participant, so the consumer shuffles the candidate serving servers and fails over per candidate. The legacy `-mgmt` synthetic targets stay map-based until the consumer SYSTEM-routing story replaces them.

Because counterparties derive the DSP surface (DID resolution 7183, DSP 8183) from the registered address, that address must serve those ports on every substrate: compose/k8s gain an nginx stream gateway (sidecar in the proxy pod on k8s) forwarding them to the ds-* services; native installs already bind them on the host.

**No migration for existing DIDs.** `didFor` moves the DID authority from the identity-hub host to the registered address. `ds_participant` rows bound under the old authority fail the binding check (`MISMATCH`), and the hub-DID drift guard refuses to touch such contexts. The assumption is wipe-and-reprovision pre-GA; nothing rebinds or migrates.

Notes:

* Member ctx-ids can contain literal `%` (percent-encoded member codes), so the DSP URL embeds the ctx-id path segment re-escaped (`%` → `%25`); a round-trip test asserts the segment decodes back to the registered ctx-id. Provider-side `EDC_DSP_CALLBACK_ADDRESS` `%s` substitution happens inside upstream EDC — a live check with a member code needing encoding (e.g. `A+B`) is still owed before relying on that path.
* The compose ds-gateway nginx confs re-resolve upstreams via Docker's embedded DNS so a restarted backend container keeps working. The k8s ConfigMap keeps a literal `proxy_pass`: Service ClusterIPs survive pod restarts, and the Docker resolver trick does not port to a static ConfigMap — accepted risk if a ds-* Service is ever recreated.
* The lxd e2e failures on earlier runs were develop's #3795 RHEL9 packaging bug (DS services crash-looping pin ss0, timing out ss1's client registration); fixed by #3808, now merged in.

Refs: XRDDEV-3314

https://claude.ai/code/session_01Kbbu6Ygc7Bs4cVq9ZEYcvE


deployment/security-server/k8s/charts/security-server/templates/dsp/ds-gateway-configmap.yaml
```diff
@@ -0,0 +1,35 @@
+{{- /*
+Stream config for the ds-gateway sidecar (proxy pod): serves the DSP surface on the
```
ricardas-buc: I would callout that this approach must be reevaluated once we start working on k8s again. I might expect to see gateway api usages, simple ingress probably does not cover all of our needs (tls termination). Either way, its fine

src/lib/ds-identity-core/src/main/java/org/niis/xroad/ds/identity/DspConventions.java
```diff
@@ -0,0 +1,149 @@
...
+ * so counter-party coordinates are derived per serving server address; selecting among the serving
+ * servers is the caller's concern.
+ */
+@UtilityClass
+public class DspConventions {
+
+    /**
+     * Port of the identity hub's DID resolution endpoint, part of every derived {@code did:web} authority.
+     */
+    public static final int DID_PORT = 7183;
```
ricardas-buc, line 48: I do remember Ovidijus has moved ports to a config

src/lib/properties-core/src/main/java/org/niis/xroad/common/properties/config/keys/AdminServiceConfigKeys.java
```diff
@@ -472,11 +472,6 @@ public final class AdminServiceConfigKeys implements ConfigKeyProvider {
             .string("participant-id")
             .build();
 
-    public static final ConfigKey<Integer> DATASPACE_IDENTITY_HUB_DID_PORT = DATASPACE
```
ricardas-buc, line 475: why?

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/DataspaceProvisioningService.java
```diff
@@ -358,26 +364,57 @@ private CredentialStatus resolveCredentialStatus(String participantId, boolean c
...
+
+    /**
+     * Resolves the server id through the repository, not {@link ServerConfService}: callers include
+     * the unauthenticated scheduled provisioning worker, which the service's authentication guard
+     * would reject. A GlobalConf that is not yet downloaded reads as "no address yet", same as a
+     * registration that has not landed.
+     */
+    private Optional<String> findRegisteredAddress() {
+        return ownerId().flatMap(owner -> {
+            var serverId = SecurityServerId.Conf.create(owner, serverConfRepository.getServerConf().getServerCode());
```
ricardas-buc: I think own address is being resolved already few different places in the code, I would propose adding that to a more convenient place for reuse.  Look for 
`globalConfProvider.getSecurityServerAddress(serverConfProvider.getIdentifier())`

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/DataspaceProvisioningService.java
```diff
@@ -410,12 +449,20 @@ private record MemberIdentity(IdentityStatus status, @Nullable String intendedDi
...
             if (bound.isEmpty()) {
-                return new MemberIdentity(IdentityStatus.UNBOUND, ParticipantIdentifierScheme.memberDid(memberId, ssHost));
+                return new MemberIdentity(IdentityStatus.UNBOUND, null);
             }
+            log.debug("Data space: registered address not in GlobalConf yet, cannot verify bound identity of {}", memberId);
+            return new MemberIdentity(IdentityStatus.UNKNOWN, null);
+        }
+        var ssHost = DspConventions.didAuthority(address.get());
+        if (bound.isEmpty()) {
+            return new MemberIdentity(IdentityStatus.UNBOUND, ParticipantIdentifierScheme.memberDid(memberId, ssHost));
```
ricardas-buc: I would also consider using com.apicatalog.did.Did when transferring Did identifier.

src/security-server/api-test/src/intTest/resources/ds-gateway/nginx.conf
```diff
@@ -0,0 +1,42 @@
+# L4 front for the Security Server's GlobalConf-registered address (the ss0 alias this
```
ricardas-buc: Since both k8s and compose rely on this dev-only approach, why not to add another dev image and reuse it in both places?

src/security-server/e2e-test/src/intTest/resources/ds-gateway/nginx.conf
```diff
@@ -0,0 +1,42 @@
+# L4 front for the Security Server's GlobalConf-registered address (the xrd-ssN alias
+# this container owns on xroad-network): counterparties derive every port of this server
```
ricardas-buc: another repeat or similar gateway config?

src/service/proxy/proxy-dsp-core/src/main/java/org/niis/xroad/proxy/controlplane/CounterPartyTarget.java
```diff
@@ -41,97 +42,55 @@
...
-                        "https://ds-control-plane:8183/api/dsp/ss0/" + DSP_PROFILE_ID)),
-                Map.entry("ss1", new CounterPartyTarget("did:web:ds-identity-hub%3A7183",
-                        "https://ds-control-plane:8183/api/dsp/ss1/" + DSP_PROFILE_ID)),
-                //For k8s: keys are the namespace-qualified proxy Service names each release
-                //registers in globalconf, not the SS hostnames.
-                Map.entry("proxy.ss0", new CounterPartyTarget("did:web:ds-identity-hub.ss0%3A7183",
-                        "https://ds-control-plane.ss0:8183/api/dsp/xrd-ss0/" + DSP_PROFILE_ID)),
-                Map.entry("proxy.ss1", new CounterPartyTarget("did:web:ds-identity-hub.ss1%3A7183",
-                        "https://ds-control-plane.ss1:8183/api/dsp/xrd-ss1/" + DSP_PROFILE_ID)));
+                mgmtEntry("xrd-ss0.lxd"),
```
ricardas-buc, line 79: will these be dropped under another story? I assumed this class should have been removed

ricardas-buc (changes requested):

deployment/security-server/k8s/charts/security-server/templates/dsp/ds-gateway-configmap.yaml
egidijusmwork: Agreed, dev-only stand-in. ADR-42 already fixes the target: the k8s ingress passes DS ports through at TCP/SNI level, TLS terminates in the ds-* pods. Plain Ingress is HTTP-only, so that is Gateway API TLSRoute (passthrough) / TCPRoute; the Gateway then routes 7183/8183 straight to the ds-* Services and the sidecar disappears 1:1. Filing it as a story on the helm/k8s track.

src/lib/ds-identity-core/src/main/java/org/niis/xroad/ds/identity/DspConventions.java
egidijusmwork: Deliberate, and it needs Ovidijus's ack since he added identity-hub-did-port in #3747. After this PR every counter-party derives did:web:{registered-address}%3A7183:v1:… from GlobalConf alone. A provider that moves its DID port to anything else mints DIDs nobody can derive or resolve, so a knob that must equal an ecosystem-wide constant is not a knob — it is a foot-gun. Same for 8183 (DSP URL). Both live in DspConventions as the single source. identity-hub-sts-port and identity-hub-credentials-port stay configurable: they are internal reach-the-hub URLs, no peer derives them. This is interim anyway — ADR-42's target is port-free DIDs served on 443. I will add one paragraph to the ADR-41 refined developer plan so this is written down once.

src/lib/properties-core/src/main/java/org/niis/xroad/common/properties/config/keys/AdminServiceConfigKeys.java
egidijusmwork: details in DspConventions:47 comment

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/DataspaceProvisioningService.java
egidijusmwork: Done: OwnSecurityServerResolver (owner, server id, GlobalConf-registered address; repository-based because the scheduled worker is unauthenticated and ServerConfService would reject it). DataspaceProvisioningService uses it. The five pre-existing ad-hoc lookups on develop (OpMonitoringDataHelper, XRoadDataPlaneManager, AcmeCertificateRenewalWorker, GlobalConfChecker, SystemService) span proxy and admin-service; I left them untouched here and will file a follow-up to converge them.

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/DataspaceProvisioningService.java
egidijusmwork: Did comes from carbon-did 0.3.0, which reaches us only transitively via EDC; nothing in X-Road code uses it and it is not in the version catalog. It is a syntax parser (Did, DidUrl, did:key resolver) with no did:web mapping, so it cannot replace the derivation in ParticipantIdentifierScheme. DID resolution and VC verification already run inside the EDC runtimes (DidResolverRegistry in the identity-hub provisioning extension, identity-trust in the control plane); admin-service only mints strings and pushes them over gRPC, and ds_participant.did is a string. Adding carbon-did to admin-service would be a second DID stack next to EDC's for a syntax check on values we produce. If we want a typed DID, a small ParticipantDid record in the framework-free ds-identity-core is the right shape — proposing that as a follow-up rather than in this PR.

src/security-server/api-test/src/intTest/resources/ds-gateway/nginx.conf
egidijusmwork: Agreed — three copies (api-test, e2e, k8s ConfigMap) is two too many. Plan: one ds-gateway dev image (nginx-unprivileged + /etc/nginx/templates envsubst, upstreams via PROXY_HOST/IH_HOST/CP_HOST), consumed by both compose files and the k8s sidecar, no mounted confs. Not in this round — it touches the dev-image build pipeline and I want it green before it lands; follow-up commit on this PR or a separate small PR?

src/security-server/e2e-test/src/intTest/resources/ds-gateway/nginx.conf
egidijusmwork: comment above

src/service/proxy/proxy-dsp-core/src/main/java/org/niis/xroad/proxy/controlplane/CounterPartyTarget.java
egidijusmwork: Yes, on purpose: only managementMap() survives, the per-member map (defaultMap) is deleted per the AC. The remaining -mgmt targets are removed by XRDDEV-3316 (consumer routes synthetic services to the provider's SYSTEM context) and XRDDEV-3318 (legacy host/-mgmt contexts removed). The class goes with them.

src/lib/ds-identity-core/src/main/java/org/niis/xroad/ds/identity/DspConventions.java
ricardas-buc: Port divergence is a known issue and it will be handled separately. To keep scope of this PR, lets keep Ovidijus approach with configuration.

src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/DataspaceProvisioningService.java
ricardas-buc: EDC resolution might happen too late, it is better to handle did formatting issues sooner. Lets try to use this library. It should also reduce amount of code on our side related to did formatting/validation.

deployment/security-server/k8s/charts/security-server/templates/dsp/ds-gateway-configmap.yaml
ricardas-buc: ok, file the story and assign it to me for refinement.

src/security-server/api-test/src/intTest/resources/ds-gateway/nginx.conf
ricardas-buc: lets fix it in this PR

src/lib/ds-identity-core/src/main/java/org/niis/xroad/ds/identity/DspConventions.java
nortal-stefanc: @ricardas-buc is related to #3816?

egidijusmwork: Done. `identity-hub-did-port` is back (default 7183) and own DIDs are minted under it; `DspConventions.didAuthority(address, port)` is the only port-aware entry point, everything else takes the resulting authority. Counter-parties still derive the constant `DID_PORT`, so the knob only agrees with peers at 7183 — Javadoc says so once, and the divergence gets its own ticket as you said.

egidijusmwork: @nortal-stefanc same convention, different side: #3816 fixes the control plane's own `edc.participant.did` default (10100 → 7183) so the runtime's self-DID sits on the DID port; this PR is about what the admin-service mints and what consumers derive from GlobalConf (7183 constant in `DspConventions`, admin-service knob defaulting to it). No overlap in files; both land independently.

src/security-server/api-test/src/intTest/resources/ds-gateway/nginx.conf
egidijusmwork: Done. One `ds-gateway-dev` image (`development/docker/ds-gateway`: nginx-unprivileged, stream servers generated at startup from `DS_GATEWAY_*` env). Both compose stacks and the k8s sidecar consume it; the three nginx.conf copies and the chart ConfigMap are gone. Built by `build-dev-infra.sh` next to postgres-dev/testca-dev, listed in the snapshot cleanup.

deployment/security-server/k8s/charts/security-server/templates/dsp/ds-gateway-configmap.yaml
egidijusmwork: Filed as XRDDEV-3377 (Gateway API TLS passthrough for the DS ports, XRDADR-42). x-jira has no assignee field, so please grab it for refinement — or I set the assignee by hand.

ricardas-buc (approved):

PR #3806, opened 2026-09-11
revert: RHEL9 upgrade runs DS Liquibase migrations before Java 25 is active

nortal-stefanc:
Reverts #3795.
Symptom: e2e (lxd) fails deterministically at the hurl bootstrap step "Register TestClient to SS1" (setup.hurl:1031) with a sanitized `core.server.clientproxy.io_error` after exhausting retries.
Actual cause (from lxd-provision-logs artifacts): on xrd-ss0, both `xroad-ds-control-plane` and `xroad-ds-identity-hub` crash-loop from install onward:
```
SEVERE Error booting runtime: Failed to bootstrap SQL schema,
error 'ERROR: permission denied for schema public'
```
With ss0's control plane down, ss1's client-registration management request (a DSP exchange targeting ss0) can never complete. It's ss0 only in every run — ss1's identical install is fine — suggesting an ordering/race in the RPM transaction rather than a plain scripting error.
Evidence:
- Branch XRDDEV-3314 at a commit without this change: lxd green, 0 occurrences of the error (run 34468260321).
- Same branch immediately after merging develop (which added only #3795 + two unrelated commits touching ACME scheduling / no packaging): lxd failed 4/4 consecutive runs — 34477555740 (×3 attempts) and 34499468293 — with 68/72/118 crash-loop occurrences on ss0 respectively.
- Other branches on current develop also fail only the lxd variant (e.g. XRDDEV-3315, XRDDEV-3190 runs today).
- k8s/compose/sidecar are unaffected (container images, not RPM).
This unblocks the lxd e2e pipeline while a Java-version-aware fix (run DB setup in `%post` when Java ≥25 is already active, defer to `%posttrans` only when it isn't yet — the actual 7.8→8.0 upgrade case #3795 was trying to fix) is finished separately.
Refs: XRDDEV-3346

### DIFF
--- a/deployment/native-packages/src/xroad/redhat/SPECS/xroad-ds-control-plane.spec
+++ b/deployment/native-packages/src/xroad/redhat/SPECS/xroad-ds-control-plane.spec
@@ -81,11 +81,11 @@ if systemctl is-active %{name} &> /dev/null; then
   touch "%{_localstatedir}/lib/rpm-state/%{name}/active"
 fi
 
-%define init_xroad_ds_control_plane_db()                       \
-    /usr/share/xroad/scripts/setup_ds_controlplane_db.sh ""
-
 %post -p /bin/bash
 umask 027
+
+# Run database setup script
+/usr/share/xroad/scripts/setup_ds_controlplane_db.sh "" || true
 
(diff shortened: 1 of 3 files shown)


nortal-stefanc:
Closing this revert, issue is resolved with #3808

PR #3815, opened 2026-09-15
feat(security-server): bind dataspace participant identities 

ovidijusnortal:


src/security-server/admin-service/application/src/main/java/org/niis/xroad/securityserver/restapi/service/DataspaceParticipantBindingService.java
```diff
@@ -0,0 +1,104 @@
...
+                    .filter(member -> dsParticipantRepository.findByMemberIdentifier(member).isEmpty())
+                    .toList();
+            if (unbound.isEmpty()) {
+                return 0;
+            }
+            if (!authCertRegistered) {
+                log.debug("Data space: no registered authentication certificate, leaving {} member(s) unbound",
+                        unbound.size());
+                return 0;
+            }
```
justasnortal: In case there are N members and authCertRegistered=false (not sure if valid in prod), there will be N db requests and the results are just dropped because !authCertRegistered

justasnortal (approved):

Issue #3562, opened 2026-05-26
Configproxy unstable when integrated with HSM


yashwanth-l:
I have followed the installation guide for X-Road Config Proxy Server as per the [guide](https://github.com/nordic-institute/X-Road/blob/develop/doc/Manuals/ug-cp_x-road_v6_configuration_proxy_manual.md).

## Details of Installations:

**OS: Ubuntu 24.04**

## Details of the setup:

We have setup Hardware Security Module integration using `Securosys Primus PKCS#11 Provider` and then installed `X-Road hardware token add-on` with the config as below:

```txt
sudo mkdir -p /etc/systemd/system/xroad-signer.service.d
sudo tee /etc/systemd/system/xroad-signer.service.d/hsm.conf << 'EOF'
[Service]
Environment="PRIMUS_HOME=/usr/local/primus"
Environment="LD_LIBRARY_PATH=/usr/local/primus/lib:/usr/lib64:/usr/lib"
Environment="PRIMUS_CONFIG_DIR=/etc/primus"
EOF
```

And later Configured `devices.ini` by adding the PKCS#11 module to X-Road's hardware token configuration:

```txt
echo '' >> /etc/xroad/devices.ini
echo '[smartcard]' >> /etc/xroad/devices.ini
echo 'library = /usr/local/primus/lib/libprimusP11.so' >> /etc/xroad/devices.ini
echo 'sign_verify_pin = true' >> /etc/xroad/devices.ini
```

Later we were able to list tokens and log in

```bash
sudo su - xroad
signer-console list-tokens
```

Note the token ID (hex string). Then log in:

```bash
signer-console login-token <TOKEN_ID>
```

## Details of the Issue:

The issue is that once we have the signer service restarted after the above operations, it works for a couple of hours and then the HSM integration disconnects. We were not able to find any resonable logs other than the disconnection has happened.

```log
2026-05-12T00:01:04.279Z [main] WARN  o.n.x.c.core.ConfigurationDownloader - Unable to download Global Configuration. Because [91b1c878-e41d-4f4a-8df1-f0942b814034] [SYSTEM] global_conf_download_url_connection_failure: Failed to connect to the download url for configuration instance ****
2026-05-12T00:01:04.280Z [main] INFO  o.n.x.c.core.ConfigurationDownloader - Downloading configuration from http://****/internalconf?version=6
[...]

PR #3830, opened 2026-09-19
fix(signer): retry uninitialized modules when configuration has not c…

nvsaigokul-sudo:
…hanged (#3562)

When an HSM module worker fails during startup (e.g., due to transient network errors, HSM unreachability, or CKR_GENERAL_ERROR), the module is not added to moduleWorkers. On subsequent scheduled refresh cycles, moduleConf.hasChanged() returns false because devices.ini has not been modified. Consequently, loadModules() exited early without retrying the uninitialized module, leaving the signer permanently unable to access the HSM tokens until restarted.

Allow loadModules() to proceed if any configured module is not yet initialized, retrying uninitialized modules while preserving already running module workers.


raits:
Hello @nvsaigokul-sudo,

Thank you for the contribution! Before we can move forward with checking and potentially merging the PR we require a signed CLA from your side.

Please review the following documents for more information:

- https://github.com/nordic-institute/X-Road/blob/develop/CONTRIBUTING.md
- https://github.com/nordic-institute/X-Road/blob/develop/CLA.md

